PR #3, opened 2017-04-18
Update programming languages document

alexmuller:


programming-languages.md
```diff
@@ -1,51 +1,82 @@
...
+Because both of these points should gradually improve over time, we think
+that Node.js is suitable for creating thin, client-facing applications
+that don't persist data.
+
+## Go
+
+The GOV.UK router is built using Go, and it's the core language for Cloud
+Foundry which runs our Platform as a Service.
+
+We suspect that Go may be suitable for some systems programming, like proxying,
```
surminus: This doesn't seem very clear to me. Is it recommended against or not? "We suspect" sounds like a sentence that could bite us in the future. Having a knowledge of the language (to work with stuff like terraform and cloudfoundry) is different to writing brand new apps (such as GOVUK Router, who no-one is confident of supporting anymore in GOVUK). 

It would be good to have some clarification on the use case on when teams should be using Go (over more established projects).

programming-languages.md
```diff
@@ -1,51 +1,82 @@
...
-There will be sensible reasons to go outside this core group. For example:
+- move around the organisation
+- develop shared components
+- improve their personal development
+- master how they operate applications
 
-- when we're extending an existing codebase/ecosystem
-- when scripting in a particular environment
-- when we're experimenting during an alpha (with an expectation that it's replaced by something from the core set for beta)
+## Java, Python and Ruby
```
annashipman: Can you make the headings clearer, so that you don't have to read the entire document to understand what's being said? (A quick scan would suggest our core languages are Java, Python, Ruby and also JavaScript, Go and Scala...)

So this would be `Our core languages are Java, Python and Ruby`. The heading for Scala would be `Do not use Scala for new projects`. Etc.

programming-languages.md
```diff
@@ -1,51 +1,82 @@
...
-GDS, so further adoption discouraged. Note: does not apply to in-browser
-JavaScript.
+[nodejs]: https://nodejs.org/
 
-### Scala
+Node.js is used by a few projects at GDS; GOV.UK Frontend, GOV.UK Pay and the Performance Platform.
 
-Used in the early days of GDS. GOV.UK Licensing is the only remaining
-application written in Scala. Not for use in new development.
+In the past we've had problems operating and maintaining products written in
```
MatMoore: Does this refer to those three products specifically? Could you give examples of the problems we had?

programming-languages.md
```diff
@@ -1,51 +1,82 @@
...
+Node.js. We suspect that this is because:
+
+- It's younger and less mature than Java, Python and Ruby
+- We have fewer developers with Node.js skills than other tools we use
+
+Because both of these points should gradually improve over time, we think
+that Node.js is suitable for creating thin, client-facing applications
+that don't persist data.
```
MatMoore: Why do we expect the number of developers with Node.js skills to go up if its not one of our core languages?

Why is the recommendation about persistence specifically? Would this exclude an app that modifies data but through an API?

programming-languages.md
```diff
@@ -1,51 +1,82 @@
...
-These are (alphabetically): Java, Python, Ruby. These are successfully being
-used by teams at the moment, and we are confident in how to host and operate
-applications written in them. New development should normally be carried out
-in one of them.
+This should make it easier for developers to:
 
-There will be sensible reasons to go outside this core group. For example:
+- move around the organisation
+- develop shared components
+- improve their personal development
```
nickcolley: Are we going to document what areas we plan on actively growing as a community? 

For example Node.js it seems we've identified room for growth and it mentions that we expect this to improve over time.

As a frontend developer at GDS I'd be encouraged to see some discussion around where we want to be in the future with that in mind.

MatMoore:
I don't know if here is the right place, but if shared components is one of the goals, could we draw attention to some of the shared components we do have? For example, GOV.UK have a template for starting up a new rails project, but nobody ever told me that it existed.

https://github.com/alphagov/govuk-rails-app-template

programming-languages.md
rboulton: Looking at the skill sets of frontend developers (and also to a lesser extent backend developers) applying to GDS, NodeJS is an extremely common skill for people who have been working in the commercial web development environment in the past couple of years.  The quality and depth of the experience people have had varies massively, but it is clearly a technology in heavy adoption, particularly for rendering the web interface.

It is now rare to find frontend developers without some NodeJS experience.

I think the recommendation about persistence is probably a little more specific than I would word it, from what I've heard.  Perhaps replace this paragraph with something like:

> Since Node.js is now frequently used in the wider development ecosystem for building the web frontend and there are thus a lot of supporting tools and technologies, we expect this to gradually improve over time. It may therefore sometimes be appropriate to use this technology for rendering the web interface in our applications. It is hard to set a specific boundary for where Node.js should be used, but our current approach is that Node.js shouldn't be used to implement any significant "application logic". GOV.UK Pay are adopting this approach by using Node.js only to create thin, client-facing applications which do not directly persist data (eg, they may talk to an API for this); they transform an API to a web interface (and back again).  This divide matches reasonably well with the division between frontend and backend areas of specialism, and should allow the frontend to be developed with a reasonable level of decoupling from the backend applications.
>
> Teams that adopt Node.js must take steps to ensure that they have access to sufficient developers with Node.js experience.

programming-languages.md
rboulton: My understanding from many discussions is that there is a general consensus of people that think Go is appropriate for use in systems programming, but a concern that we don't and won't have the skills in place to maintain these well when they're rarely used. How about the following to capture this:

> Go is an appropriate language for some systems programming, like proxying, routing and transforming HTTP requests.  However, since we don't maintain much Go and thus are likely to have few people with the skills to do so, any such code poses a significant risk.  Components should only be written in Go if there is no alternative maintained open source tool which could be used, and after a team has developed a plan to ensure that it will have the skills to maintain the component in future.

surminus: I think that defines it clearly :+1:

alexmuller:
Thanks for all the comments - I've made some changes which hopefully clear things up.

> @MatMoore: I don't know if here is the right place, but if shared components is one of the goals, could we draw attention to some of the shared components we do have? For example, GOV.UK have a template for starting up a new rails project: https://github.com/alphagov/govuk-rails-app-template

I think this is a good place to document components that can be reused across GDS, but the Rails template feels like it's quite heavily coupled with some of the GOV.UK conventions so I'm not sure another project would be able to pick it up and use it easily.

programming-languages.md
```diff
@@ -1,51 +1,107 @@
...
+where Node.js should be used, but our current approach is that Node.js shouldn't
+be used to implement any significant "application logic". GOV.UK Pay are adopting
+this approach by using Node.js only to create thin, client-facing applications
+which do not directly persist data (they may talk to an API for this); Node.js
+transforms an API to a web interface and back again. This divide matches reasonably
+well with the division between frontend and backend areas of specialism, and
+should allow the frontend to be developed with a reasonable level of decoupling
+from the backend applications.
+
+Teams that adopt Node.js must take steps to ensure that they have access to
```
davidillsley: Rather than push this onto the teams, we should ask them to consult the forum before adopting node.js for any new work.

programming-languages.md
```diff
@@ -1,51 +1,107 @@
...
+expires: 2017-10-01
 ---
 
-All programming languages have their own strengths and weaknesses.
-This means that we want to use a range of programming languages in
-order to build services and supporting tools to meet users needs.
-This document does not apply to choosing 'off the shelf' software
-(open source or not).
+We want to use a range of programming languages at GDS because we think
+using the right tool for the job will give us the best chance of building
```
davidillsley: I think 'right tool for the job' isn't a great framing given what this document is trying to do. We're explicitly acknowledging that we're going to use tools which are 'good enough for the job' and align with skills and other things going on in GDS.

programming-languages.md
```diff
@@ -1,51 +1,107 @@
...
+[using client-side JavaScript][manual_js].
+
+[manual_js]: https://www.gov.uk/service-manual/technology/using-progressive-enhancement
+
+The only way we're comfortable writing server-side JavaScript at the
+moment is with [Node.js][nodejs].
+
+[nodejs]: https://nodejs.org/
+
+Node.js is used by a few projects at GDS; GOV.UK Frontend, GOV.UK Pay and the Performance Platform.
```
bruntonspall: Can we make clear that "These projects have made the decision to use Node.JS, and the tech forum is watching the output of this.  No new projects should use Node.JS without prior consultation with Michael Brunton-Spall"

annashipman: The forum or the role, rather than a named person?

PR #6, opened 2017-04-28
Document how we should send emails.

rjw1:
For both applications and low level systems

### DIFF
--- a/sending-email.md
+++ b/sending-email.md
@@ -4,12 +4,12 @@ layout: recommendation
 expires: 2017-11-01
 ---
 
-## Using GOV.UK Notify
+## Sending notifications to users
 
 Use [GOV.UK Notify](https://www.notifications.service.gov.uk/) when you need to send notifications to users, for example password resets and confirmation emails.
 
-## Using Amazon SES
+## Sending notifications to engineers
 
-Use [Amazon SES](https://aws.amazon.com/ses/) when sending automatic notifications from systems(e.g cron or jenkins) to engineers.
+Use [Amazon SES](https://aws.amazon.com/ses/) when sending automatic notifications from systems (for example, cron or jenkins) to engineers.
 
 Only use email notifications from systems when the issue can't be picked up through monitoring or logging.
--- a/sending-email.md
+++ b/sending-email.md
@@ -1,0 +1,15 @@
+---
+title: Sending email
+layout: recommendation
+expires: 2017-11-01
+---
+
+## Using GOV.UK Notify
+
+Use [GOV.UK Notify](https://www.notifications.service.gov.uk/) when you need to send notifications to users, for example password resets and confirmation emails.
+
+## Using Amazon SES
+
+Use [Amazon SES](https://aws.amazon.com/ses/) when sending automatic notifications from systems(e.g cron or jenkins) to engineers.
+
+Only use email notifications from systems when the issue can't be picked up through monitoring or logging.
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 _site/
+.DS_Store


sending-email.md
```diff
@@ -0,0 +1,20 @@
...
+
+## User needs for sending email
+
+- applications notifying users about progress or actions to take
+- low level systems notifying engineers of failure if they cant be made to log errors
+
+## How to send email
+
+Applications notifying users should use [GOV.UK Notify](https://www.notifications.service.gov.uk/).
+Low level systems notifying engineers should use [Amazon SES](https://aws.amazon.com/ses/)
```
rboulton: Can we add a reference to the recommendation in https://alphagov.github.io/gds-tech/alerting.html#principles-for-setting-up-an-alerting-system of "Don’t rely on emails as an automated alerting method for incident responders".

I'm not sure "Low level" is very clear here.  I think we just mean "Automated notifications sent to engineers" - is that right?

What systems do we have currently that send such emails?

rboulton:
Commendably brief - would like a little more clarity on when it's appropriate to be sending emails using SES, as mentioned in a comment. Also, perhaps a note on when we've done this in existing systems (and whether that's something we'd do again) - though only add this if it helps to make it clear when using SES is appropriate.

sending-email.md
rjw1: Yeah, I was wondering how to describe it.

rjw1:
@rboulton I've pushed an update.

I've always been noted for my conciseness.

sending-email.md
```diff
@@ -0,0 +1,20 @@
...
+title: Sending email
+layout: recommendation
+expires: 2017-11-01
+---
+
+
+## User needs for sending email
+
+- applications notifying users about progress or actions to take
+- automated notifications sent to engineers by systems such as cron or jenkins remembering that we [don’t rely on emails as an automated alerting method for incident responders](https://alphagov.github.io/gds-tech/alerting.html#principles-for-setting-up-an-alerting-system)
```
davidillsley: I don't get this, and don't think I agree with it. Results of cron jobs should be written, aggregated, and alerted through logs and associated systems. CI notifications shouldn't be forced via SES unless it's appropriate.

rjw1: yes they should, however not everything gets caught like that. they were examples of things which may get sent via email. 

For instance gov.uk integration deploys arnt easily monitored and people dont notice failed deploys.

bruntonspall:
Waiting on wording changes as agreed in the Tech Forum today

rjw1:
@bruntonspall @davidillsley updated. Is this strong enough?

techforevil:
@rjw1 you please set up a pair writing session with @rosalieDM this week (by 19 May)?

rjw1:
Rewritten in a pair with @RosalieDM. 
@bruntonspall and @davidillsley would you like to have another look.

techforevil:
awaiting @bruntonspall approval

rjw1:
In a discussion on slack it was suggested that this recommendation means that someone should look at documenting on how to use notify with devise for rails apps. /cc @alphagov/notify

pixeltrix:
For Devise the easiest method would probably be to [override `send_devise_notification` in your user model][1] since you'd need to bypass Action Mailer's rendering of templates. You can configure a custom mailer but it probably still expects that to conform to the Action Mailer contract of `Mailer.email(args).deliver_now`. 

[1]: https://github.com/plataformatec/devise/blob/a62faa2c8258ba1c35fe84e147f42c35eccccb8a/README.md#activejob-integration

venusbb:
Hi, regarding Ruby Devise, it sounds like a technical blog may be more suitable. We would give support to writing the blog on this. But we don't think it necessarily fits in part of the Notify's documentation.

sending-email.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Sending email
+layout: recommendation
+expires: 2017-11-01
+---
+
+## Using GOV.UK Notify
```
philandstuff: can we express the section headers in terms of the need, not the solution? ie "sending notifications to users" rather than "GOV.UK Notify"

danmitchell-:
Devise issue solved e.g. https://github.com/openregister/managing-registers/commit/02568d8630aadac521df92d7f91d4d69cf6214ea

rjw1:
@danmitchell- could you chat with Rosalie about writing a blog post about this

PR #7, opened 2017-05-02
Create the hosting recommendation

bruntonspall:
Create a first stab at a hosting recommendation

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -45,6 +45,7 @@ The following standards will help you name your service, choose an appropriate p
 
 These standards will enable us to maintain a consistent operating environment across our services:
 
+- [hosting](standards/hosting.html)
 - [managing DNS](standards/dns-hosting.html)
 - [monitoring your service](standards/monitoring.html)
 - [operating systems](standards/operating-systems.html)
--- a/source/standards/hosting.html.md.erb
+++ b/source/standards/hosting.html.md.erb
@@ -67,6 +67,5 @@ Services which have the /potential/ to raise switching costs include:
   * CloudFront
   * DynamoDB
   * Lambda
-  * Route 53
   * Simple Notification Service (SNS)
   * Simple Queue Service (SQS)
--- a/source/standards/hosting.html.md.erb
+++ b/source/standards/hosting.html.md.erb
@@ -37,19 +37,36 @@ Amazon Web Services (AWS) offer scalable computing, storage and deployment servi
 * Amazon Web Services
   https://aws.amazon.com
 
-### Lock in concerns
+### Consider switching costs
 
-In order to minimise costs if you switch services, we determine that for the IaaS model, we will only use the core features of AWS that are common with the biggest competitors, for example:
+AWS has a large number of different services available.  Some are common to many providers, such as compute and file storage.  Others are more specific to AWS itself.
 
-  * Compute
+#### Common services
+
+It's fine to use the core features of AWS that are common with the biggest competitors, for example:
+
+  * Compute (EC2)
   * Networking (load balancing and network segregation)
   * File storage (EBS and S3)
-  * Hosted databases (Amazon Amazon Relational Database Service and Amazon DynamoDB)
+  * Hosted databases (Amazon Relational Database Service (RDS))
 
-  Other platform-like services should be weighed against the cost of switching to another service before adopting, for example:
+#### Less common services
 
-  * CloudFront
-  * Route 53
-  * Simple Queue Service (SQS)
-  * Simple Notification Service (SNS)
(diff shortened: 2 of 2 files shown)


hosting.md
```diff
@@ -0,0 +1,32 @@
...
+
+GDS has selected two cloud platforms:
+
+### PaaS
+
+* GOV.UK PaaS
+
+### IaaS
+
+* Amazon Web Services
```
bruntonspall: Add a note about what AWS products we are selecting.  We are aiming for minimal IaaS tools, rather than the value add tools, which helps us reduce lockin

hosting.md
```diff
@@ -0,0 +1,32 @@
...
+## User Needs
+
+* Users want systems that are highly scalable and available
+* GDS wants to manage the environments easily and simply, with confidence via automated tools
+
+## Principles
+
+GDS will build new tools on only one cloud platform for each of IaaS and PaaS.
+
+We prefer systems built on a PaaS to systems built on IaaS where possible.
```
bruntonspall: Add stronger statement, as per programming languages

hosting.md
```diff
@@ -0,0 +1,32 @@
...
+GDS follows the [Government Cloud First policy](https://www.gov.uk/guidance/government-cloud-first-policy) meaning that we prefer the use of PaaS and IaaS solutions over managing our own hardware
+
+## User Needs
+
+* Users want systems that are highly scalable and available
+* GDS wants to manage the environments easily and simply, with confidence via automated tools
+
+## Principles
+
+GDS will build new tools on only one cloud platform for each of IaaS and PaaS.
```
surminus: I was wondering what the scope of "new tools" is? As an example, if we wanted to use storage services for something like offsite backups or DNS services I think it would make sense to allow the scope to use more than a single vendor.

bruntonspall: Yes.  We're going to change this language a little to make that clearer.

techforevil:
@bruntonspall can you please schedule some time for pair writing with @rosalieDM this week (by 19 May)? ta!

techforevil:
@bruntonspall you said there's some work to be done on this before you can sit down with a tech writer - can you let me know what that is please and whether Alex or bob can do anything to help

hosting.md
tomskerous: If we say we prefer PaaS to IaaS, should that be first?

hosting.md
```diff
@@ -0,0 +1,40 @@
...
+### IaaS
+
+* Amazon Web Services
+  https://aws.amazon.com
+
+### Lock in concerns
+
+In order to minimise lock-in, we determine that for the IaaS model, we will use only the core features of AWS that are common with the biggest competitors.  That is to say we will be using  for example: Compute, Networking (load balancing and network segregation), File storage (EBS and S3), Hosted databases (RDS and DynamoDB).
+
+Other platform like services (Cloudfront, Route 53, SQS, SNS, Directory Service and so on) need to be weighed against the lock-in cost before adopting.
```
rboulton: We'll include lambda in this list.

hosting.md
```diff
@@ -0,0 +1,40 @@
...
+  https://cloud.service.gov.uk 
+
+### IaaS
+
+* Amazon Web Services
+  https://aws.amazon.com
+
+### Lock in concerns
+
+In order to minimise lock-in, we determine that for the IaaS model, we will use only the core features of AWS that are common with the biggest competitors.  That is to say we will be using  for example: Compute, Networking (load balancing and network segregation), File storage (EBS and S3), Hosted databases (RDS and DynamoDB).
```
philandstuff: can we phrase things in terms of switching costs instead of lock in?

I think "lock in" is a poor term. Firstly, it leads people to think in binary terms: "we are/aren't locked in", instead of "it'll cost us a little/a lot to switch provider". Secondly, it leads people to divide services into two buckets: lock-in-prone services and generic services, and that the way to avoid lock-in is to avoid those lock-in-prone services.

But what happens then?  For example: a team is told that they can't use hosted elasticsearch, because it's not a common enough service, so they build, host and manage their own elasticsearch.  What they've done is increased the total size of their self-managed infrastructure, which increases the switching costs of moving that infrastructure to another provider.  The whole exercise was self-defeating.

Switching costs need to be viewed as a property of the whole system.  Trying to take a shortcut by making a list of "approved" vs "unapproved" services leads to poor decision making, sprawling and poorly-factored infrastructure, and ultimately greater switching costs and greater lock in.

soupdragon99:
Just added some editorial changes, bullet points etc.

bruntonspall:
:+1:

philandstuff:
rebased to resolve conflicts. I think this is ready to merge now - @bruntonspall want to do the honours?

PR #8, opened 2017-05-05
Add docs on retired applications

alexmuller:


source-code.md
```diff
@@ -23,3 +23,11 @@ Consider encrypting the contents of the repository.
 
 Teams are responsible for ensuring that their Git repositories
 are backed up to another location.
+
+### Retired applications
+
+We have an organisation called [gds-attic](https://github.com/gds-attic/)
+which is used for source code which is no longer in production.
```
rboulton: Agreed to add "for public repos".

We'll write separate guidance for private repos.

rjw1: I've rewritten this to make it obviously about public repos.

PR #9, opened 2017-05-15
Add note on 2FA being enforced on GitHub

alexmuller:


source-code.md
```diff
@@ -28,3 +28,5 @@ Consider encrypting the contents of the repository.
 
 Teams are responsible for ensuring that their Git repositories
 are backed up to another location.
+
+GitHub organisations should have two-factor authorisation enforced.
```
rboulton: Suggest adding "for all users via the GitHub settings", to make it clear this isn't just a policy we tell individual users to follow.

rjw1: UPdated to reflect this.

PR #26, opened 2017-05-30
Add initial guidance on deploying software

alexmuller:


deployment.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Teams should create standalone artifacts which can:
+
+- have automated tests run against them to ensure they work as expected
+- be deployed to a pre-production environment for manual testing
+- be promoted to a production environment
+
+The artifact should contain the running application, including
+any dependencies.
```
whpearson: Can we allow multiple artifacts to refer to each other? And pull in dependent artifacts (a la debian files).

deanwilson: Can we have something that covers being able to cryptographically prove it's the same between environments please?

index.md
```diff
@@ -24,6 +24,7 @@ in this repository. Projects at beta and live should follow these recommendation
 
 ### Operating services
 
+- [Deploying software to production](deployment.html)
```
sihugh: Should these all be alphabetical?

deployment.md
davidillsley: I think 'including any dependencies' is where this gets tricky as it doesn't necessarily make sense to carry *all* dependencies around all the time.

deployment.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Deployment
+layout: recommendation
+expires: 2017-12-01
+---
+
```
rboulton: Describe the need: "so that we can be confident of what code is running, and be able to run it again".

techforevil:
agreed in principle but needs more work (user needs, what is a deployment artifact, dependencies, guidance for different languages, signing, how does it work w/ PaaS). 

will return at the next forum!

deployment.md
dcarley: Related to Richard's and David's comments: it would be good to describe *why* this is desired. Is it for:

- consistency: that one deployment behaves like another?
- security: that dependencies haven't been modified?
- availability: that you can deploy without depending on third-party services?
- or: some combination of the above?

For example, is using a dependency manager that pins the versions of dependencies and sub-dependencies sufficient?

philandstuff:
sorry @alexmuller 😢

alexmuller:
Don't apologise, you've closed my last open GDS pull request. Now if you could go through and do all my open issues too that would be great ;)

PR #37, opened 2017-06-20
Updated logging guidance

philandstuff:
This isn't ready to merge yet.

I set an expiry of 2017-12-01 rather than 2017-12-20 so that it
doesn't expire too close to the christmas break.


philandstuff:
closing, will reopen after more work

PR #43, opened 2017-07-14
bau: add Java styleguide

willp-bl:
Pair: WillPa

### DIFF
--- a/source/manuals/programming-languages.html.md.erb
+++ b/source/manuals/programming-languages.html.md.erb
@@ -38,3 +38,9 @@ We've got a consistent style for:
 - [Ruby](programming-languages/ruby.html), 
 - [Python](programming-languages/python/python.html)
 - [Java](programming-languages/java.html)
+
+Some of the guidelines in the style guides are codified in a
+[.editorconfig](programming-languages/.editorconfig) file. Place a copy of this
+file in your project’s repository to have tooling that supports
+[EditorConfig](http://editorconfig.org/) automatically adhere to the
+guidelines.
--- a/source/manuals/programming-languages.html.md.erb
+++ b/source/manuals/programming-languages.html.md.erb
@@ -33,5 +33,8 @@ Some good reasons to ignore a particular guideline:
 - When the code needs to remain compatible with older versions that
   don't support the feature recommended by the style guide.
 
-We've got a consistent style for [Ruby](programming-languages/ruby.html) and
-[Python](programming-languages/python/python.html) and will add one for Java.
+We've got a consistent style for:
+
+- [Ruby](programming-languages/ruby.html), 
+- [Python](programming-languages/python/python.html)
+- [Java](programming-languages/java.html)
--- a/source/manuals/programming-languages/.editorconfig
+++ b/source/manuals/programming-languages/.editorconfig
@@ -1,0 +1,7 @@
+[*.java]
+indent_style: space
+indent_size: 4
+end_of_line: lf
+charset: utf-8
+insert_final_newline: true
+continuation_indent_size: 8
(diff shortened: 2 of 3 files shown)


philandstuff:
Do you use anything (other than the mentioned IntelliJ code import settings) to enforce code style?  Is Sonar part of the build process? Should it be?

It's nice that the Ruby styleguide is enforced as a set of rubocop configurations. I wonder if something similar might be possible here.

source/documentation/programming-languages/java.html.md.erb
```diff
@@ -0,0 +1,43 @@
...
+
+Variable and field names should match the class they are instantiating, or have a descriptive name in the context of their use.
+
+Use the default code indenting in IntelliJ.
+
+Ensure line feeds are present at the ends of files.  This can be enabled as an option in IntelliJ: `Settings->Editor->Other->Ensure line feed at file end on save.`
+
+## Dependency injection (DI)
+
+Some programmes use [Guice](https://github.com/google/guice) as a dependency injection framework, however, consider whether use of DI is appropriate for your project.
```
mrlumbu: this could be expanded in the future - how might one go about evaluating whether DI is appropriate?

philandstuff:
@alexbishop1 has said he'll look into codifying style in [editorconfig](http://editorconfig.org/) configuration

philandstuff:
rebased to resolve conflicts introduced by #44

alexbishop1:
As promised, I had a look into creating an EditorConfig file based on the Java style guide.

Disappointingly, there’s not a huge overlap between what can be put in an EditorConfig file and what the Java style guide specifies but I came up with the following (the order of the properties matches the EditorConfig documentation):

```
[*.java]
indent_style: space
indent_size: 4
end_of_line: lf
charset: utf-8
insert_final_newline: true
continuation_indent_size: 8
```

The `indent_style`, `indent_size` and `continuation_indent_size` properties match the IntelliJ defaults, per the style guide’s “Use the default code indenting in IntelliJ”. The `insert_final_newline` property matches the style guide’s instruction to “Ensure line feeds are present at the ends of files”. The `end_of_line` and `charset` properties don’t match anything in the style guide but I put them in because it’s what we do anyway.

After deliberately changing the IntelliJ settings so they didn’t match, I put the above in a `.editorconfig` file and dropped it in the root of one of our Java projects. It seemed to work. I also installed the plugins for Eclipse and Sublime Text and tried the `.editorconfig` file with those: both worked (with some quirks) but they don’t support all of the properties.

Looking through the list of other EditorConfig properties, we might want to specify something for `trim_trailing_whitespace`, `max_line_length`, `curly_bracket_next_line`, `spaces_around_operators` and `indent_brace_style`. However, IntelliJ only supports the first of these.

There is a proposal for EditorConfig to have a property for specifying the number of imports required before using a wildcard import but it doesn’t seem to have gained much traction (the proposed property doesn’t even have a name yet). This is a shame, as it’s one of the few places where the style guide differs from the IntelliJ defaults and it would be nice to not have to remind every developer to change the setting manually.

I presume we want to commit this `.editorconfig` file somewhere in this repository so that a developer can take a copy and drop it into their project. I’m not entirely sure where that should be. The spec for EditorConfig plugins states that they should look for `.editorconfig` files in the same directory as the file being edited and every ancestor directory. Therefore, projects will generally want to have their `.editorconfig` stored in the root, with all settings for all file types in the same file (so we’d probably want to have a single GDS Way `.editorconfig` file for everything, rather than one for Java, one for Ruby etc.).

Nooshu:
Building on the "single GDS Way .editorconfig file for everything", [this PR](https://github.com/alphagov/gds-way/pull/107) has been opened to address this issue. Any feedback would be appreciated.

PR #47, opened 2017-07-20
Update logging guidance

philandstuff:
I set an expiry of 2017-12-01 rather than 2017-12-20 so that it
doesn't expire too close to the christmas break.

### DIFF
--- a/source/standards/logging.html.md.erb
+++ b/source/standards/logging.html.md.erb
@@ -18,8 +18,8 @@ archival, and vice versa.
 
 ### Shared Elasticsearch, Logstash and Kibana (ELK) stack
 
-GDS has a shared account with logit, which offers on-demand ELK stacks
-as a service.
+GDS has a shared account with [logit](https://logit.io/), which offers
+on-demand ELK stacks as a service.
 
 This is suitable for a short-term place to store logs in order that
 they are available and queryable.  This should not be considered a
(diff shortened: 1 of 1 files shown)


source/standards/logging.html.md.erb
```diff
@@ -1,31 +1,104 @@
...
-whether there's something you can use.
+In general, tools for querying logs are not a good fit for long-term
+archival, and vice versa.
 
-## Sumo Logic
+### Shared Elasticsearch, Logstash and Kibana (ELK) stack
 
-Some teams have used [Sumo Logic](https://www.sumologic.com/)
-successfully for storing and querying logs.
+GDS has a shared account with logit, which offers on-demand ELK stacks
```
ajlanghorn: Is it worth putting a link to Logit here? I remember Googling for them, and realising that their value wasn't `p` - as such, it was difficult to work out what the log-odds were of me finding their site in the first place.

soupdragon99:
I have read through the guidance and made some minor changes to readability.

source/standards/logging.html.md.erb
```diff
@@ -7,25 +7,98 @@ expires: 2017-09-01
...
 
-## Elasticsearch, Logstash and Kibana (ELK)
+This is suitable for a short-term place to store logs so
+they are available and queryable.  This should not be considered a
+durable store of logs; it is *not* suitable for long-term archival of
+logs.
+
+### Sumo Logic
+
+Some teams have used [Sumo Logic][] successfully for storing and
```
rhowe-gds: Did we intentionally lose the link to the Sumo Logic website?

PR #49, opened 2017-07-28
update the new page template

philandstuff:
This updates the new page template to ensure there's a `<h1>` tag and an
expiry banner.

Because we have erb-in-erb, the double percent sign is needed to
render correctly on the page.  ie

    <%%= partial :expires %>

renders to

    <%= partial :expires %>

which is what we want.


nickcolley:
image

philandstuff:
ooh nice example of Alexander's dark band 👍

PR #54, opened 2017-08-01
add clarification banner

philandstuff:
adds a banner to make it clear that this content is intended for a GDS
internal audience only.

copy by @soupdragon99 

screenshot: 
screen shot 2017-08-01 at 09 53 14

### DIFF
--- a/source/layouts/core.erb
+++ b/source/layouts/core.erb
@@ -46,6 +46,7 @@
 
         <div class="app-pane__content toc-open-disabled">
           <main id="content" class="technical-documentation" data-module="anchored-headings">
+            <div class="page-banner"><p>The GDS Way and its content is intended for internal use by the GDS community.</p></div>
             <%= yield %>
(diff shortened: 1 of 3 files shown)


J-Lambo:
@philandstuff Looks good. Is there a way to add a link to the Service Manual to help users who are looking for guidance that applies to them? https://www.gov.uk/service-manual/technology

Other than that, I think it's clear and helpful.

philandstuff:
@J-Lambo I spoke with @soupdragon99 about this; we felt it would make things more confusing rather than less.

J-Lambo:
As an external user who has been told this guidance isn't for me, my next question would be where to go next. If it's not included in the banner, could it be included on your first page/in introduction?

bruntonspall:
Gut feel, it should point to either the digital service manual, or the tech code of practice.

Otherwise, generally +1

PR #55, opened 2017-08-01
Add links back to the repo

sihugh:
There's a [general call for contributions][1], but no link to the repo.  I've added an inline link and a header link, but happy if people would rather this went elsewhere.

[1]: https://github.com/alphagov/gds-tech/blob/f2cec941568445c146255e11ecd418a7b25b9383/source/index.html.md.erb#L14
screen shot 2017-08-01 at 16 54 05

### DIFF
--- a/config/tech-docs.yml
+++ b/config/tech-docs.yml
@@ -10,6 +10,7 @@ phase: Beta
 # Links to show on right-hand-side of header
 header_links:
   Documentation: /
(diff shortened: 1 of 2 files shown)


sihugh:
I don't have permission to push a branch to this repo which is why the build hasn't kicked off

philandstuff:
this was actually merged after a rebase but github's UI makes it look like it was closed. see isaacs/github#2 😞

PR #56, opened 2017-08-02
deploy app to gds-way and redirect traffic from old app

philandstuff:
This deploys the site to a new app called `gds-way`.  It also sets up
a static nginx at the old app to redirect all traffic to the new URLs.

I tested this locally by deploying the redirect to a site called
`gds-tech-docs-test` and verifying that the redirects worked as
expected.

### DIFF
--- a/redirect/nginx.conf
+++ b/redirect/nginx.conf
@@ -1,0 +1,13 @@
+worker_processes 1;
+daemon off;
+
+error_log <%= ENV["APP_ROOT"] %>/nginx/logs/error.log;
+events { worker_connections 1024; }
+
+http {
+  server {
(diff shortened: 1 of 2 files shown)


redirect/nginx.conf
```diff
@@ -0,0 +1,13 @@
+worker_processes 1;
+daemon off;
+
+error_log <%= ENV["APP_ROOT"] %>/nginx/logs/error.log;
+events { worker_connections 1024; }
+
+http {
+  server {
+    listen <%= ENV["PORT"] %>;
+    server_name localhost;
```
davbo: the listen header doesn't need 0.0.0.0 and having the single server {} makes this the default.

PR #64, opened 2017-08-25
Initial Node.js Styleguide

maxcbc:
Added Initial Node.js Styleguide

@jonheslop — I just made a major overhaul after a few discussions with @maxcbc.
* Decided that rather than write up examples we can just link to [Node.js best practices](https://github.com/i0natan/nodebestpractices) as it's much more thorough and well referenced.
* Styleguide was split across pages, seeing as I reduced the amount of content and there is no breadcrumb, I merged it all into one place to make it easier to navigate (i.e. _just scroll_)
* Left in Max's notes on testing asynchronously and using chai.expect over chai.should as this isn't covered so much in [Node.js best practices](https://github.com/i0natan/nodebestpractices)


source/manuals/programming-languages/nodejs.html.md.erb
```diff
@@ -0,0 +1,118 @@
...
+
+## Import statements and strict mode
+
+Each `.js` file in your application should be in [strict mode], `'use strict'` should be at the top of each file,
+this will enforce best practices and prevent some potentially unexpected behaviours.
+
+Import statements (i.e. `require('x')`) should always be at the top of each file and should be grouped in the following way:
+1) Built-in modules - Modules that come pre-packaged with Node.js such as `http`, `crypto` and `path`.
+2) NPM dependencies - Modules that you have installed via NPM such as `express`, `request` or `lodash`.
+3) Other application files - Other files the application such as `./config/environment.js`
```
vanitabarrett: This doesn't quite make sense, I think you meant "other files the application **uses**" ?

source/manuals/programming-languages/nodejs.html.md.erb
```diff
@@ -0,0 +1,118 @@
...
+This can often (but not always) make for smaller, more easily readable and maintainable code,
+though judgement should be exercised as to when this is the case.
+
+In Node.js applications, it is better to use `const` and `let` rather than `var`.
+
+A useful guide for what ES2015+ features are supported in each version of Node.js can be found at [Node Green].
+
+## Import statements and strict mode
+
+Each `.js` file in your application should be in [strict mode], `'use strict'` should be at the top of each file,
```
nickcolley: If we're using ES6 this is not necessary right? https://stackoverflow.com/questions/31685262/not-recommended-to-use-use-strict-in-es6

maxcbc: Node uses commonJS modules rather than ES6 modules, meaning strict mode is still required.

In a node module without strict mode doing
``` foo = 'bar'```
creates a `foo` property on the global object, i.e `global.foo`, same as like the window object in the browser.

Comment (maxcbc) [reply source/manuals/programming-languages/nodejs.html.md.erb]: Also sorry for taking like 2 months to reply, I got a little distracted. I'm on it now.

File: source/manuals/programming-languages/nodejs/style-guide.html.md.erb [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+
+## Organise your imports
+
+Import your dependencies at the top of each file, and group them into 'core', 'npm' and 'local' dependencies, like in the below example.
+This makes your imports clearer and allows others to  more easily understand and maintain cross dependencies of each file.
+
+```
+'use strict'
+
+// Core Dependencies
```
joelanman: I think these should be sentence case

maxcbc: As in the below?

`// Core dependencies`

joelanman: Yeah, tiny thing but worth it if this will be replicated in every project

maxcbc: Yeah sure. It's the tiny things that matter more often than not.

Will update the PR.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
```diff
@@ -0,0 +1,119 @@
...
+
+
+### let and const
+
+For Node.js code, be explicit about variable mutability using [const] and [let]:
+
+```
+const foo // GOOD: explicitly immutable variable declaration
+let bar // GOOD: expicitly mutable variable declaration
+var baz // BAD: not explicit about mutability of variable
```
Nooshu: Are there cases where the use of var would be applicable? For example you could use a var keyword to convey to a developer that this variable will be used across a whole function, rather than scoping it to a particular block (if / else etc).

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
```diff
@@ -0,0 +1,119 @@
...
+```
+
+This makes code easier to understand, as well as providing a feedback loop if a [const] variable is inadvertently reassigned.
+
+### Promises
+
+For Node.js applications, we use the native `Promise` implementation available in Node.js over alternatives like `q` or `bluebird`.
```
Nooshu: Is it worth adding details about why an alternative was needed and what version of Node.js started to support native Promises? Just for some context around this point.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
```diff
@@ -0,0 +1,119 @@
...
+const config = require('../../config')
+const controller = require('./controller')
+const utilFunction = require('../../utils/util_function.js')
+```
+
+## Avoiding Unnecessary Function Declarations
+
+When building high volume applications in Node, it is important to be conscious of (and kind to) the V8 Engine's garbage collector.
+
+One area to be aware of, in order to maximise performance, is avoiding unnecessary function declarations within another function's scope.
```
Nooshu: Would be worth adding a simple bad / good example of this in code to help aid understanding of the problem.

source/manuals/programming-languages/nodejs/testing.html.md.erb
```diff
@@ -0,0 +1,80 @@
+---
+title: Testing Node.js Apps at GDS
+---
+
+# <%= current_page.data.title %>
+
+## Libraries
+
+At GDS we use Mocha as our test runner and Chai as our assertion library.
+These form part of a larger group of commonly used testing libraries, a non-exhaustive list of libraries we use can be found below.
```
Nooshu: More of a general question, does there need to be any justification over why we use these libraries over others that are available? Should a section be added around libraries that we should definitely avoid since we have had issues in the past?

source/manuals/programming-languages/nodejs/testing.html.md.erb
```diff
@@ -0,0 +1,80 @@
...
+
+However it is important to note that any unintended errors thrown in any asynchronous tests are passed to `done()`.
+This means that the content of the errors can be fed back to a developer, so that the developer is not just presented with generic error text such as: `Error: Timeout of 2000ms exceeded`.
+
+An example of this pattern can be found below.
+
+```
+'use strict'
+const request = require('request')
+const {expect} = require('chai')
```
Nooshu: Is it worth adding a comment here saying that this is a destructuring pattern? For those not that familiar with ES2015+ it may be helpful to read up on this syntax.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
maxcbc: I would still say it's still better to use let to convey this usage, except declared at the top of the function. See below example:
```
function doSomething(someBool) {
	let variableUsedAcrossFunction
	if (someBool) {
		variableUsedAcrossFunction = 'world'
	} else {
		variableUsedAcrossFunction = 'planet earth'
	}
	return `hello ${variableUsedAcrossFunction}`
}
```

joelanman: Also, it is explicit about mutability no? vars can be changed, consts can't. I think the point is more about scope

source/manuals/programming-languages/nodejs/testing.html.md.erb
maxcbc: Added 👍

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
maxcbc: Added bit on when they've been supported from. I didn't want to go too much into the history of it in the style guide. Though I ought to do a blog post on it.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
maxcbc: I need to work on a generic example of this.

philandstuff:
This needs a link from elsewhere so people can navigate to it.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
```diff
@@ -0,0 +1,122 @@
...
+
+A full list of what features are supported in each version of Node.js can be found at [Node Green].
+
+
+### let and const
+
+For Node.js code, be explicit about variable mutability using [const] and [let]:
+
+```
+const foo // GOOD: explicitly immutable variable declaration
```
nickcolley: Since const does not actually guarantee immutability, I've seen people avoid it entirely, what do you think?

rboulton:
Performance platform team should be involved in reviewing this PR.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
Nooshu: Yeah the point is more around scope. It is possible to use a mixture of let and vars in a way that you convey more meaning rather that explicitly using only one or the other.

For example if setup correctly:
- I see a `let` keyword I know that it's usage is restricted to the block it is set (by design).
- I see a `var` keyword I now know that this is being used across the whole function. 

Simple example:
````
// use both let and var together as it means different things
function foo(x,y) {
    
    try {
        // because we have used let, z is bound to this block
        let z = bar( x * 2 );
    } catch(e){
        //..
    }
    
    // we will use this across the function
    var z = x * 4;
    
    if(x > y){
        // scoped to this block
        let tmp = x;
        x = y;
        y = tmp;
    }
    
    // i scoped to this loop
    for(let i = 0;i< 10; i++){
        //...
    }
    
}
````

I know this method is favoured by Kyle Simpson from the "You don't know JS" book series so I just raised as something that may be worth considering.

File: source/manuals/programming-languages/nodejs/style-guide.html.md.erb (hunk 9 above)
Comment (Nooshu) [reply source/manuals/programming-languages/nodejs/style-guide.html.md.erb]: Is it still worth using to convey the meaning that this variable shouldn't be changed to a developer? For example if I see an array or an object set used with `const` I would avoid changing the inner values (even though I can). I know another pattern that tends to be used is uppercase names for `const`.

Comment (brid) [reply source/manuals/programming-languages/nodejs/style-guide.html.md.erb]: I'd also add that in the case of const objects, it's more about intention than reality, as we all know all this blocks is re-assignment.

Might also be worth highlighting the upsides/downsides of let/const in terms of memory use. With const you rely on the garbage collector eventually freeing up memory, so their use is contextual.

**Comment by rossjones** [comment]
A couple of small general comments, I've asked someone more familiar with the performance platform code to review as well.  I doubt very much that we will be changing the existing code to meet the style-guide as it's more under active maintenance than active development.

In the guidance, it would be handy to have something about suggested build tools, gulp/webpack/whatever.  Particularly as some tools like grunt have contributed extensions that are unmaintained, it would be useful for a consistent approach across node apps so that devs don't have to learn a new build tool each week.

As my current pain is related to updating dependencies on spotlight (the performance platform node app) it would be really useful to provide guidance (or better still insistence) that dependency tracking is used (gemnasium or whatever the newer thing is) and how it should be used.  Obviously this isn't node specific but it bites me more often in node than elsewhere.

**Comment by joelanman** [comment]
We use greenkeeper on the prototype kit and Elements

**Comment by pcraig3** [comment]
Yeah, just to add on to what @rossjones already said. 

Performance Platform has some legacy node applications but there is no active development in node. Future projects are probably going to be Ruby or Python. 

The code we are working on now was written before ES6, so it's unlikely to be rewritten. In that sense, we don't have opinions over how to write JavaScript because we're not writing anything new. 

The main problem we're facing is just the anarchic tendency of microservice architecture (as @rossjones mentioned). Our different repos are using different task runners, different testing libraries, slightly different versions of other random dependencies, and various of these repos import other ones. 

Basically it's the central tension that occurs in all centralised/decentralised organisations: how do you empower teams on the ground without losing central co-ordination? 

So, as Ross said, it would be good to be more explicit about things like task runners and util libraries (lodash vs underscore) or even what versions of dependencies like jQuery we want to use, but not in a way that would take agency away from teams. The stuff around testing is good, but you probably don't want to write down every single dependency. It would be nice to just say in a general way that if you want to include a dependency: 
- if nobody else in GDS is using anything similar, then pick whatever makes sense
- if there's a project already using a similar library, do your best to pick that one

But I don't know how you would coordinate this kind of thing other than vigilance and good intentions.

File: source/manuals/programming-languages/nodejs/style-guide.html.md.erb [hunk 10]
```diff
@@ -0,0 +1,122 @@
...
+
+Because with Node.js you control the version of the interpreter you're code is running on,
+you have access to many the latest features of JavaScript, such as:
+
+- [let] and [const]
+- [Arrow functions]
+- [Array.prototype] methods like `Array.map()`, `Array.find()` and `Array.values()`
+- The [spread operator]
+
+However use caution when using cutting edge JavaScript features and be sure to read the [documentation].
```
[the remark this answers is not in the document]
nickcolley: Maybe Stage - 3 candidates are most safe? But nothing lower? 

https://babeljs.io/docs/plugins/preset-stage-3/

If we wanted to be super cautious maybe even stage 3 is too early.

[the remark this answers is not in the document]
nickcolley: I'd propose we avoid even stage-3 candidates and 'track v8'.

So we only use the ES6 features that are available in our current LTS Node version's V8.

This means avoiding transpiling entirely.

See https://zeit.co/blog/next#data-fetching-is-up-to-the-developer

"Our stance of what features we transpile (like async/await) can be summed up as: we target V8's capabilities. Since our goal is code-sharing between server and client, this gives us great performance when executing the code on Node and developing on Chrome or Brave, for example."

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
```diff
@@ -0,0 +1,122 @@
...
+
+# <%= current_page.data.title %>
+
+## General style
+
+As with all JavaScript at GDS, for Node.js applications we use the [standard] style.
+
+## Use strict mode
+
+[Strict mode] should be enabled for Node.js modules, as in the browser this prevents polluting the global scope (in
```
brid: Strict mode is already enabled in ES6 for module code.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
maxcbc: @philandstuff I think this whole section needs clarifying. Basically we were using an anti-pattern on Pay which I'm trying to avoid being repeated here. @brid Is probably better placed to explain.

The pattern basically involved unnecessary functions were being declared, which were used to encapsulate a particular stage in the overall function's execution. While I can understand how this separated particular functionality into more human-readable stages, the inner functions were effectively being used in place of code comments, and had a negative performance impact.

A very-basic example is below, when the function is run, each of the inner functions has to be recreated in memory every time. When this pattern is used for writing multiple functions run along an express.js route, it leads to many unnecessary inner functions being created in memory every time the endpoint is hit. Overtime this can max out the memory so that the garbage collector just can't run, so the app falls over.

```
function doTheThing(someNumber) {
	const timesBy100 = () => {
		someNumber = someNumber * 100
	}

	const findOddness = () => {
		someNumber = someNumber % 2
	}

	const init = () => {
		timesBy100()
		findOddness()
	}

	init()

	return someNumber
}
```

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
maxcbc: @nickcolley I still generally find it helps keep intention clear using `const` over `var` or `let`. But I agree that if we are going to use `const` the style guide should include a reminder (and link to MDN) to how it only guarantees referential immutability for complex types. 

@brid In reality, my general feeling is that if we need to manually reassign variables to a `null` value when we've finished with them in order to help the GC; we're probably doing something wrong elsewhere in the code. I think we should perhaps include a section in the style guide on avoiding unnecessary variable declarations. Helping developers to think about writing performant code as they go along, rather than having to optimise things.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
maxcbc: For ES6 modules, not for commonJS modules.

In a node module without strict mode doing
`foo = 'bar'`
Without declaring a variable first creates a `foo` property on the global object, i.e `global.foo`, same as like the window object in the browser.

source/manuals/programming-languages/nodejs.html.md.erb
maxcbc: Sorry for not replying sooner. I rewrote this section. You we're right, it was just confusingly written. Hopefully it is clearer now.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
brid: Absolutely, although now we are in Node 8.5.x territory it's entirely plausible to start using .mjs, so it would be worthwhile to say that the use of 'use strict' depends on context.

source/manuals/programming-languages/nodejs/style-guide.html.md.erb
DanailMinchev: Being not Node.js expert, coming from other languages (Java) I would like to comment as well:
For me it's more clearer to use `let` on the top of the function, instead of `var` and to be aware that `var` actually is visible in the whole function scope.
I can see advantages of using @Nooshu 's suggestion, but still remembering the use cases for `var` instead of declaring the variables on the top of the function using `let` is more "risky" when the person doesn't know the use cases of `var`.
This is only opinion from non native Node.js developer :)
Thanks

jonheslop:
After a talk with @Nooshu and @maxcbc on Friday we’re going to approach the Node.j styleguide in a different way, so think it would be cleaner to close this PR and open a new one.

Thanks for you contributions and keep an eye out for the new one 👀

PR #65, opened 2017-08-30
ensure we only have a single h1 on the page

philandstuff:
There's already an h1 further up the page, we shouldn't have another
here


philandstuff:
@soupdragon99 if this looks good could you merge?

PR #69, opened 2017-09-27
Import chapter about Licensing from alphagov/styleguides

tijmenb:
[alphagov/styleguides](https://github.com/alphagov/styleguides) has been around since 2012. Some chapters in there are only relevant to the GOV.UK team (the puppet guidance for example). @rubenarakelyan and I are looking to see if we can move parts of it to the [GOV.UK Developer docs](https://docs.publishing.service.gov.uk) and other chapters into The GDS Way™.

This imports the [Licensing chapter from alphagov/styleguides](https://github.com/alphagov/styleguides/blob/master/licensing.md). It was written by @annashipman in https://github.com/alphagov/styleguides/commit/91b15bc2014952a4e399605b8881561c88a1e808.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -63,6 +63,7 @@ decision or standard.
 - [Programming language style guides](manuals/programming-languages.html)
 - [Setting up logging](manuals/logging.html)
 - [Sending logs securely to Logit.io](manuals/logit-io.html)
+- [Licensing](manuals/licensing.html)
 
 ## Adding new guidance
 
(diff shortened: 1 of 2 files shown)


annashipman:
Makes sense to me but I have a question: would it then be removed from alphagov/styleguides? I ask because that guidance is linked to in various places, for example in [the blog post about the benefits of coding in the open](https://gds.blog.gov.uk/2017/09/04/the-benefits-of-coding-in-the-open/) (search for "at GDS we use MIT").

rubenarakelyan:
Yup we'll remove the content of the file in the styleguides repo and just put a link in there to this file.

PR #85, opened 2017-10-20
Remove the logit manual and link to reliability engineering docs

davbo:
Currently we're duplicating this guidance in both places. This removes it from the GDS Way so the reliability engineering docs becomes the default place to learn how to use services managed by our team.

PR #87, opened 2017-10-25
Include nginx.conf in build files

davbo:
Currently our nginx configuration is being ignored as it isn't part of the
files being pushed to CloudFoundry. Since we mostly just serve static files it
wasn't obvious until we're trying to make additional configuration changes.

This copies the nginx.conf into source files so middleman will copy it into the
build directory which is pushed to the PaaS.


davbo:
Once I can get this merged I'll reopen #85 to get the redirect working.

philandstuff:
will this cause redirects to `gds-tech-docs.cloudapps.digital`? i suspect it will :/

PR #89, opened 2017-11-02
Running Services as containers

surminus:
This is intended as a conversation starter to discuss GDS guidance on running containers.

Containers are being used to run lots of different types of services, and there are often times when these services are not appropriate to run on the PaaS (as a GOV.UK example, we run Gemstash and etcd as containers to reduce Puppet code), or where the PaaS is currently not suitable for the service (ie GOV.UK Pay).

This commit serves to lay down some princibles for running your service in this way. I have made some assumptions which I hope starts some further discussion on the guidance we want to provide (especially concerning the use of ECS).

It is explictly not related to development princibles on containers, which should be a separate set of guidance.


source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,51 @@
+---
+title: Running services as containers
+expires: 2018-05-02
+---
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+## Princibles
```
samcrang: Typo

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+
+<%= partial :expires %>
+
+## Principles
+
+This guidance should not be applied to development environments using containers.
+
+Where possible, web applications should use the PaaS as per the [hosting guidance](standards/hosting.html).
+
+It is not recommended to run bare containers on the PaaS, as provided by the [documentation](https://docs.cloud.service.gov.uk/#deploy-a-docker-image-experimental).
```
rjw1: I wonder if the link should be on "not recommended" and lose the "as provided by the [documentation]"

maybe even make it.
Running bare containers on the PaaS is [not recommended] by the PaaS team.

surminus: While the official Cloud Foundry docs do say it is possible, I think currently the PaaS team do not officially support this usage. Running a service without support from the platform is a high risk. If this support changes that this guidance could certainly change.

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+
+A container registry should host your images.
+
+You should use [Dockerhub](https://hub.docker.com/) or [Amazon EC2 Container Registry](https://aws.amazon.com/ecr/).
+
+### Running containers
+
+#### Bare Docker on EC2
+
+You can run services using Docker on an EC2 instance. This is appropriate for services that:
```
grimley517: Grammar query - should 'There is appropriate' be 'There are appropriate' as plural?

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+
+## Principles
+
+This guidance should not be applied to development environments using containers.
+
+Where possible, web applications should use the PaaS as per the [hosting guidance](standards/hosting.html).
+
+It is not recommended to run bare containers on the PaaS, as provided by the [documentation](https://docs.cloud.service.gov.uk/#deploy-a-docker-image-experimental).
+
+If it is not possible to use the PaaS, or the service is not appropriate (such as when an amount of state is required), then guidance should be followed to ensure best practice when running your service as a container.
```
tomskerous: Slightly worried that this could read by a naive user (e.g. me) as "using containers on PaaS is a way of keeping state" - which isn't really true. You shouldn't be keeping state in containers. Could we make that a little bit clearer? Ta!

surminus: Yes, that doesn't read very well, I meant as you described, ie:

"Running a service that requires state is not suitable on the PaaS" 

I'll see if I can reword to make clearer.

dcarley: Are containers the right solution to host somewhere else if state is required? I don't think they are. It might be better to omit the part about state altogether.

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+
+ - Uptime is not the priority
+ - Do not require zero downtime deployments
+ - Can be managed by configuration management
+
+We recommend against running Docker as root.
+
+The service should have monitoring and logging like any other non-container based application.
+
+#### Amazon ECS as a scheduler
```
surminus: I wonder if this should now be updated to [Amazon EKS](https://aws.amazon.com/eks/)?

Kubernetes has an advantage over ECS as it potentially reduces the risk of lock-in and moving costs.

soupdragon99:
@surminus I see this PR has been open for a while, would we be ready to publish this as guidance for the GDS Way after a tech writer review?

Or is there more work required?

stephenharker:
Containers are being successfully deployed already across live services within GDS and the weight of the industry is firmly behind this approach. We should embrace this tech fully. There have been discoveries already on Kubernetes and Docker Swarm, Amazon ECS is in now in production, I don't think there's any question about our running services as containers.

surminus:
I don't think there is a question about running containers per se, the question is about how to ensure we consistently run containers as an organisation, and _not_ to end up running some on PaaS, some in ECS, some on Kubernetes and some on Docker Swarm which makes it difficult to switch between teams.

tlwr:
Probably premature to add EKS to the guidance given that it is still only in invite-only preview in a single region in the US (Oregon) (AFAIK).

Having guidance on how one orchestrates and schedules containers is valuable, given that containers are the industry standard and containers have been around for a while.

I like the guidance already present with respect to ECS. It would be good to open source some of what we have written, which will hopefully be possible soon.

The overhead of setting up ECS is quite low, and the main effort has been bringing our existing infrastructure up to speed with ECS (which would be the same for any scheduler IMO).

soupdragon99:
Thanks, I'll need to review this before we can publish.

source/standards/running-services-as-containers.html.md.erb
alext: The reason we don't recommend using Docker images on CloudFoundry is that doing so ties the availability of your application to the availability of the Docker registry in use (eg Docker Hub). If the registry is unavailable for any reason, your application could fail because the platform will be unable to start new instances of the application. This does not just apply to new deployments, but any event that would cause the platform to reschedule an instance of the application (eg cell failures, platform deployments etc).

soupdragon99:
@pauldougan I have made some changes to this, be good to have your review.

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+* [Docker][] to [build and deploy images][], as it’s widely used across GDS and industry
+* [Dockerhub][] or [Amazon EC2 Container Registry][] to host your images following the [PaaS responsibility model][]
+* the recommended [monitoring and logging solution][]
+
+You can also run your service using bare metal Docker on an [Amazon Elastic Compute Cloud (Amazon EC2)][] instance if:
+
+* service uptime is not a priority
+* zero downtime deployments are not required
+* it can be managed by configuration management
```
soupdragon99: we could put a link here to the configuration management guidance once it's also published

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+You should not run:
+
+* local images on the PaaS, as [deploying bare metal containers][] are not supported
+* Docker containers with root access using the Docker daemon as this has [security implications][]
+
+If your service requires high-availability, you should use the [Amazon Elastic Container Service (Amazon ECS)][] as your container orchestration service. By using [Amazon ECS clustering][] you can protect against unhealthy instances, and balance service workload across corresponding healthy instances.
```
soupdragon99: do we recommend this for all services on the PaaS? are there any services that don't actually require this and what would be an example?

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+If your application is either not currently running on PaaS, or it’s not possible to use PaaS as a platform due to technical constraints, you should...
```
soupdragon99: Can we expand on this, giving some more detail and an example or two.

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+If you are running a development environment using containers then you should…
```
soupdragon99: Is there some guidance or reliable site we can link people to for more information?

surminus:
I'd just like to confirm that even though I raised the PR I will not be amending or expanding as I leave tomorrow :)

source/standards/running-services-as-containers.html.md.erb
pauldougan: you can only deploy docker images from a registry, the cf push command simply does not support it. Its less a question of should not, the cli simply does not support deployment of a local docker image.

source/standards/running-services-as-containers.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+If your web application is running on [GOV.UK Platform as a Service (PaaS)][] you should follow the [recommended hosting guidance][].
+
+If your web application is running on PaaS you should use:
+
+* [Docker][] to [build and deploy images][], as it’s widely used across GDS and industry
+* [Dockerhub][] or [Amazon EC2 Container Registry][] to host your images following the [PaaS responsibility model][]
```
alext: This contradicts the current advice from the PaaS team. Docker support is currently experimental, and not recommended for production services. The current implementation of Docker support in Cloudfoundry has some limitations that mean we don't consider it suitable for production use - https://docs.cloud.service.gov.uk/#deploy-a-docker-image-experimental.

The most significant of these is that it it ties the availability of the application to the availability of the Docker registry being used - if the registry is unavailable, your application could fail at any time.

If deploying to the PaaS we strongly recommend using the buildpack approach over Docker images.

dcarley:
It might be a bit late, and I'm sorry to Laura for the notifications, but..

Should something be included about maintaining container images like there is for [virtual machines](https://gds-way.cloudapps.digital/standards/operating-systems.html#operating-systems-for-virtual-machines)? Do any teams have good advice on the subject?

soupdragon99:
@pauldougan have a meeting next week to work on this guidance, I think there are some more changes to go in here before it's good to go.

Humphaz:
I will have a go of defining some guidance for "Containers" in general, rather than a specific technology. I'll also discuss aspects like composed containers in comparison with container controllers like Docker Swarm or Kubernete's.

soupdragon99:
That's great thanks @Humphaz

PR #92, opened 2017-11-09
Copy edits to naming things

soupdragon99:
Edited content, headings, layout to match with GOV.UK content style.


soupdragon99:
Thanks @36degrees that's a fair point, I had a chat with @davbo and it would be good to broaden this out by reverting back to using 'How to name things' as the title, replace references from 'applications' to software components and include your examples in the guidance.

source/index.html.md.erb
```diff
@@ -37,7 +37,7 @@ Standards are based on formal decisions made by the tech forum. The standards ha
 
 The following standards will help you name your service, choose an appropriate programming language and where to store your code:
 
-- [naming things](standards/naming-things.html)
+- [how to name things](standards/naming-things.html)
```
36degrees: This is pedantry and I don't really care that much, but if we're changing things to try and get a consistent style this doesn't seems to match the other things in this list…

To me, it'd make sense to be either:

- how to name things
- how to choose programming languages
- how to store source code

or

- naming things
- choosing programming languages
- storing source code

:man_shrugging:

soupdragon99:
Have made suggested changes to guidance @36degrees @davbo

source/index.html.md.erb
soupdragon99: I agree, I'm working my way through the published guidance and will update as I go through each piece 😄

soupdragon99:
@36degrees @davbo have made some changes, please see latest commit and my comments, thanks Pete.

source/standards/naming-things.html.md.erb
```diff
@@ -1,53 +1,40 @@
 ---
-title: Naming things
+title: How to name software components
```
36degrees: Should this match the link in index.html.md.erb? ('How to name things')

soupdragon99:
@36degrees @davbo Sorry to be a real pain here.... but after further comms discussions, we are going to need to use 'products' and not 'things' as the guidance has potentially wider visibility than a solely GDS audience, 'software products' being an industry name that covers libraries, plugins etc.

So the title here would be how to name 'How to name software products' and sub title would be 'Your product name should be self-descriptive'

I hope that is OK, thanks again!

soupdragon99:
Changes made following 2i review.

PR #93, opened 2017-11-10
Specify that retired repos should be archived

timblair:
We previously moved retired repositories to the [gds-attic GitHub organisation](https://github.com/gds-attic). GitHub has recently added the ability to [make repos read-only by archiving them](https://github.com/blog/2460-archiving-repositories), which means we can flag those retired repos without the cost of managing a separate organisation.

### DIFF
--- a/source/standards/source-code.html.md.erb
+++ b/source/standards/source-code.html.md.erb
@@ -36,8 +36,8 @@ GitHub organisations should have two-factor authorisation enforced for all membe
 
 ### Retired applications
 
-If an application is no longer used in production its public repository should
-be [archived](https://help.github.com/articles/archiving-repositories/).
+If an application is no longer used in production its repository should be
+[archived](https://help.github.com/articles/archiving-repositories/).
 
 The application's README should be updated to detail why the repository has
 been archived, and should link to a new location if the function of the
--- a/source/standards/source-code.html.md.erb
+++ b/source/standards/source-code.html.md.erb
@@ -36,8 +36,9 @@ GitHub organisations should have two-factor authorisation enforced for all membe
 
 ### Retired applications
 
-If an application is no longer used in production its public repository should be migrated to the [gds-attic](https://github.com/gds-attic/) organisation on GitHub.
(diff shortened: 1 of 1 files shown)


nickcolley:
Would we move gds-attic repos back to alphagov then archive them?

timblair:
@nickcolley: I'd say so, yes.

danmitchell-:
+1

surminus:
I can't tell if this also applies to private repos, but if so that solves a problem as we can't currently move private repos to gds-attic. 👍

timblair:
Current wording is public only, but that was due to the previous limitation on the gds-attic repo. The new process should definitely apply to private repos too. I’ll update the wording.

rboulton:
Do archived private repositories still contribute towards the limit on the number of private repositories allowed in our organisation?  If so, I'd say we should still be archiving private repositories initially, but might need to have some process to remove them fully (archiving out of github, or perhaps to a different paid org with the new account terms which have a cost per user but no limits on private orgs, and only a small list of users).

dcarley:
> Do archived private repositories still contribute towards the limit on the number of private repositories allowed in our organisation? If so, I'd say we should still be archiving private repositories initially, but might need to have some process to remove them fully (archiving out of github, or perhaps to a different paid org with the new account terms which have a cost per user but no limits on private orgs, and only a small list of users).

[About archiving repositories](https://help.github.com/articles/about-archiving-repositories/) says:

> Note: If you have a legacy per-repository billing plan, you will still be charged for your archived repository. If you don't want to be charged for an archived repository, you must upgrade to a new billing plan. For more information, see "Organization billing plans."

From the [alphagov billing settings](https://github.com/organizations/alphagov/settings/billing) looks like we changed from per-repo to per-user, so my understanding is that they won't contribute to our private repo quota.

philandstuff:
@dcarley nope, we're on per-repo

dcarley:
Ah, I misread the "$x per user/month" as the current state rather than something to upgrade to.

PR #98, opened 2017-11-20
Add .editorconfig file with settings specific to FE tech

Nooshu:
Editorconfig mentioned in GDS Way but doesn't exist for this repo. I've added settings for scss, js and erb. Some settings may be able to made globally (using *), then lang specific settings as listed when required.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,9 +3,18 @@
 # top-most EditorConfig file
 root = true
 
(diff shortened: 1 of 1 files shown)


Nooshu:
PR can be closed. Added seperate PR to the tech docs repo:
https://github.com/alphagov/tech-docs-template/pull/121

Issue #99, opened 2017-11-20
Last list anchors broken on mobile / tablet viewport


Nooshu:
To reproduce:
- Goto: https://gds-way.cloudapps.digital/#standards
- Shrink the viewport down to approx 760px
- Try clicking on "storing source code" link
- Link not clickable due to following headings negative margin causing an overlap.

![margin-issue](https://user-images.githubusercontent.com/1223960/33012104-d653bcf8-cdd7-11e7-9e8b-f1baf78f580f.png)

Issue caused [here](https://github.com/alphagov/gds-way/blob/bb6a00f98d1bae9f708fe0e99589be818c3171c8/source/stylesheets/modules/_technical-documentation.scss#L54)
````
h3, h4, h5, h6 {
    @include heading-offset($gutter);
  }
````

Proposed fix:

1. Either remove usage of the mixin in this case and add specific styling manually (less padding, no negative margin)
2. Continue to use the mixin but add an override after it's usage

Option 1 seems like the cleanest to me with minimal impact.

philandstuff:
Thanks for the detailed error report! I've seen something similar before but it's good to see an analysis of the problem, its causes, and ways to fix it. We've added this to our backlog to take a look at it.

Nooshu:
No problem @philandstuff. [PR for option 1](https://github.com/alphagov/gds-way/pull/100) here if the team are happy with the fix. If not just throw it away :)

Nooshu:
Closing issue as it has been duplicated [here](https://github.com/alphagov/tech-docs-template/issues/123).

PR #100, opened 2017-11-20
Remove tech docs header 3-6 mixin, replaced with standard padding

Nooshu:
PR created for option 1 in [issue 99](https://github.com/alphagov/gds-way/issues/99)

### DIFF
--- a/source/stylesheets/modules/_technical-documentation.scss
+++ b/source/stylesheets/modules/_technical-documentation.scss
@@ -52,7 +52,7 @@
   }
 
   h3, h4, h5, h6 {
-    @include heading-offset($gutter);
+    padding-top: 26px;
   }
 
   h3 {


36degrees:
Good spot.

As mentioned in other PRs this will need to be opened against alphagov/tech-docs-template.

This only fixes the issue for links that appear above an h3 or above. Further down the same page, the problem still exists – for example the Licensing link above the 'Adding new guidance' `h2`.

Making the same change there would cause issues with headings being hidden behind the sticky 'table of contents' nav when you jump to them - the `heading-offset` mixin is designed to offset the headings so that when you jump to the anchors they appear in the correct place.

Not sure what the correct fix is here but I think it needs a bit more thinking.

soupdragon99:
@Nooshu @36degrees @pgchamberlin I see [#104 ](https://github.com/alphagov/gds-way/pull/104) has been merged, can we close this PR now or is there more work to be done here?

PR #103, opened 2017-11-20
Adding style guide for Node.js

jonheslop:
After a [false start](https://github.com/alphagov/gds-way/pull/64) here is a second attempt
at a Node.js style guide. After a meeting with @maxcbc and @Nooshu we decided to
outline our preffered Formatting and Linting as using this will bring style inline
without requiring extensive documenation here.

We also outline how Node.js projects should be structured. We think a modular, feature
based structure is best because it emphasises the fact that code should be written in
a modular resuable way, we don't recommend an MVC style structure as this separation by
roles fragments the code and makes it more difficult to read and write.

### DIFF
--- a/source/manuals/programming-languages.html.md.erb
+++ b/source/manuals/programming-languages.html.md.erb
@@ -35,9 +35,10 @@ Some good reasons to ignore a particular guideline:
 
 We've got a consistent style for:
 
-- [Ruby](programming-languages/ruby.html), 
+- [Ruby](programming-languages/ruby.html)
 - [Python](programming-languages/python/python.html)
 - [Java](programming-languages/java.html)
+- [Node.js](programming-languages/nodejs/)
 
 Some of the guidelines in the style guides are codified in a
 [.editorconfig](programming-languages/editorconfig) file. Place a copy of this
--- a/source/manuals/programming-languages/editorconfig
+++ b/source/manuals/programming-languages/editorconfig
@@ -5,3 +5,11 @@ end_of_line: lf
 charset: utf-8
 insert_final_newline: true
 continuation_indent_size: 8
+
+[*.{js,json}]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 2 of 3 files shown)


rboulton:
I've not reviewed this change, but am very happy to see all the work on it.  In terms of process for merging, this is just a change under the "manuals" section of the gds-way, so it doesn't need to wait for a formal approval to merge. Once there's reasonable consensus from the relevant parts of the community, please merge it. It can be discussed further if needed in later PRs. If you need a person to define "reasonable consensus", I nominate @Nooshu

maxcbc:
@rboulton Cool, thanks Richard. Thats was our thinking. @Nooshu you are volunteered as tribute!

maxf:
I agree this is good. But should we go further and tackle the problem of JavaScript having so many ways to write things. Not just at the commonjs level, but also things like `=>` vs `function`, `class` vs `function`, `for` vs `forEach` vs `map`, Promises vs `async`/`await`, TypeScript vs Flow, etc.

Every other language has the same issue, of course. But JS happens to be worse, and this can result in code being very hard to read and modify by devs inheriting a JS codebase (which is the main problem the GDS Way is trying to solve, I think). And I'm not even talking about packages like Ramda that let you write your JS in an almost unintelligible way for people not used to pure functional programming or other paradigms. Should we encourage some of those practices and reject others?

I'm afraid I don't have a PR that answers this, but I think an extra conversation about higher-level abstract patterns is worth having.

nickcolley:
@maxf that sounds like a good idea, could you raise that as an issue so we don't block this initial PR?

jonheslop:
@maxf Yeah we want to define all that stuff but this is sort of an MVP style guide, so we didn’t wanna go deep into all that stuff. I’ve got lots of ideas about the sort of things we should be saying about this sort of thing though. We should start that convo off but @nickcolley is right, we should open a separate issue about adding that stuff in

maxf:
Sure, I didn't mean to block this. I'll open an issue.

nickcolley:
Great work, thanks everyone for getting this over the line 🚀

maxcbc:
Woop woop woop. Well done @jonheslop

PR #104, opened 2017-11-20
Update README to include tech docs template information

Nooshu:
I've added paragraph on how to make functional changes and bugfixes to GDS Way. Rather than making changes directly, fix it in the tech docs template then apply it. From a new starters point of view this wasn't clear to me. I'd been making changed directly to the GDS Way repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,13 @@ generated output.
 Including files manually like this lets us specify the position they appear in
 the page.
 
+## Making functional changes
+
+The GDS Way is built from the [Tech Docs Template](https://github.com/alphagov/tech-docs-template)
+repository. Any functional changes and bug fixes should be made here first, then follow the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +33,13 @@ generated output.
 Including files manually like this lets us specify the position they appear in
 the page.
 
+## Making functional changes
+
+The GDS Way is built from the [Tech Docs Template](https://github.com/alphagov/tech-docs-template)
+repository. Any functional changes and bug fixes should be made here first, then follow the
```
pgchamberlin: In the phrase, "any functional changes and bug fixes should be made here first", it reads like "here" means the GDS Way - because that's the site the user will be reading. Perhaps it should say, "changes and bug fixes should be made to that project first".

Nooshu: Agree @pgchamberlin, updated accordingly.

pgchamberlin:
:+1:

PR #107, opened 2017-11-23
Update the standard editorconfig file

Nooshu:
Added scss, js, json and erb to the standard editorconfig file. Ideally all languages that are used across GDS can be added so it is a simple copy / paste for new projects.

### DIFF
--- a/source/manuals/programming-languages/editorconfig
+++ b/source/manuals/programming-languages/editorconfig
@@ -32,6 +32,15 @@ max_line_length = 119
 indent_size = 2
 trim_trailing_whitespace = true
 
+[*.sh]
+indent_size = 2
+
+[*.yml]
+indent_size = 2
+
+[Makefile]
+indent_style = tab
+
 [**vendor**]
 indent_style =
 indent_size =
--- a/source/manuals/programming-languages/editorconfig
+++ b/source/manuals/programming-languages/editorconfig
@@ -5,26 +5,29 @@
 # uncomment if top-most EditorConfig file
 # root = true
 
-[*.java]
+[*]
 indent_style: space
-indent_size: 4
-end_of_line: lf
-charset: utf-8
-insert_final_newline: true
-continuation_indent_size: 8
-
-[*.{scss,erb}]
 charset = utf-8
 end_of_line = lf
-indent_style = space
+insert_final_newline: true
+
+[*.java]
+indent_size: 4
+continuation_indent_size: 8
+
+[*.{scss,erb,html}]
 indent_size = 2
-insert_final_newline = true
 trim_trailing_whitespace = true
 
 [*.{js,json}]
-charset = utf-8
-end_of_line = lf
-indent_style = space
 indent_size = 2
-insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.py]
+indent_size = 4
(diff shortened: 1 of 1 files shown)


Nooshu:
Added vendor section as requested @selfthinker.

source/manuals/programming-languages/editorconfig
```diff
@@ -1,15 +1,50 @@
...
+charset = utf-8
+end_of_line = lf
 insert_final_newline: true
+trim_trailing_whitespace = true
+
+[*.java]
+indent_size: 4
 continuation_indent_size: 8
+/* this wasn't explicitly set, so setting to false until told otherwise */
+trim_trailing_whitespace = false
```
alexbishop1: This wasn’t set because the GDS Way Java style guide doesn’t say anything about it: https://github.com/alphagov/gds-way/pull/43#issuecomment-323073119

Nooshu: Thanks for the update @alexbishop1, I'll read through those comments.

source/manuals/programming-languages/editorconfig
```diff
@@ -1,15 +1,50 @@
-[*.java]
+# http://EditorConfig.org
+# https://gds-way.cloudapps.digital/manuals/programming-languages/editorconfig
+# Copy into project root and rename to .editorconfig
+
+# uncomment if top-most EditorConfig file
+# root = true
+
+[*]
```
samcrang: I think I'd prefer not to see options applied globally—I suspect it will be confusing or incorrect when applied to certain file types.

source/manuals/programming-languages/editorconfig
```diff
@@ -1,15 +1,50 @@
...
-indent_style = space
 indent_size = 2
-insert_final_newline = true
-trim_trailing_whitespace = true
+
+[*.py]
+indent_size = 4
+max_line_length = 119
+
+[**.rb]
```
samcrang: Is there a reason this is the only file type where options are applied to files under directories? Should the others do this too? Should this do the same as the others?

source/manuals/programming-languages/editorconfig
Nooshu: Good point, it does act as a catch-all so may have unintended consequences

source/manuals/programming-languages/editorconfig
Nooshu: Good catch, updated to match the others

Nooshu:
Definitely a good topic to bring up at the Tech Forum. My stance is that we should be using one for all projects. Ideally only a single file in the root of a project to simplify maintenance. It would be good to get the various different teams to agree to a GDS standard for each language then use this for each new project.

Where we stand on "legacy" applications may be a little more difficult. It could be that the agreed standard is tweaked to fit this application (i.e. the .editorconfig in its root folder) rather than making a large number of changes retroactively.

Nooshu:
After a conversation with @rboulton I propose we merge this PR as is as a first attempt at a global .editorconfig file. If any teams or language experts disagree with anything listed a seperate PR should be raised.

pgchamberlin:
Good plan. Awaiting the :+1: and merge from @rboulton.

PR #110, opened 2017-12-06
Minor text changes - sending email notifications

soupdragon99:
Slight change to syntax of headings and to 'How to send notifications to engineers' section to tighten language.


soupdragon99:
These are the changes we discussed this morning to the last section.

I also removed 'you should' from the section titles to make them easier to scan.

-## How you should send notifications to users
 +## How to send notifications to users

PR #117, opened 2018-01-03
python - add section on dependencies

davidread:
This is based on what we discussed at the python meetup 12/12/2017.

I'm not totally sure this is the right place to put this content, since it's described as a Python 'style guide' in documents that link to it, even though the document itself is just 'Writing python at GDS' with a path suggesting it is for all things python. I'm happy to make it into a separate manual if people prefer?

And please do pitch in with any other improvements.

### DIFF
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -41,6 +41,55 @@ _For more on linting (including standard settings and config) see the [linting][
 * Lint checks using [PyFlakes][]
 
 
+### Dependencies
+
+It is important to have a good strategy for specifying your project's Python
+libraries. There are two desirable characteristics:
+
+1. Reproducibility
+
+    Pin your full dependencies or you'll get unpredictability between your dev
+    environment and other environments. You want a new starter to avoid small
+    hard-to-spot problems. And you want parity between what you test locally,
+    what is tested by CI and what you deploy, or you risk new issues appearing
+    on a live server. Additionaly these things can be hard to diagnose.
+
+    Simply manually maintaining a requirements.txt doesn't capture
+    sub-dependencies. And we've noticed [minor problems with using
+    piptools/pip-compile and pipenv/Pipfile][dm-deps-commit].
+
+    **Recommendation**: Put your top-level requirements in a
+    `requirements-app.txt` file and use [this Makefile
+    script][dm-deps-commit-makefile] to generate a fully specified
+    `requirements.txt`, which is used whenever you need to pip-install the
+    dependencies. The script creates a temporary virtualenv, pip-installs the
+    `requirements-app.txt` and pip-freezes the result as `requirements.txt`.
+
+2. Kept up-to-date
+
+    Security issues are found in libraries, so it is important to choose
+    libraries that are maintained and to ensure your team has a strategy to
+    ensure security updates are installed without significant delay.
+
+    There are several web services which check a project's python dependencies
+    daily. When new releases are available, the service automatically creates a
+    Pull Request to update your requirements file, prompting the team's typical
+    processes to review, merge and deploy. A couple of GDS teams have found that
+    keeping up with every single update can be difficult - a daily nag that one
+    is soon inclined to start ignoring. At least two of these web services (snyk
+    and pyup) track which of the updates are for security reasons (as opposed to
+    just bugfixes/features) so there is scope to try getting PRs only for these.
+    However it relies on their database being correct about which updates are
+    security related, and we've spotted at least one case where a security fix
+    was missed.
+
+    **Recommendation**: Trial a service, such as [requires.io][], [pyup.io][],
+    [dependabot.com][] or [snyk.io][]. Ideally try two and compare them. Try
+    different settings. Report findings back to GDS's Python group. We're keen
+    to track this quickly emerging area to maximize our security assurance of
+    public services.
+
+
 ### Updating this manual
 
 This manual, and by extension the [GDS Python Style Guide][GDSPSG], is not presumed to be infallible or beyond dispute.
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -41,6 +41,55 @@ _For more on linting (including standard settings and config) see the [linting][
...
+
+It is important to have a good strategy for specifying your project's Python
+libraries. There are two desirable characteristics:
+
+1. Reproducibility
+
+    Pin your full dependencies or you'll get unpredictability between your dev
+    environment and other environments. You want a new starter to avoid small
+    hard-to-spot problems. And you want parity between what you test locally,
+    what is tested by CI and what you deploy, or you risk new issues appearing
```
samuelhwilliams: Going to have to nitpick for an Oxford comma here please: 'what is tested by CI, and what you deploy' 😛

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -41,6 +41,55 @@ _For more on linting (including standard settings and config) see the [linting][
...
+    on a live server. Additionaly these things can be hard to diagnose.
```
samuelhwilliams: 'Additionally'

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -41,6 +41,55 @@ _For more on linting (including standard settings and config) see the [linting][
...
+
+    Simply manually maintaining a requirements.txt doesn't capture
```
samuelhwilliams: This sentence might be better phrased as 'Manually maintaining a direct list of requirements in `requirements.txt` doesn't capture sub-dependencies, and doing that yourself is not desirable.'

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -41,6 +41,55 @@ _For more on linting (including standard settings and config) see the [linting][
...
+    sub-dependencies. And we've noticed [minor problems with using
+    piptools/pip-compile and pipenv/Pipfile][dm-deps-commit].
+
+    **Recommendation**: Put your top-level requirements in a
+    `requirements-app.txt` file and use [this Makefile
```
galund: If anyone other than just Digital Marketplace are also using `requirements-dev.txt` then we could probably also mention that here too (i.e. dependencies for test/development not needed for production).

source/manuals/programming-languages/python/python.html.md.erb
davidread: 👍

source/manuals/programming-languages/python/python.html.md.erb
davidread: 👍

source/manuals/programming-languages/python/python.html.md.erb
davidread: Good spot. I've tried to make it clearer by splitting it into two sentences:

    Traditionally a `requirements.txt` lists the app's immediate dependencies,
    but that leaves sub-dependencies unpinned. It makes sense to use a tool to
    generate these, but we've noticed [minor problems with using
    piptools/pip-compile and pipenv/Pipfile][dm-deps-commit]

source/manuals/programming-languages/python/python.html.md.erb
davidread: It's such a common thing to do and it seems reasonable to standardize on the filename. requirements-dev.txt gets 10k results on Github, compared to 6k results for dev-requirements.txt, so I'll add your suggestion.

source/manuals/programming-languages/python/python.html.md.erb
samuelhwilliams: 👍

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -41,6 +41,60 @@ _For more on linting (including standard settings and config) see the [linting][
...
+    Pin your full dependencies or you'll get unpredictability between your dev
+    environment and other environments. You want a new starter to avoid small
+    hard-to-spot problems. And you want parity between what you test locally,
+    what is tested by CI, and what you deploy, or you risk new issues appearing
+    on a live server. Additionally these things can be hard to diagnose.
+
+    Traditionally a `requirements.txt` lists the app's immediate dependencies,
+    but that leaves sub-dependencies unpinned. It makes sense to use a tool to
+    generate these, but we've noticed [minor problems with using
+    piptools/pip-compile and pipenv/Pipfile][dm-deps-commit].
```
davbo: Given Pipfile seems to be the future way to solve this (or at least the linked commit says that) could we say here why those minor problems stop us adopting it now?

This will help future reviewers of this doc understand if we can revisit this decision.

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -41,6 +41,60 @@ _For more on linting (including standard settings and config) see the [linting][
...
+
+    **Recommendation**: Put your top-level requirements in a
+    `requirements-app.txt` file and use [this Makefile
```
davbo: If we do want to recommend using this Makefile is there a better way to share it than linking directly to a commit which changes it?

source/manuals/programming-languages/python/python.html.md.erb
galund: (Caveat: I didn't try Pipfile myself, and @allait would be better placed to comment, however...)

The linked commit explains how dependencies coming from packages installed from version control aren't being locked. This would be a pretty major problem, in fact.

Perhaps we should change the wording and just say "some problems" rather than "minor problems" :)

I didn't find an issue in github relating to that particular problem, though, so maybe we should do some further testing and get one raised.

Either way, Pipfile's readme is caveated fairly heavily with a warning "This project is under active design and development. Nothing is set in stone at this point of time." -- so I think Alexey's feeling that it isn't "ready for production" is right.

source/manuals/programming-languages/python/python.html.md.erb
galund: I definitely think that a "GDS Way" template python project would be a nice thing to have...

source/manuals/programming-languages/python/python.html.md.erb
davidread: @davbo Thanks for this. Admittedly the purpose of the link is not clear, so I've made it clear that it is the details.

I've also removed 'minor' - the severity is debatable.

Let's hope pipenv fix this issue and we can look again at adopting it. We agreed to review all this in 6 months, which has reminded me to add a line at the top of the section:

> The tools mentioned in this topic are changing quickly, so this section would benefit from review in July 2018

I've not added an expiry for the whole python document (as the GDS Way standards do) because the other manuals don't have an expiry and I don't think the choice of PEP8 and linting changes much.

source/manuals/programming-languages/python/python.html.md.erb
davidread: Sounds good, but I'd encourage this to be a separate PR for someone.

davidread:
I've asked again, but no more comments, so I think this is good to merge. I don't have rights - @davbo are you happy to do the honours?

PR #119, opened 2018-01-12
First draft of Node.js guidelines 

maxf:
The goal of this PR is to start a discussion about writing Node guidelines

- Most of the proposed content is arbitrary. Everything is open for discussion
- There may be a need for more sections
- There may be a need for going deeper in some sections, in particular about the syntax

### DIFF
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -4,43 +4,67 @@ title: Using Node.js at GDS
 
 # <%= current_page.data.title %>
 
-## Important Note
+**Important Note**
 
-New projects at GDS should never use Node.js without prior consultation with the 
-GDS Way Forum. Topics like this are discussed at our monthly meetings. To contact 
-the forum [email](mailto:programme-technical-leaders@digital.cabinet-office.gov.uk) explaining 
-the need and linking to whatever else might be relevant. The [programming language recommendations](../../../standards/programming-languages.html#javascript) 
-section has more detail on how we currently use Node.js at GDS and why this 
+New projects at GDS should never use Node.js without prior consultation with the
+GDS Way Forum. Topics like this are discussed at our monthly meetings. To contact
+the forum [email](mailto:programme-technical-leaders@digital.cabinet-office.gov.uk) explaining
+the need and linking to whatever else might be relevant. The [programming language recommendations](../../../standards/programming-languages.html#javascript)
+section has more detail on how we currently use Node.js at GDS and why this
 decision has been made.
 
-## Guidance
-The guidance set out here should be followed in conjunction with the advice on the main [programming languages manual page](../../programming-languages.html).
+## Introduction
 
-If you want to add guidance on Node.js please see [Updating this manual](#updating-this-manual) below.
+The guidance set out here should be followed in conjunction with the advice on
+the main [programming languages manual page](../../programming-languages.html).
 
-### Style
-#### Formatting and linting
-To maintain a consistent and readable JavaScript codebase we use [StandardJS]. This includes a built in formatter, linter and automatic code fixer (auto-fix). To make sure everything stays clean we recommend running this tool before each commit. It is possible to setup a pre-commit hook to automate this process. When [StandardJS] automatically fixes errors it has clear and concise messaging and it also tells you when an automatic fix isn't possible.
+The advice here is specifically about Node.js, and general guidance on writing
+code is out of scope.
 
-We’ve encapsulated our recommended process into a separate [repository](https://github.com/alphagov/nodejs-starter) where everything is controlled by a few lines of config within your `package.json` file. Once you have run `npm install`, any time you make a commit, [StandardJS] runs. It will then format and lint your code automatically. [StandardJS] will auto-fix obvious errors then continue with your commit. If there’s something it can’t fix, then the commit fails with a useful error message such as:
+If you want to add guidance on Node.js please see [Updating this
+manual](#updating-this-manual) below.
 
-```
-❯ Running tasks for *.js
-   ✖ standard --fix
-     → standard: Use JavaScript Standard Style (https://standardjs.com)
-     git add
-✖ standard --fix found some errors. Please fix them and try committing again.
-  ~/nodejs-starter/index.js:5:9: 'hello' is assigned a value but never used.
+## Node versions
 
-```
+> Only use Long Term Support (<abbr title="Long term support">LTS</abbr>)
+versions of Node.js.
 
-#### Editor plugins
-It’s useful to highlight lint errors before starting the commit process. To do this you can get live linting in your text editor of choice by installing one of [Standard’s plugins](https://standardjs.com/#are-there-text-editor-plugins). Users of [SublimeLinter](http://www.sublimelinter.com/) will find [StandardJS] will work out of the box using the default config.
+These are even numbered versions (e.g Node 6.x or 8.x). However, it is important
+to keep an eye on the [Node.js LTS Schedule] as to when versions move in and out
+of LTS.
 
-### Structure
-As well as having a consistent code style, it is important that our projects and repositories are structured in similar ways. This allows developers to find things quickly and easily. We want our code to be modular and reusable so our project structure should reflect this. Below are our four guiding principles: (thanks to [Rising Stack](https://blog.risingstack.com/) for this [blog post](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/)).
+> Do not use the `--harmony` or in-progress feature flags
 
-#### Organise files around features, not roles
+*Staged* or *in-progress* features aren't considered stable by the Node implementors.
+
+> Don't use language features beyond EcmaScript2016
+
+Even though ES2017 is final, some features aren't yet implemented in LTS versions. Check using [node.green](http://node.green/).
+
+## Source formatting and linting
+
+> Use the JavaScript Standard Style ([StandardJS])
+
+StandardJS includes a built in formatter, linter and automatic code fixer
+(auto-fix). To make sure everything stays clean we recommend running this tool
+before each commit. It is possible to setup a pre-commit hook to automate this
+process. When [StandardJS] automatically fixes errors it has clear and concise
+messaging and it also tells you when an automatic fix isn't possible.
+
+We’ve encapsulated our recommended process into a separate
+[repository](https://github.com/alphagov/nodejs-starter) where everything is
+controlled by a few lines of config within your `package.json` file. Once you
+have run `npm install`, any time you make a commit, [StandardJS] runs. It will
+then format and lint your code automatically.
+
+Most editors can check StandardJS syntax through plugins that in some cases comes pre-installed.
+Check the [editor plugin list](https://standardjs.com/#are-there-text-editor-plugins).
+
+## Project directory structure
+
+
+> Organise files around features, not roles
+
 The following structure means you don’t require lots of context switching to find related functions and your `require` statements don’t need overly complicated paths.
 
 ```
@@ -56,7 +80,8 @@ The following structure means you don’t require lots of context switching to f
 |   └── user.njk
 ```
 
-#### Don’t put logic in index.js files
+> Don’t put logic in index.js files
+
 Use these files to `require` all the modules functions.
 
 ```js
@@ -64,12 +89,13 @@ Use these files to `require` all the modules functions.
 const product = require('./product')
 
 module.exports = {
-  create: product.create
+create: product.create
 }
 ```
 
-#### Store test files within the implementation
-Keeping tests in the same directory makes them easier to find and more obvious when a test is missing. Keep the projects global test config and setup scripts in a separate `test` directory.
+> Store test files within the implementation
+
+Keeping tests in the same directory makes them easier to find and more obvious when a test is missing. Keep the project's global test config and setup scripts in a separate `test` directory.
 
 
 ```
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -80,24 +117,321 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+
+- [Node.js Project Structure Tutorial](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/) on [Rising Stack](https://blog.risingstack.com/)
+
+## Language constructs
+
+### Declarations
+
+> Use `const` and `let`, avoid `var`
+
+Embrace immutability (more below) and don't let hoisting confuse you.
```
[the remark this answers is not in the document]
Nooshu: This [article](http://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html#declarations_names_and_hoisting) looks promising.

Nooshu:
Excellent work @maxf. Great starting point to get feedback from the wider community.

jonheslop:
Thanks @maxf this is brill, just a note on formatting, if you add “js” after the opening of a code block you get syntax highlighting. Example — https://github.com/alphagov/gds-way/pull/119/files#diff-2bfc83a0e827145cef9db0e81de317f4R96 — which in the gist has highlighting

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -80,24 +117,321 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+servers.
+
+## Transpiling
+
+> Stick to JavaScript
+
+Avoid anything that compiles to JavaScript (except for static type checking, see
+below). Examples to avoid include CoffeeScript, PureScript and [many others](https://github.com/jashkenas/coffeescript/wiki/List-of-languages-that-compile-to-JS).
+
+> Use TypeScript
```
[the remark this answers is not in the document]
Nooshu: Maybe this should be changed to something along the lines of:

> If you require static typing consider using TypeScript. This has the advantage of... [list advantages]. But please be aware that there are disadvantages too... [list disadvantages]

Rather than a recommendation it could offer an opinion and list of considerations to be made on a project by project basis.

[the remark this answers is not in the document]
nickcolley: I spoke to Max and he recognised that he's not had a chance to use Typescript in a project yet, we should just remove this for now until people have had a chance to try it out and see how it works.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -80,24 +117,321 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+- Reduce the number of dependencies in your applications to a minimum
+- Use tools like [Snyk](https://snyk.io/) to check packages for vulnerabilities
+
+> Use `npm init`, `npm start`, `npm script`, etc.
+
+Making full use of NPM's features will simplify your continuous integration and deployment.
+
+> Lock down dependencies
+
+Avoid incompatible upgrades that may break your application. NPM 5 does it by
```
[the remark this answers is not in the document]
maxf: I've added something on greenkeeper. By pinning I assume you mean locking down?

source/manuals/programming-languages/nodejs/index.html.md.erb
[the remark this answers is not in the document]
maxf: added

source/manuals/programming-languages/nodejs/index.html.md.erb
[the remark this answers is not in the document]
maxf: good points. I've modified the text accordingly.

source/manuals/programming-languages/nodejs/index.html.md.erb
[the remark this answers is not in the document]
nickcolley: https://before-you-ship.18f.gov/infrastructure/pinning-dependencies/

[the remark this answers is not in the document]
maxf: Since that article was written npm has addex package.lock.json, which automatically pins down versions, without the need for `npm shrickwrap`.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -4,41 +4,76 @@ title: Using Node.js at GDS
...
-It’s useful to highlight lint errors before starting the commit process. To do this you can get live linting in your text editor of choice by installing one of [Standard’s plugins](https://standardjs.com/#are-there-text-editor-plugins). Users of [SublimeLinter](http://www.sublimelinter.com/) will find [StandardJS] will work out of the box using the default config.
+> Don't use language features beyond EcmaScript2016
 
-### Structure
-As well as having a consistent code style, it is important that our projects and repositories are structured in similar ways. This allows developers to find things quickly and easily. We want our code to be modular and reusable so our project structure should reflect this. Below are our four guiding principles: (thanks to [Rising Stack](https://blog.risingstack.com/) for this [blog post](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/)).
+Even though ES2017 is final, some features aren't yet implemented in LTS versions. Check using [node.green](http://node.green/).
+
+## Source formatting and linting
+
+> Use the JavaScript Standard Style ([StandardJS])
```
paroxp: Did we decide on using [XO](https://github.com/sindresorhus/xo)? Or perhaps the [AirBnb](https://github.com/airbnb/javascript)?

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -4,41 +4,76 @@ title: Using Node.js at GDS
...
+
+StandardJS includes a built in formatter, linter and automatic code fixer
+(auto-fix). To make sure everything stays clean we recommend running this tool
+before each commit. It is also possible to setup a pre-commit hook to automate this
+process. When StandardJS automatically fixes errors it has clear and concise
+messaging and it also tells you when an automatic fix isn't possible.
+
+We’ve encapsulated our recommended process into a separate
+[repository](https://github.com/alphagov/nodejs-starter) where everything is
```
paroxp: Would it be worth to point to alphagov/gds-nodejs-boilerplate and collaborate with Jon on the "everything is controlled"?

maxf: Yes, I made some fixes on it yesterday and will point to it in this doc

source/manuals/programming-languages/nodejs/index.html.md.erb
maxf: Jon has sent me changes I'm going to make to the page to mention we follow xo

paroxp:
The guide doesn't mention any testing principles. I would be interested what other teams are doing. Is it worth putting a section together?

For instance, is `tape` common in GDS? Or should we stick with `mocha` and `chai`?

maxf:
@paroxp Maybe. We're going to have to discuss if we mention any specific advice on testing at an upcoming meeting.

maxf:
I have made updates from comments received: 
- use XO (text provided by @jonheslop )
- mention the boilerplate repo

I'm still unclear about versions, whether we recommend stopping at ES2016, or instead any specific node version.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -4,43 +4,75 @@ title: Using Node.js at GDS
...
+Even though ES2017 is final, some features aren't yet implemented in LTS versions. Check using [node.green](http://node.green/).
 
-### Structure
-As well as having a consistent code style, it is important that our projects and repositories are structured in similar ways. This allows developers to find things quickly and easily. We want our code to be modular and reusable so our project structure should reflect this. Below are our four guiding principles: (thanks to [Rising Stack](https://blog.risingstack.com/) for this [blog post](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/)).
+## Source formatting and linting
+
+> Lint your JavaScript code with [XO]
+
+We use the JavaScript style used by [XO] to make all our codebase use the same standard. However we make two exceptions: we disallow semicolons and we mandate spaces over tabs. Your `package.json` file should [reflect these exceptions](https://github.com/alphagov/nodejs-starter/blob/master/package.json#L24).
```
Nooshu: I would like to propose that we allow semicolons as the TC39 is now [recommending there usage](https://github.com/tc39/ecma262/pull/1062). It is huge PR with a tonne of comments so it is a controversial topic. As we look to be moving away from Standard this seems like the ideal time to make this change too.

I wouldn't expect this to be a retroactive change, but it should be used for new projects going forwards.

Nooshu: In terms of tabs vs spaces I would like to stick with spaces. Spaces is what is used across all languages at GDS (according to the [.editorconfig in the GDS way)](https://github.com/alphagov/gds-way/pull/107/files) and is what we currently use for JavaScript so I see no reason to change this.

There isn't an outside body recommending any particular stance so this is purely down to GDS developer preferences. It would be good to keep it consistent between teams and languages.

paroxp: Worth to mention: `xo --fix` to be your friend.

paroxp: I agree. It looks like spaces are being more widely adapted in general guides from any big company such as AirBnb or Mozilla.

Nooshu:
Any more input on this PR? It would be good to get this merged as it seems to have stalled at the moment.

maxf:
`Don't use language features beyond EcmaScript2016` doesn't match the discussion we had, the outcome of which was along the lines of "it's ok to use any feature supported by the LTS version you're using". It would be good to have an official decision on this. (CC @nickcolley who supported the current wording).

Nooshu:
I'm in favour of "it's ok to use any feature supported by the LTS version" as long as we a nailing down the supported node version in the package.json. If we are using Node 8.x.x features we need to be defining it in the "engines" key. Note that "engineStrict" is now depreciated so this will only give the user a warning rather than fail completely.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -4,43 +4,75 @@ title: Using Node.js at GDS
...
+## Node versions
+
+> Only use Long Term Support (<abbr title="Long term support">LTS</abbr>)
+versions of Node.js.
+
+These are even-numbered versions (for example, Node.js 6.x or 8.x). However, it is important
+to keep an eye on the [Node.js LTS Schedule] as to when versions move in and out
+of LTS.
+
+> Don't use the `--harmony` or in-progress feature flags
```
maxcbc: Agree we should add something about flags. I think the line perhaps should be more of a catch-all, to cover anything like --experimental features e.t.c

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -82,24 +116,339 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+servers.
+
+## Transpiling
+
+> Stick to JavaScript
+
+Avoid anything that compiles to JavaScript (except for static type checking, see
+below). Examples to avoid include CoffeeScript, PureScript and [many others](https://github.com/jashkenas/coffeescript/wiki/List-of-languages-that-compile-to-JS).
+
+If you would like to use static typing consider using JavaScript extensions like
```
maxcbc: I disagree. We shouldn't encourage typescript or flow. It will fragment Node.js usage. IMHO we should just use plain Javascript and write tests.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -82,24 +116,339 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
 ```
 
-#### Use a config directory
-Keep your configuration together in a config folder.
 
+See also:
+
+- [Node.js Project Structure Tutorial](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/) on [Rising Stack](https://blog.risingstack.com/)
+
+## Language constructs
```
maxcbc: I was convinced when we added the first version that language conventions are probably not something we should stipulate in the GDS way. 
1) It's too prescriptive as the language develops.
2) Its too divisive to agree on
3) The development community at large on the interweb can probably do it better

By stipulating here we lose the advantage of having a prescriptive linter like `standards` or `xo` and end up with meetings about tabs, spaces, semicolons, const, let, var etc.

An exception is choice of libraries, such as `we use express 4.x as our server framework`, `we use mocha, chai and sinon as our testing framework`

maxcbc: That being said, there is pretty much nothing I disagree with here in terms of chosen style.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -82,24 +116,339 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+want to use external libraries.
+
+Generally, anything that departs from the standard JavaScript syntax in a way
+that a Node developer would have trouble reading your code is advised against.
+
+## Frameworks
+
+> Use Express for web applications, avoid lesser-known frameworks
+
+Express is very common for writing web applications that talk to back-end
```
maxcbc: We shouldn't necessarily explain each choice in detail. I think its sufficient to say.
```
## Frameworks and Modules
We standardise on common frameworks and modules across GDS to maximise share-ability of code between projects and familiarity for developers across the GDS community.

### Server Framework
We use express.js 4.x as our server framework.
```

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -82,24 +116,339 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+}
+
+// Over:
+for (let i=0,j=0;i<10 && j<10;j++,i=(j==10)?i+1:i,j=(j==10)?j=0:j,console.log(i,j)){}
+```
+
+
+## Node Package Manager (<abbr title="Node package manager">NPM</abbr>)
+
+> Limit your reliance on external code, and check the NPM packages you use
```
maxcbc: I don't think this is good advice. The power of Node is that there is NPM behind it. IMHO I think the advice should be the opposite.
``` Don't re-invent the wheel, the NPM registry has packages to help solve a lot of common problems. Leverage existing open-source code where possible to maximise your productivity```

Followed by some warnings and advice on how to judge modules prior to using them.

File: source/manuals/programming-languages/nodejs/index.html.md.erb [hunk 12]
```diff
@@ -82,24 +116,339 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+Avoid incompatible upgrades that may break your application. NPM 5 does it by
+default, but be careful when upgrading or adding a new package. Tools like
+[Greenkeeper](https://greenkeeper.io/) can be useful to automatically check
+dependencies as part of your continuous integration.
+
+> Regularly check for unused dependencies, and make sure you don't have dev dependencies in production
+
+Don't slow down your deployments with unused code.
+
+> If you build modules you think could be useful for others, publish them on npm
```
maxcbc: 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

maxcbc:
@Nooshu I think we should have a policy that we support the latest and latest -1 LTS (i.e. 8.x, 6.x at the moment) versions in modules. And say that you should be using one of these in applications and that you should make use of the features version you are using.

source/manuals/programming-languages/nodejs/index.html.md.erb
paroxp: Maybe it shouldn't be encouraged but neither should it be discouraged. I'd still like to consider TypeScript on JS projects in general inside the team.

source/manuals/programming-languages/nodejs/index.html.md.erb
paroxp: Although I agree with `Don't re-invent the wheel` ideology, I'm not a big fan of importing big library just to use one small function within it.

maxcbc: I don't disagree, but I've only found this a significant issue on the client rather than server side. The vast majority of NPM libraries in my experience are pretty small (<3 files). Notable exceptions  include `lodash`, but that can be imported at a granular level i.e: `lodash/get`.

paroxp: Sounds fair.

source/manuals/programming-languages/nodejs/index.html.md.erb
maxcbc: My concern would be around hiring. Typescript shrinks the pool of available developers, its sort of the same logic around why we don't use Scala/Kotlin for JVM development.

I'd probably go so far as to say that if we need type safety, we should use a type-safe language like Java e.t.c, it's better at it and can have other benefits . Give it a year or so and chances are we'll get some isomorphic benefits there too via progressively-enhancing-WASM usage.

Nooshu:
@maxcbc that seems fair to me. So should a developer be actively testing that the application runs on both versions? 

Supporting LTS - 1 then becomes the lowest common denominator. So you won't be able to use the latest bleeding edge functionality right away, and by the time you do get to use it (when LTS becomes LTS - 1) it will be well and truly tested / bug fixed.

Nooshu:
Adding this link @nickcolley posted as it may be useful for this discussion in terms of technical application: [adding node as a dependency in your package.json](https://twitter.com/housecor/status/962347301456015360)

nickcolley:
@Nooshu we're using `nvm` for the time being which is probably a better bet for now.

maxcbc:
@Nooshu I don't think there's any need for an application to support both, except when teams are going through the process of upgrading. 

However when it comes to re-usable modules I would say definitely yes, they should test on both, because they may be reused on teams building apps using either. This can easily be tested in a CI pipeline i.e. using something like Travis.

paroxp:
I agree with Max. 

It's beneficial for MODULES to be tested and supported on different version due to team's adaptation of node. For instance, if they're not planning to upgrade soon-ish, ideally they'd have the trust in a module not to break due their delay.

The APPLICATION doesn't need that, as it's developed by that team anyway.

source/manuals/programming-languages/nodejs/index.html.md.erb
maxf: Typescript is similar enough to JS that we shouldn't have a hiring problem, as long as we don't require applicants to know it. It's very easily picked up. I'm aware it's a slippery slope to mention any sort of transpiling to JS and I'm ok not to mention typescript or others, but I wouldn't prevent it from being used. (I'd go as far as to say that weak typing is a thing of the past, which is no longer justified in modern software engineering.)

source/manuals/programming-languages/nodejs/index.html.md.erb
maxf: I think there is a place for recommendations that sit between style (space/tabs, comments, wrapping, etc) and frameworks (mocha, express, etc). I called it language constructs and while I'm all for copying existing conventions like xo and airbnb, they don't cover those constructs. It's difficult to strike a balance as to what to recommend and it could indeed lead to many discussions, but there's virtue in saying use class for classes, or async/await for promises, etc.

source/manuals/programming-languages/nodejs/index.html.md.erb
maxf: Agreed. I'll change that.

source/manuals/programming-languages/nodejs/index.html.md.erb
maxf: I'll change that to general advice that's less opinionated.

source/manuals/programming-languages/nodejs/index.html.md.erb
maxf: will change

source/manuals/programming-languages/nodejs/index.html.md.erb
jonheslop: I’m ok with everything in the _Language constructs_ section, I think it’s a good outline of sensible approaches that are good for consistency

👍

pgchamberlin:
@Nooshu if the Node community are happy then we should get @soupdragon99 to review the content and get it merged imo.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -4,43 +4,79 @@ title: Using Node.js at GDS
 
 # <%= current_page.data.title %>
 
-## Important Note
+This document describes how we write Node.js code at GDS. It is a list of
+guidelines that developers should follow in order to standardise code across all
```
pgchamberlin: "Make code more consistent" might be better than "standardise"

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -4,43 +4,79 @@ title: Using Node.js at GDS
 
 # <%= current_page.data.title %>
 
-## Important Note
+This document describes how we write Node.js code at GDS. It is a list of
+guidelines that developers should follow in order to make code more consistent
```
pgchamberlin: 👍

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -82,24 +120,343 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+> Use `const` and `let`, avoid `var`
+
+Embrace immutability (more below) and don't let
+[hoisting](http://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html#declarations_names_and_hoisting)
+confuse you.
+
+
+### Functions
+
+> Prefer `function` for top-level function declarations and `=>` for function literals
```
arnau: Is it worth mentioning that if you have a legitimate case for diverging from this suggestion because both have different scoping (so they are not really equal alternatives) it is ok to do so?

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -82,24 +120,343 @@ Keeping tests in the same directory makes them easier to find and more obvious w
...
+
+```js
+// Use:
+const foo = function (x) {
```
arnau: What is the reason for preferring this form over a named function `function foo(x) { ... }`?

maxf: It's to avoid any confusion that would be caused by hoisting.

arnau: Ok, that's what I thought 👍 . Is it worth making a note about it either in this section or in the previous one?

arnau: Ah apologies, I just saw the comment about not using this PR for further comments.

PR #125, opened 2018-02-16
Add guidance for configuration management

surminus:
I have written this and suspect that most people would argue for opening up the guidance to just say "use configuration management" rather than pinning on specific tools. However, I would argue that using a variety of tools makes managing different environments more difficult.

This guidance recommends Puppet for larger environments, and Ansible as an acceptable choice for smaller environments.

It should also take into account not using config management for relatively simple deployments where something like cloudinit is all that is needed.

### DIFF
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -1,0 +1,24 @@
+---
+title: Configuration Management
+expires: 2018-05-16
+---
+
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+When running VMs, such as in Amazon EC2, where the complexity of the configuration
+in the instance becomes too great for managing during deployment or within AMIs,
+use configuration management.
+
+Using configuration management ensures that the configuration of an environment is
+documented as code.
+
+## Tools
+
+In large environments [Puppet][] should be used.
+
+In smaller environments, [Ansible][] is an acceptable choice.
+
+[Puppet]: https://puppet.com/
+[Ansible]: https://www.ansible.com/
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -7,14 +7,17 @@ expires: 2018-06-30
 
 <%= partial :expires %>
 
-You should use [configuration management][] to manage, automate and standardise your infrastructure. When using configuration management you store your infrastructure as code, such as your virtual machines (VM) environment configuration.
+You should use [configuration management][] to manage, automate and standardise your infrastructure. When using configuration management you store your [infrastructure as code][], such as your virtual machines (VM) environment configuration.
 
 For example, you can use configuration management during your VMs deployment when running them in [Amazon Elastic Compute Cloud (Amazon EC2)][] or in your [Amazon Machine Image (AMI)][]. This ensures build consistency and provide an audit trail of what was changed.
 
-We recommend using [Puppet][] for configuration management, as it’s scalable and widely used. You can find out more about configuration management in the [Service Manual][].
+We recommend using [Puppet][] for configuration management, as it’s scalable and widely used.
+
+You can find out more about configuration management in the [Service Manual][].
 
 
 [configuration management]: https://puppet.com/solutions/configuration-management
+[infrastructure as code]: https://www.gov.uk/service-manual/technology/manage-your-software-configuration#use-infrastructure-as-code
 [Amazon Elastic Compute Cloud (Amazon EC2)]: https://aws.amazon.com/ec2/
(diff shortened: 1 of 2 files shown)


tuzz:
I think this is good 👍 

A bit of standardisation would make it easier to share expertise across projects and make on-boarding to a new project go more smoothly. My (perhaps naive) impression of the various config management tools is they all achieve similar things and most of the time it doesn't matter which you use, so being a little prescriptive seems reasonable for the benefits brought about by consistency.

I was wondering if it would help to give an example / rough estimate of what constitutes a 'big' or 'small' project. Does 'small' mean one server, running one app, with one database? It might also be useful to link to successful projects using Puppet or Ansible to give a reference people can use as a starting point if they'd like to follow other conventions for structure.

source/standards/configuration-management.html.md.erb
```diff
@@ -0,0 +1,24 @@
...
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+When running VMs, such as in Amazon EC2, where the complexity of the configuration
+in the instance becomes too great for managing during deployment or within AMIs,
+use configuration management.
+
+Using configuration management ensures that the configuration of an environment is
+documented as code.
```
rhowe-gds: We should explicitly state that this code should be treated as per other code and kept in source control, with some statement about paying special attention to any sensitive/secret data (either managing secrets entirely separately, or if no suitable mechanism exists perhaps storing them alongside the code, ideally encrypted)

source/standards/configuration-management.html.md.erb
```diff
@@ -0,0 +1,24 @@
...
+expires: 2018-05-16
+---
+
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+When running VMs, such as in Amazon EC2, where the complexity of the configuration
+in the instance becomes too great for managing during deployment or within AMIs,
+use configuration management.
```
rhowe-gds: Maybe not a statement for the first iteration of this guidance because I suspect it is somewhat controversial but I would like to have a statement that the consideration should be given to ensuring that where possible responsibilities of configuration management should be clearly delineated against the responsibilities of any application deployed on the instance. Close coupling of application behaviour and configuration management systems can lead to a complex service implementation which is difficult to evolve over time.

pgchamberlin:
Good point from @tuzz that examples of what constitutes large and small environments would be helpful, but overall I think :+1: and we can iterate and improve.

source/standards/configuration-management.html.md.erb
pgchamberlin: Yeah, that's a really good point. I agree that's probably overkill for the first pass guidance too.

source/standards/configuration-management.html.md.erb
pgchamberlin: Missed this comment before :+1:'ing this PR. We shouldn't duplicate the content from the `how to store source code` section though. Maybe we could add a link to that?

soupdragon99:
Thanks I'll need to review this before we can publish.

source/standards/configuration-management.html.md.erb
```diff
@@ -0,0 +1,25 @@
...
+
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+You should use configuration management to track and control changes in large software deployments. When using configuration management you ensure your environment configuration is documented and [stored as code][].
+
+For example, when you run virtual machines (VM) in [Amazon Elastic Compute Cloud (Amazon EC2)][] and the VMs instance configuration becomes too complex to manage either during its deployment or within your [Amazon Machine Image (AMI)][] use configuration management.
+
+You should use [Puppet][] in large software environments, and for smaller environments, you can use [Ansible][].
```
soupdragon99: I'd echo Chris's comment, we could with an example of what large and small means in this context.

source/standards/configuration-management.html.md.erb
```diff
@@ -0,0 +1,25 @@
...
+expires: 2018-05-16
+---
+
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+You should use configuration management to track and control changes in large software deployments. When using configuration management you ensure your environment configuration is documented and [stored as code][].
+
+For example, when you run virtual machines (VM) in [Amazon Elastic Compute Cloud (Amazon EC2)][] and the VMs instance configuration becomes too complex to manage either during its deployment or within your [Amazon Machine Image (AMI)][] use configuration management.
```
soupdragon99: We could do with saying why your configuration would become too complex, and under what circumstances. What do we mean when we say something's 'complex' in this context?

soupdragon99:
draft guidance submitted for 2i.

soupdragon99:
@stephenharker be good to have your review on this.

stephenharker:
This looks good to me as an initial commit. We can refine and enhance from this base.

Issue #134, opened 2018-04-12
Clarity about including Open Government Licence in repositories


idavidmcdonald:
https://gds-way.cloudapps.digital/manuals/licensing.html#use-mit says that all repositories should have a `LICENCE` file and that we use the MIT licence.

https://gds-way.cloudapps.digital/standards/publish-opensource-code.html#how-to-publish-open-source-code says that we use the MIT licence for code and code samples in documentation and Open Government Licence (OGL) for documentation.

Some questions I'd like clarity on (and then I'm happy to try and improve the documentation based on the answers):

1) Should absolutely all repositories contain a `LICENCE` file? For example the gds-way repository itself does not have a LICENCE file. Note, if you build the gds-way docs then there is a link at the bottom of all pages to the OGL licence. Is that sufficient rather than having an actual `LICENCE` file in the repository?

2) For documentation repositories, should we therefore be including two licences (MIT and OGL)? Does this differ if the repository is the documentation (e.g. https://github.com/alphagov/monitoring-doc) versus the repository gets built into an actual website serving documentation (such as the gds-way)?

soupdragon99:
The answer to your questions is basically 'yes' we should have an MIT licence file in the GDS Way repo, all repositories should be licensed.

We should also use the OGL licence for documentation, as you point out this is visible once a doc site is built out of the template, but it should also be mentioned in the README along with the MIT licence for clarity.

One of my colleagues put some guidance together around this as it was a question others have asked. I'll send you a link to the doc.

idavidmcdonald:
Many thanks.

Shall I put in a PR to update the GDS way to try and clarify this (whilst also adding the required licences)?

idavidmcdonald:
I've started a PR by adding the licences to the gds-way repo itself. Hopefully I've done it correctly given that hopefully this repository can contain instructions but also be an exemplar for how to do licensing. I haven't started on any of the documentation (although happy to do so if you think that is sensible).

soupdragon99:
#135 approved, merged and closed.
Thanks @idavidmcdonald

idavidmcdonald:
I'd like to reopen this actually Peter. Whilst the licence has now been added I still feel we could improve the documentation around knowing what to add. I'll put something together based on some of the questions you answered above if you wouldn't mind reviewing after please.

soupdragon99:
No problem at all, happy to help.

idavidmcdonald:
https://github.com/alphagov/gds-way/pull/137

[...]

PR #135, opened 2018-04-13
Add Licence

idavidmcdonald:
Related to Issue https://github.com/alphagov/gds-way/issues/134.

Worth double checking that links are going to the correct locations as I may have made assumptions


soupdragon99:
@philandstuff I took the liberty of adding you as a reviewer. This looks fine to me, but a dev review would be good, thanks.

PR #139, opened 2018-04-24
Add standard for tracking technical debt

boffbowsh:
This is based on GOV.UK’s RFC for tracking technical debt:
https://github.com/alphagov/govuk-rfcs/blob/master/rfc-069-classifying-and-measuring-tech-debt.md 
and accompanying blog post: https://insidegovuk.blog.gov.uk/2018/02/19/classifying-and-measuring-tech-debt-in-gov-uk/

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -44,6 +44,7 @@ The following standards will help you name your service, choose an appropriate p
 - [how to store source code](standards/source-code.html)
 - [how to manage third party software dependencies](standards/tracking-dependencies.html)
 - [how to optimise frontend performance](standards/optimise-frontend-perf.html)
+- [how to track technical debt](standards/technical-debt.html)
 
 ### Operating services
 
--- a/source/standards/technical-debt.html.md.erb
+++ b/source/standards/technical-debt.html.md.erb
@@ -1,0 +1,67 @@
+---
+title: How to track technical debt
+expires: 2018-11-01
+---
+
+# <%= current_page.data.title %>
+
+<%= partial :expires %>
+
+While planning for the technical work we do in GDS, we need a common language between technologists and product people to describe the scale and scope of technical debt. The language should enable communication of the benefits of individual pieces of work that pay down tech debt. It should also enable us to be clearer about the consequences of not doing something, or leaving a piece of work in a certain state. It should also allow us to track the reduction or accumulation of technical debt over time. It should not add significant overhead to planning, and assigning a rating to a piece of technical debt should be quick and simple.
+
+### Example consequences of tech debt
+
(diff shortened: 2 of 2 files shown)


soupdragon99:
Thanks @boffbowsh I'll put some time aside to review this week.

boffbowsh:
@heathd @thomasalanlee @louzoid-gds @michaelabenyohai can you review this please?

@pgchamberlin can you review or nominate someone from RE to take a look please?

Happy to set up a discussion about this sooner than the next GDS Way Forum, or we can discuss there.

thomasalanlee:
Looks great, wish I'd seen this sooner.

boffbowsh:
@heathd @louzoid-gds @michaelabenyohai any thoughts from you?

boffbowsh:
I'll close this for comments on Monday 21st so @soupdragon99 can start work on tidying this up. Silence == agreement ;)

boffbowsh:
Last call for this.

louzoid-gds:
Looks good to me!

bleach:
This looks very workable and I think there's value in all teams doing the same thing, particularly so the product management community get used to a consistent way of accounting for it. 

The text is perhaps a bit unclear on how the prioritisation works though. These two parts don't seem 100% consistent, one says high risk does not equate high priority for doing something, the other uses the risk registry as an analogue and calls attention to the fact that they do.

> This [risk] rating signifies the risk and associated cost with not dealing with the item right now. In some circumstances (such as an upcoming planned rewrite or retirement) we would accept a high risk as the item would go away naturally anyway. For this reason, the risks don't necessarily map directly to priorities.

> [referring to risk registers as an analogue] The higher the overall rating, the more the programme should be concerned about it.

I think there should be an overall rating and the tech debt registry should be a list ordered by the tech leadership team's priority to address it; that's what I think is most useful to product managers. The text currently suggests that the overall rating reflects risk, but that the priority is affected by other factors.

I'd say that if something is expected to go away soon anyway, then the risk is reduced by being timebound. So one option is to carry on calling it an overall risk and to call out other factors that reduce or increase the overall risk (eg. we know that roadmap item x is likely to require us to modify this code, so the fact that it's a mess means there is more product development risk). Or just consistently call it an "overall rating" and remove text that suggests it's a risk.

Again, I really think this is a good step forward and I've been recommending people start following it while it's been in draft. :clap:

michaelabenyohai:
This looks good to me!

boffbowsh:
I'm going to merge this as-is, if other teams try Graham's approach and find it useful, they can propose an amendment.

boffbowsh:
@soupdragon99 can you approve this please?

PR #152, opened 2018-07-04
Add standard - How to optimise frontend performance

soupdragon99:
New guidance for frontend performance.

### DIFF
--- a/source/standards/optimise-frontend-perf.html.md.erb
+++ b/source/standards/optimise-frontend-perf.html.md.erb
@@ -22,7 +22,7 @@ For example:
 ||Compress static resources|Use [minification][] or [GNU Gzip][] to compress CSS and JavaScript code|
 ||Set correct Headers|Set correct [Cache-Control][] and [ETag][] headers on assets for optimal caching|
 |Medium|Look for empty image `src` attributes|Avoid using empty image `src` attributes as some browsers always send requests to them, resulting in additional traffic|
-||Minimise DNS lookups|Use fewer domains to reduce the number of DNS lookups per page|
+||Minimise DNS lookups|Use fewer third-party domains to reduce the number of DNS lookups per page|
 ||Maximise parallelisation|Serve static assets from a different subdomain so browsers can download more assets in parallel|
 ||Investigate [lazy loading][]|For pages with many images, only load images in the immediate browser viewport|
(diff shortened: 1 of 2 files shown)


source/standards/optimise-frontend-perf.html.md.erb
```diff
@@ -0,0 +1,103 @@
...
+For example:
+
+|Priority|Example|Action
+|---|---|---|
+|High|Position styles correctly|Set styles at the top and scripts at the bottom of a webpage|
+||Minimise HTTP requests|Minimise tiling icons, CSS and JavaScript files to reduce size and loading time|
+||Compress static resources|Use [minification][] or [GNU Gzip][] to compress CSS and JavaScript code|
+||Set correct Headers|Set correct [Cache-Control][] and [ETag][] headers on assets for optimal caching|
+|Medium|Look for empty image `src` attributes|Avoid using empty image `src` attributes as some browsers always send requests to them, resulting in additional traffic|
+||Minimise DNS lookups|Use fewer domains to reduce the number of DNS lookups per page|
```
Nooshu: I think this should say "third-party domains" else it could be seen as a contradiction to the point below.

source/standards/optimise-frontend-perf.html.md.erb
```diff
@@ -0,0 +1,103 @@
...
+||Maximise parallelisation|Serve static assets from a different subdomain so browsers can download more assets in parallel|
+||Investigate [lazy loading][]|For pages with many images, only load images in the immediate browser viewport|
+||Investigate the impact of loading multiple [@font-face][] assets|Investigate @font-face assets when dealing with common issues like [FOUT, FOIT and FOFT][]|
+|Low|Set images and sprites correctly|Set images and sprites horizontally as it’s easier for browsers to parse|
+||Reduce cookie size|Because every cookie is sent with each HTTP request, consider using a cookie-free domain for static assets|
+||[AJAX][] requests using JSON|Avoid adding too much data to a JSON Object because this causes performance errors with parsing|
+||Investigate using [WebSockets][]|Consider using WebSockets rather than [XMLHttpRequest][], because an HTTP request packet has 1,684 bytes of overhead, compared to 8 bytes for a WebSocket packet|
+
```
Nooshu: We should add in using a service worker here:

||Investigate using a [Service Worker][]| A service worker has the ability to cache critical assets on a users machine. These assets can then be retrieved at a later date from the local filesystem rather than being transferred over the network, thus improving performance|

[Service Worker]: https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API

PR #153, opened 2018-07-04
bump expiry date for Alerting

soupdragon99:
Propose extending the guidance for six months, the guidance also needs updating.

### DIFF
--- a/source/standards/alerting.html.md.erb
+++ b/source/standards/alerting.html.md.erb
@@ -43,7 +43,7 @@ to be shared while they're being worked on.
 
 Dashboards or information radiators can be useful to show the current
 status of alerts, but they should not be the primary alerting mechanism.
-If you use them, ensure that you have a single authoritative source
+If you use them, ensure you have one reliable source
 rather than multiple dashboards.
 
 People respond better to their preferred alerting method. Use a tool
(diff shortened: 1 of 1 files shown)


danielroseman:
I've pushed some changes which address the main issues I had with this page.

source/standards/alerting.html.md.erb
```diff
@@ -61,14 +63,15 @@ track alerts, escalate them (automatically or manually), acknowledge them
...
 
-The default tool at GDS to be alerted of problems with your service
-is [PagerDuty](https://www.pagerduty.com/).
+Your monitoring system should use [PagerDuty](https://www.pagerduty.com/)
+to send alerts about problems with your service.
 
-We recommend using [Pingdom](https://www.pingdom.com/) as well
-in case PagerDuty becomes unavailable.
+We recommend using [Pingdom](https://www.pingdom.com/) as well, as a 
+top-level check on your service's availability.
```
soupdragon99: what do we mean by a top-level check?

soupdragon99:
I think this guidance will need a full review, but let's get the date extended and this tweak in. 
@danielroseman could you approve this?

source/standards/alerting.html.md.erb
danielroseman: Can't think of a better way to phrase this - the point is that all Pingdom does is request a page on your website to check it's actually up, it doesn't do any detailed monitoring.

PR #155, opened 2018-07-04
bump expiry date for How to manage access to your third-party service accounts

soupdragon99:
Proposed new expiry date


soupdragon99:
@stephengrier could you give this a quick review and OK the extended date?
Thanks.

PR #156, opened 2018-07-04
bump expiry date - Operating systems for virtual machines

soupdragon99:
Proposed new expiry date


soupdragon99:
@stephenharker @heathd could you give this a quick review and OK the extended date?
Thanks.

PR #157, opened 2018-07-09
Fix gzip reference

galund:
Some arguably better references for frontend optimisation.


source/standards/optimise-frontend-perf.html.md.erb
```diff
@@ -30,7 +30,7 @@ For example:
 ||Reduce cookie size|Because every cookie is sent with each HTTP request, consider using a cookie-free domain for static assets|
 ||[AJAX][] requests using JSON|Avoid adding too much data to a JSON Object because this causes performance errors with parsing|
 ||Investigate using [WebSockets][]|Consider using WebSockets rather than [XMLHttpRequest][], because an HTTP request packet has 1,684 bytes of overhead, compared to 8 bytes for a WebSocket packet|
-||Investigate using a Service Worker| A service worker caches critical assets on a users machine. These assets can be later retrieved from the local filesystem instead of being transferred over the network improving performance|
+||Investigate using a service worker script| Consider using a service worker to cache critical assets on users machines instead of transferring them over the network
```
philandstuff: "service worker" is the correct term of art.  "Service worker script" is more likely to cause confusion.

It probably doesn't need to be capitalized (like it was in the original) though.  I think "Investigate using a service worker" is probably correct here.

Reference: https://developers.google.com/web/fundamentals/primers/service-workers/

Nooshu:
Thanks for the links @galund, looks good to me!

PR #162, opened 2018-07-20
mark the GDS Way docs as INTERNAL instead of ALPHA

willp-bl:


timblair:
We discussed this in the Forum today, and decided that we were just going to remove the badge entirely. Points raised were:

* Adding an `INTERNAL` badge has caused problems in the past with the GOV.UK Dev Docs when people in other departments thought we were accidentally publishing something that should be private. We then got chased a number of times by IA about this.
* The content in the GDS Way is already useful and relevant. There may be gaps in information, but people should already be using it. An alpha or beta label is not relevant.
* We already have a note on the first page saying that this how _GDS_ does things, and that it's only published for anyone else who might be interested in seeing how we do things. It's not guidance for other departments.

PR #163, opened 2018-07-23
Incidents

idavidmcdonald:
## Adds link to GDS Incident Management Framework document
I believe I've found the correct file using Google Drive search but please could someone with knowledge of incident management confirm it is the correct one as there are several other files named in a similar way.

## Adds link to incident report
I've spent time myself looking for the official GDS wide incident report and seen others do the same so feel this is useful to include. Again, I'd like someone to confirm that is the correct incident report (I took it from the above mentioned framework).

I'd also like to include a link to a good example incident report. There may be one in the guidance contain in the incident report template but the link only works for GOV.UK team members (I'm looking to get that fixed at the moment).

## Further question:
Should the Support Operations page be split into two pages, one covering incident management and the other covering what support operations is, zendesk and support operations dashboard? It feels like this page is currently covering two topics and I wouldn't know to look for support operations if I want to deal with an incident. Probably looking for a tech writer or someone in support ops to decide if this is sensible. I can raise a Github Issue for this if so.

### DIFF
--- a/source/standards/supporting-services.html.md.erb
+++ b/source/standards/supporting-services.html.md.erb
@@ -7,7 +7,7 @@ expires: 2018-08-01
 
 <%= partial :expires %>
 
-The Support Operations team help GDS service teams resolve technical incidents. The team uses the [Information Technology Infrastructure Library (ITIL)][] and the GDS Incident Management Framework to restore normal operations after an incident and minimise the impact on users.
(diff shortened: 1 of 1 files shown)


soupdragon99:
@idavidmcdonald thanks for putting this together, when the Support Ops page was written it was intended as an overview of the team not strictly to provide detailed incident management procedures, hence it's high-level.

My first thoughts are that if we had a separate piece of incident management guidance it would be focused on high-level guidelines for what GDS teams should do as opposed to operational specifics on 'how to x' 

I've looped in @awmassa from Support Ops for his input on this and on linking out to the incident management docs, it's Support Ops who maintain the documents.

PR #166, opened 2018-07-27
fix link to paas tech docs

philandstuff:
PaaS's tech docs have moved from a single-page to a multi-page
format.  This means our existing links based on URL fragments don't
work any more.

I grepped the whole repo for `cloud.service` to find links to the tech
docs and changed every link that needed changing.


rhowe-gds:
`200 OK`

PR #167, opened 2018-07-31
fix slack links

philandstuff:
we've changed instance.


philandstuff:
I did check all these channels still exist on the new instance, by the way 😄

PR #172, opened 2018-08-21
added a section on checking readiness for production

chrisholmes:
This is a simple bit of guidance that teams can use to help check whether their services, or new parts of services, are ready to be used by users.

The focus on ensuring that the service is supportable and that other teams are informed in advance.

### DIFF
--- a/source/standards/production-readiness.html.md.erb
+++ b/source/standards/production-readiness.html.md.erb
@@ -1,5 +1,5 @@
 ---
-title: How to check if your service is production ready
+title: Check if your service is ready for production
 last_reviewed_on: 2018-09-01
 review_in: 6 months
 ---
--- a/source/standards/production-readiness.html.md.erb
+++ b/source/standards/production-readiness.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Check if your service is ready for production
-last_reviewed_on: 2018-09-01
+last_reviewed_on: 2018-10-18
 review_in: 6 months
 ---
 
--- a/source/standards/production-readiness.html.md.erb
+++ b/source/standards/production-readiness.html.md.erb
@@ -1,5 +1,5 @@
 ---
-title: How to know when your service is ready for production
+title: How to check if your service is production ready
 expires: 2019-03-01
 ---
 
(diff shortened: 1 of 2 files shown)


soupdragon99:
Made some simple changes to this to bring it in line with GOV.UK style etc.

source/standards/production-readiness.html.md.erb
```diff
@@ -1,28 +1,50 @@
...
+* alert people when the service has problems
+* follow a documented deployment process
+
+You must also:
+
+* have agreement to proceed from [Information Assurance (IA)][]
+* inform PERSON A
+* inform GROUP B
+
+## Further reading
```
soupdragon99: Is there anything we could add in here that would help people?

source/standards/production-readiness.html.md.erb
```diff
@@ -1,28 +1,50 @@
...
-- It MUST be possible to search logs 
-- It MUST be possible to monitor the health of the service
-- It MUST be possible to view application metrics
-- It MUST be possible to be alerted when the service has problems
-- It MUST be possible to follow a documented deployment process
-- You MUST have agreement to proceed from IA
-- You MUST inform PERSON A
-- You MUST inform GROUP B
-```
+To avoid problems when your service goes into production you should use a checklist to confirm whether your service is ready. Each check should reference a standard that's already in place before you can take any service into production. **For example…**
```
soupdragon99: It would make this stronger if we can give an example here and in the next paragraph.

source/standards/production-readiness.html.md.erb
```diff
@@ -1,28 +1,50 @@
...
+Before taking your service into production and introducing it to live users make sure it meets the relevant technical standards set by GDS and your team.
 
-You can use a checklist to confirm whether your service is ready for production. Checks in the list should reference a standard that must be in place before you can take any service into production. You can also include communication checks to ensure that specific groups or individuals are informed in advance.
+Some of these standards are provided in the GDS Way but your team may also have its own specific standards. For example, your team may require that all its new services are connected to [Logit][] before going to production.
 
-The checklist should be reusable so that your team can apply it every time you wish to take a new service, or subcomponent, into production. 
+If you do not implement a standard before people start using your service then you could have difficulties operating your service or supporting it when a fault occurs. **For example…**
```
soupdragon99: see comment below re: and example or two, nothing in depth, just a brief mention would suffice.

soupdragon99:
@chrisholmes are there any changes you want to make?

source/standards/production-readiness.html.md.erb
```diff
@@ -0,0 +1,48 @@
...
+
+You must also:
+
+* have agreement to proceed from [Information Assurance (IA)][]
+* inform PERSON A
+* inform GROUP B
+
+## Further reading
+
+To find out more about making sure your service is production ready, you can read:
```
soupdragon99: Not sure if we need this, but it would be helpful if there are any other resources people could read.

soupdragon99:
@chrisholmes unless you have any objections this is good to go

PR #180, opened 2018-09-13
Remove specific app plans from Scala advice

sihugh:
This doesn't add much to the point about our problems supporting Scala apps in future, and I'm not even sure it's true any more.


sihugh:
I don't have permission to merge this, so I'd appreciate someone doing this who can 🙏

PR #184, opened 2018-09-24
Update golang guidance

tlwr:
Add Golang to core languages section to reflect reality


willp-bl:
Could you split the Node and Go parts of this PR into separate PRs please?

alext:
> Once this PR is merged we will need to add a golang styleguide

There is already one here: https://github.com/alphagov/styleguides/blob/master/go.md, although it's a little old now, and some of the guidance around (eg dependency management tools) could do with updating.

PR #185, opened 2018-09-24
Update node guidance

tlwr:
- Remove notice about Node.JS being immature because it is ~9 years old
- Add note and context about TypeScript in the frontend languages section

### DIFF
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -43,19 +43,6 @@ not be used to implement any significant ‘application logic’. For instance
 GOV.UK Pay has created thin, client-facing applications that do not store
 data (although they may retrieve data from an API).
 
-In the past we've had problems operating and maintaining products written in
-Node.js (for example maintaining dependencies, picking the appropriate
-libraries to use, understanding the appropriate application structure, and
-ensuring security of the code). We suspect that this is because we have fewer
-developers with Node.js skills than other tools we use.
-
-Node.js is now frequently used in the wider software development ecosystem for
-building web application frontends. This increase in Node.js use means there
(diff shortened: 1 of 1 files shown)


source/standards/programming-languages.html.md.erb
```diff
@@ -31,7 +31,11 @@ The service manual has information on
 [manual_js]: https://www.gov.uk/service-manual/technology/using-progressive-enhancement
 
 GDS projects that use Node.js include:
-GOV.UK Frontend, GOV.UK Pay, Performance Platform and GOV.UK Platform as a Service.
+
+- GOV.UK Frontend
+- GOV.UK Pay
+- Performance Platform
+- GOV.UK Platform as a Service (TypeScript)
```
richardTowers: Verify are also using TypeScript in the form of https://github.com/alphagov/passport-verify, although this is a client library used by other bits of Government rather than something we run directly, so maybe it doesn't count (I can see a world where we provide PHP client libraries but don't support PHP as a GDS way language).

MatMoore:
Can we also remove this paragraph?

> In the past we've had problems operating and maintaining products written in
Node.js (for example maintaining dependencies, picking the appropriate
libraries to use, understanding the appropriate application structure, and
ensuring security of the code). We suspect that this is because we have fewer
developers with Node.js skills than other tools we use.

The skills we have available at GDS is changing over time and unless we can point to something specific and detailed (like a blog post or postmortem), I don't think this warning actually help teams to make decisions about whether to use Node or not. I could easily replace Node.js with any of our core languages here.

tlwr:
I think we can remove that paragraph and the 2 following paragraphs as they do not make sense without the first. Thoughts?

PR #189, opened 2018-10-02
Replace linting XO with StandardJS

nickcolley:
This pull request follows discussion with @jonheslop who was one of the original contributors of this guidance.

With new joiners starting the question has been asked "How should I be linting my JavaScript?",
at the moment we have two answers depending on if it's Node.js or client-side JavaScript.

This change consolidates it down to one choice which reflects what the majority of teams at GDS are doing to lint their JavaScript.

As of writing, there are no live services at GDS that are using XO linted code:

1. GOV.UK Pay client side / server side JavaScript uses StandardJS
2. GOV.UK Publishing client side JavaScript uses StandardJS (based on the styleguide only)
3. GOV.UK PaaS are using TSLint with TypeScript
4. GOV.UK Design System client side / server side JavaScript uses StandardJS

Explained in with the [guidance for the linting 'Why standard'](https://github.com/alphagov/styleguides/blob/master/js.md#why-standard),
StandardJS was introduced originally to avoid bikeshedding type discussions around specific rules, and I think the same applies here in that I don't personally prefer StandardJS's rules, but rather I prefer consistency.

I have also removed the references to the nodejs-starter application as I think it's need is fulfilled by the gds-nodejs-boilerplate application but I'm happy to change this, let me know.

We will also need to update https://github.com/alphagov/gds-nodejs-boilerplate to reflect this change.

cc @alphagov/gds-frontend-developers

### DIFF
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -42,24 +42,9 @@ Use the [engines](https://docs.npmjs.com/files/package.json#engines) key.
 
 ## Source formatting and linting
 
-> Lint your JavaScript code with [XO]
+> Lint your JavaScript code with [StandardJS]
 
-We use the JavaScript style used by [XO] to make all our codebase use the same
-standard. However we make one exception: we mandate spaces over tabs. Your
-`package.json` file should [reflect this
-exception](https://github.com/alphagov/nodejs-starter/blob/master/package.json#L24). Using
-`--fix` helps with fixing issues automatically.
-
-We’ve encapsulated our recommended process into a separate
-[repository](https://github.com/alphagov/nodejs-starter) where everything is
-controlled by a few lines of config within your `package.json` file. Once you
-have run `npm install`, any time you make a commit, [XO] runs. It will
-then format and lint your code automatically.
-
-There is also a GDS boilerplate repository for Node.js apps. See below.
-
-Most editors can check XO syntax through plugins that in some cases comes pre-installed.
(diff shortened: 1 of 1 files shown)


nickcolley:
As a side note, we may want to consider consolidating the guidance in the styleguide repository into gds-way.

tombye:
Nice, well-explained bit of work this. Thanks for doing it @nickcolley.

I'm unclear about is the difference between the [nodejs-starter](https://github.com/alphagov/nodejs-starter) and the [boilerplate](https://github.com/alphagov/gds-nodejs-boilerplate). Is the former redundant?

Nooshu:
> As a side note, we may want to consider consolidating the guidance in the styleguide repository into gds-way.

I also want to consolidate the excellent work the Design System team has done around "How we write JavaScript" into the GDS Way in the near future too.

nickcolley:
Going to leave this open until the end of the week for more people to review.

jonheslop:
Looks good to me, it’s good to be consistant.

@tombye re the difference between [nodejs-starter](https://github.com/alphagov/nodejs-starter) and [boilerplate](https://github.com/alphagov/gds-nodejs-boilerplate).

I made the [starter](https://github.com/alphagov/nodejs-starter) first for myself when before I started working on [gaap-analytics](https://github.com/alphagov/gaap-analytics), it’s a JS project but it’s not an app.

The [boilerplate](https://github.com/alphagov/gds-nodejs-boilerplate) is basically the most stripped back version of all the pay apps, replicating the packages used and the structure (except I used XO not StandardJS, but we can update it)

nickcolley:
Thanks everyone for the feedback. 👍

PR #191, opened 2018-10-03
Extend expiry date - monitor your service

soupdragon99:
I've extended the date for six months. This just needs a review and an approval that it's good to go.

### DIFF
--- a/source/standards/monitoring.html.md.erb
+++ b/source/standards/monitoring.html.md.erb
@@ -6,21 +6,19 @@ review_in: 6 months
 
 # <%= current_page.data.title %>
 
+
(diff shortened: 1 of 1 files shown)


philandstuff:
Hmm, reviewing this, I'm not sure what this bit means:

> apply metrics-based monitoring because it’s not platform-specific

What is this trying to say? I really can't tell.  You should apply metrics-based monitoring because it gives you useful information about your service, not merely because it's not platform-specific.

Can we maybe remove this line, or replace it with something like "you should apply metrics-based monitoring to measure aggregated numerical data about your service"?

philandstuff:
cc @emilygds

soupdragon99:
@philandstuff thanks for the review, I have added in your suggested change. Are you happy to extend the date on this on the basis that the guidance is good enough for the time being?

PR #194, opened 2018-10-05
Enable tech docs multipage functionality - DO NOT MERGE

soupdragon99:
This PR contains the pattern for the new tech docs multipage config pulled from the tech doc source repo. Currently published GDS Way documentation has been reviewed and amended to fit this new pattern.

This is for contribution and comment. I have taken the view of transferring the content as it currently stands in terms of heading and layout as far as possible. In this PR I have suggested three headings for content, which are: 

- Build Software
- Operate Services
- Secure your service

These categories are open for comment. I think it would be good to add in placeholder titles for the backlog in the GDS Way trello board as I think it will encourage more contributions. But, I'd like input from people as to how we can structure this as there's no doubt several ways we can potentially structure the content beyond the above three categories.

See the [PaaS documentation ](https://docs.cloud.service.gov.uk/#gov-uk-platform-as-a-service)as an example of this new pattern.


timblair:
> As a general thing, this change has lost the distinction between "standards" and "manuals." Is that on purpose?
>
> * yes, to move towards a more 'service based' feel to the site. I've had anecdotal feedback that because we have things on similar topics in manuals and guidance and it was confusing.

That's fine. Just wanted to make sure it was an explicit choice.

> We've lost the statement on each page that "The GDS Way and its content is intended for internal use by the GDS community."
> 
> * do you mean the banner on each page of guidance? the new format renders multiple pages into a single section, I'm not sure how we can change this?

I mean this bit of text at the top of each page, which is there to say "this is GDS internal guidance, not something that we're telling the rest of government that they should follow":

screen shot 2018-11-09 at 16 52 56

> The "Add new guidance" needs updating due to single/multi-page thing.
>
> * I've removed the partial ref from 'add new guidance' was there anything else you had in mind?

It's more the bit that says "When you create a new Markdown file follow this pattern and make a pull request." That's no longer true, given the addition of the partials to pull the individual content pieces in.

My suggestion is to remove that page entirely, and put the template + any additional information in the README (which is where someone will be if they're making changes). We already have various links to the GitHub repo, and it makes sense to me to have all the "how to add / edit a thing" content in one place.

soupdragon99:
@timblair thanks for the update Ti. I've made a couple of changes based on your feedback. Combining the 'about' section with the landing page content and also moving the 'add new guidance' section to the readme.

willp-bl:
I have squashed all the commits and opened a new PR, so I am closing this one.

New PR: https://github.com/alphagov/gds-way/pull/215

PR #195, opened 2018-10-10
Link to Python dependencies advice and general review

galund:
Advice about Python dependencies was not linked to from the one place in our guidance that you would most likely look for it.

We had various slightly duplicative sentences and links into the Service Manual, which this commit attempts to rationalise.

I've grouped together the notes about vulnerability monitoring.

If there's anything else anyone wants to tweak in this guidance, then now is a good time, as it was due for review tomorrow (quite by chance).

https://trello.com/c/ncD4BuBf/56-python-dependencies

### DIFF
--- a/source/standards/tracking-dependencies.html.md.erb
+++ b/source/standards/tracking-dependencies.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: How to manage third party software dependencies
-last_reviewed_on: 2018-04-11
+last_reviewed_on: 2018-10-10
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


lfdebrux:
Is it worth adding a line saying that Snyk can also monitor Javascript and Python dependencies?

galund:
Can do. I'm a bit uncertain of the value of the Snyk case study - it's a bit unclear what the result of the evaluation actually is, and contains a few weird bits (like I don't understand where it says "The team decided it was best not to allow a third party product to add a file to their code repositories." - because the preceding sentence doesn't seem to be suggesting Synk would do that!).

source/standards/tracking-dependencies.html.md.erb
```diff
@@ -10,43 +10,29 @@ When you develop and operate a service, it’s important to keep any third party
...
 * address vulnerabilities in private before making the fix public
 
-## Dependency management tools
+### Snyk
 
-A common way to manage third party dependencies is to use a dependency management tool. [The Service Manual][] has some suggestions about [how to work with third party code][] and about some common dependency management tools.
+[Snyk][] is a dependency management tool which can [monitor code dependencies][] for security vulnerabilities. Using Snyk allows you to automatically check GitHub pull requests (PRs) for vulnerable dependencies and potentially create a pull request to fix them. It also supports the other languages that GDS uses, including Python, Java and JavaScript.
 
-## Case study - Snyk
+Snyk is being used successfully in the Digital Marketplace and GOV.UK Pay programmes, but configured to ignore vulnerabilities that have no fix, and without automatic generation of pull requests.
```
katstevens: Should we state that we're using both the pre-merge check and the weekly reports? If so I'd reword this to something like:

```
Snyk is being used successfully in the Digital Marketplace and GOV.UK Pay programmes. The Digital Marketplace uses Snyk as a pre-merge check on all pull requests for its public repositories. The developers also review the weekly report for new vulnerabilities, applying fixes manually instead of using Snyk's automatic pull request feature. The reports can be configured to ignore vulnerabilities that have no fix. 
```

PR #210, opened 2018-11-14
Python manual/advice update (dependencies)

galund:
@lfdebrux and I have attempted to update our advice re Python dependencies as follows:

* rearrange the sections
* lose some detail where it's covered better elsewhere (e.g. Snyk)
* expand on our treatment of dependencies, to cover applications and libraries in quite different ways.

There's nothing really _new_ here, which is kinda of the point of the GDS Way I guess.

https://trello.com/c/ncD4BuBf/56-python-dependencies

### DIFF
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -16,14 +16,15 @@ projects at GDS.
 
 We follow [PEP8][], in some cases [PEP8][] doesn't express a view (e.g. on the usage of language
 features such as metaclasses) in these cases we defer to the [Google Python style guide][GPSG].
-We suggest using [PEP8][] and the [Google Python style guide][GPSG] (in order of preference) as references unless
-something is explicitly mentioned in the [GDS Python Style Guide][GDSPSG].
+We suggest using [PEP8][] and the [Google Python style guide][GPSG] (in order of preference) as references
+unless something is explicitly mentioned in the [GDS Python Style Guide][GDSPSG].
 
-The [GDS Python Style Guide][GDSPSG] is the place for detailing GDS specific rules/ exemptions to [PEP8][] or the
-[Google Python style guide][GPSG]. An example of which is allowing a line length of 120 instead of 79.
+The [GDS Python Style Guide][GDSPSG] is the place for detailing GDS specific rules/exemptions to [PEP8][]
+or the [Google Python style guide][GPSG]. An example of which is allowing a line length of 120 instead
+of 79.
 
-As always the rules in the [GDS Python Style Guide][GDSPSG] should be followed only in conjunction with the advice
-on consistency on the main [programming languages manual page][GDSPSG].
+As always the rules in the [GDS Python Style Guide][GDSPSG] should be followed only in conjunction
+with the advice on consistency on the main [programming languages manual page][GDSPSG].
 
 If you want to add a new rule or exception please create a pull request against this repo.
 
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -1,26 +1,23 @@
 ---
 title: Writing Python at GDS
-last_reviewed_on: 2018-07-31
-review_in: 6 months
+last_reviewed_on: 2018-11-09
+review_in: 3 months
 ---
 
 # <%= current_page.data.title %>
 
-### About this manual
+## About this manual
 
 This manual is designed to aid developers in writing Python code that is clear and consistent, within, and across,
 projects at GDS.
 
 
-### Guidance
+## Coding standards
 
 We follow [PEP8][], in some cases [PEP8][] doesn't express a view (e.g. on the usage of language
 features such as metaclasses) in these cases we defer to the [Google Python style guide][GPSG].
 We suggest using [PEP8][] and the [Google Python style guide][GPSG] (in order of preference) as references unless
 something is explicitly mentioned in the [GDS Python Style Guide][GDSPSG].
-
-
-### Additional rules
 
 The [GDS Python Style Guide][GDSPSG] is the place for detailing GDS specific rules/ exemptions to [PEP8][] or the
 [Google Python style guide][GPSG]. An example of which is allowing a line length of 120 instead of 79.
@@ -30,8 +27,7 @@ on consistency on the main [programming languages manual page][GDSPSG].
 
 If you want to add a new rule or exception please create a pull request against this repo.
 
-
-### Linting
+## Linting
 
 _For more on linting (including standard settings and config) see the [linting][linting] section of this manual._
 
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -1,106 +1,150 @@
...
 
-It is important to have a good strategy for specifying your project's Python
-libraries. There are two desirable characteristics:
+There are two ways of specifying dependencies in Python world: as a specific version or
+as an allowable range.
 
-1. Reproducibility
+Different considerations apply to dependency management depending whether you are packaging a library, or creating
+a deployable artifact such as an application or a bundle of scripts: in general, you should only specify specific
+versions at the last responsible moment.
```
theseanything: What is meant by `the last responsible moment`?

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -1,106 +1,150 @@
...
+Your pinned dependencies should be fully specified in a file called `requirements.txt`, and checked
+into your version control system (VCS). For projects with only a small number of dependencies, maintaining
+this manually (for example, installing with `pip install`, then using `pip freeze`) may be adequate.
+
+For larger projects, GDS recommends the following approach, which was developed after various issues
+were found in current mainstream tooling (see [this commit message][dm-deps-commit], and note also the
+[lack of support for specifying VCS dependencies](https://gds.slack.com/archives/CAD6NP598/p1541777461082500)
+in `pip-compile`).
+
+Put your top-level requirements in a `requirements-app.txt` file.
```
theseanything: What's meant by a `top-level requirement`? And whats the reasoning behind this?

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -1,106 +1,150 @@
...
+dependencies can be upgraded to get bug fixes and security fixes, without introducing
+breaking changes to your build.
+
+Your pinned dependencies should be fully specified in a file called `requirements.txt`, and checked
+into your version control system (VCS). For projects with only a small number of dependencies, maintaining
+this manually (for example, installing with `pip install`, then using `pip freeze`) may be adequate.
+
+For larger projects, GDS recommends the following approach, which was developed after various issues
+were found in current mainstream tooling (see [this commit message][dm-deps-commit], and note also the
+[lack of support for specifying VCS dependencies](https://gds.slack.com/archives/CAD6NP598/p1541777461082500)
```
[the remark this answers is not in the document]
galund: ha that was a copy-paste fail. Gawd knows what slack message that was :-o

source/manuals/programming-languages/python/python.html.md.erb
galund: Top-level, as in: the libraries that this application depends upon, rather than the dependencies of _those_ dependencies (i.e. not the sub-dependencies).

They need to be specified somewhere. One option is to have them in a `setup.py` file (which is what we recommend for libraries); however, wrangling them is much easier when they are in plain text files. In `pip-compile`, the file would be called `requirements.in` and maybe we should consider using that name instead, or making clear that we really don't care what the name is, just that you follow this model for having top-level requirements and then a resulting pinned full dependency list.

If there's a form of words that could make this clear (assuming it now is) then please do suggest, very happy to modify this further.

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -1,106 +1,150 @@
...
+dependencies can be upgraded to get bug fixes and security fixes, without introducing
+breaking changes to your build.
+
+Your pinned dependencies should be fully specified in a file called `requirements.txt`, and checked
+into your version control system (VCS). For projects with only a small number of dependencies, maintaining
+this manually (for example, installing with `pip install`, then using `pip freeze`) may be adequate.
+
+For larger projects, GDS recommends the following approach, which was developed after various issues
+were found in current mainstream tooling (see [this commit message][dm-deps-commit], and note also the
+[lack of support for specifying VCS dependencies](https://github.com/jazzband/pip-tools/issues/355)
```
MatMoore: Given that this is a moment time in evaluation of a set of tools, some of which are quite new, I would prefer for this section to explicitly mention specific tools that were evaluated (eg pipenv version such and such) and summarise the reasons against adoption at this time (eg lack of support for VCS dependencies), instead of linking out to a decision made for a particular project.

I think it's important that people who are thinking of deviating from the recommendation can make an informed decision, based on what we know now and any additional information available to them at the time.

Similarly, people reviewing this document in X months time should be able to quickly evaluate whether the conclusion still makes sense given how the tools have changed since it was written.

source/manuals/programming-languages/python/python.html.md.erb
galund: We mean that you should specify broad version ranges where possible (e.g. so that if you're building a library, people aren't stuck using a specific version just because it happens to claim compatibility with the _exact_ version of some other library you are already using).

But conversely when you pulling together libraries to build an application, then having a precise "known good" set of dependencies that should actually be deployed is very valuable for the reasons given elsewhere in this document.

LRM is a general principle that's googleable but I can see we haven't made its application here particularly obvious, again happy to accept alternative formulations (cc @lfdebrux).

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -1,106 +1,150 @@
...
+in `pip-compile`).
+
+Put your top-level requirements in a `requirements-app.txt` file.
+
+Use a "freeze" script (as seen in [this Makefile][dm-deps-commit-makefile]) to generate the
+fully specified `requirements.txt`, which is used whenever you need to pip-install the
+dependencies.
+
+* The script creates a temporary virtualenv, pip-installs the `requirements-app.txt` and
```
MatMoore: This feels like information that should belong in a readme.

Can we package up the makefile solution into a standalone repo so we don't have to copy and paste it across projects? Even if it was just a template, it seems like it would be simpler to link to that than use Digital Marketplace as the template and have to update this if they change anything.

source/manuals/programming-languages/python/python.html.md.erb
MatMoore: can we define it on first use, or link to an external explanation?

lfdebrux:
@MatMoore Hi Mat, thanks for your comments, I will respond to them individually but I wanted to make a general observation as well, which is that the work George and I were doing here was to bring the GDS way guidance more into line with current practice at GDS; while what is currently hosted is representative, it doesn't really go into details on some problem areas around Python dependency management (specifically the difference between apps and libraries).

The medium term aim is to (re)start discussion around how we do dependency management with Python in GDS, but we felt that to do that adequately we first needed to have the current situation properly written down.

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: I see your point, however I don't know how much detail is supposed to go into these manuals; the small amount I know about previous efforts in this area is that there have been a lot of experiments and a lot of dead ends, so documenting everything could take a lot of words!

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: > Can we package up the makefile solution into a standalone repo so we don't have to copy and paste it across projects? Even if it was just a template, it seems like it would be simpler to link to a standalone repo than use Digital Marketplace as the template and have to update this if they change anything.

I'm sure this is something we could do, although it would have to be with the understanding that someone would have to maintain it.

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: Using the phrase in this context was a brain-fart of mine; I don't think anyone else has used that form of words when referring to Python dependency management before.

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: I think this is related to the question around `last responsible moment`; clearly what is needed here is an exposition of the dependency hierarchy in the introduction that defines the terms `dependency`, `sub-dependency`, and maybe clearly explains why the distinction is important when dealing with Python packaging.

lfdebrux:
Two questions for GDS Way maintainers; what is the expected level of knowledge of a reader? and how much detail is expected to go into an individual page?

A lot of times George and I have debated whether to put something in or not and have been worried about increasing the amount of verbage to explain something that is sort of incidental but also useful to understand the context of why we jump through these hoops.

source/manuals/programming-languages/python/python.html.md.erb
benvand: I think the important thing we're trying to communicate here are the steps one should take to `freeze` their requirements. I don't think it would be too much to put in the actual steps a dev should take, something like:

>>>
To get a requirements.txt with the pinned dependencies and sub-dependencies for your project you should:
1. Put your known, first level dependencies in a requirements-app.txt file
2. Create and activate a new, clean, virtualenv
3. `pip install -r` your `requirements-app.txt` file
4. Run `pip freeze`

At this step `pip freeze` will output everything currently installed in your new virtualenv. The dependencies you specified in your `requirements-app.txt` and any sub-dependencies (dependencies of dependencies). This is the pinned list of requirements that should go in your `requirements.txt`

5. Clean everything up by deactivating and `rm -rf`ing the new virtualenv.

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: A suggestion from @benvand, maybe the best way to deal with this issue is to put some form of words at the top of the page saying `If you have an issue with or want to deviate from the recommendations in this manual, please contact #python-community on Slack in the first instance.`

source/manuals/programming-languages/python/python.html.md.erb
benvand: That way we don't need to include the part below about the specific drawbacks of our makefile (also the makefile version doesn't allow range-pinned dependencies whereas the above process does)

source/manuals/programming-languages/python/python.html.md.erb
galund: So that last point is one reason I added that link to a specific github issue. But in truth we haven't fully re-reviewed the cluster**** that is the official Python packaging toolset: it's not a moment in time evaluation, it's a collection of a couple of evaluations over a period of a year.

This time, I've done some evaluation of `pip-compile`, and that issue/feature request seems me to be the biggest blocker. We should take another look at `pipenv`, but I'd wager @allait's previous comments do still apply (see the referenced PR).

I personally suspect that `pipenv` simply tries to do too much (especially given that containerisation options move along in their own way), and that what we really want is a working `pip-compile`.

@benvand has already started a [reimplementation in Python](https://github.com/alphagov/digitalmarketplace-utils/pull/473/files) of our current approach for use in Digital Marketplace. Once that's actually been rolled out anywhere then we could certainly reference it from here, and I have a few ideas about how to improve it (slightly).

All this feels interim, and I'd definitely encourage anyone that has the energy to consider either fully re-evaluating `pipenv` on a complex project like Digital Marketplace or perhaps Notify, or look at hacking on that feature in `pip-compile`.

I'm not sure there's a way we can succinctly document current practice in such a way as to much assist with that 'quickly evaluate' thing, but if you can reword at all in that direction then go for it 👍

source/manuals/programming-languages/python/python.html.md.erb
galund: Yeah I think this is a good idea, we'd build on @benvand's thing. I'd probably like it if he go it into production for at least one DM system first, though :-)

The reason the explanation is here is because we aren't really mandating _that_ Makefile, just the effect it produces, so you need (here) to know what it does if you are to reproduce that (via copy-paste or by building your own or using `pip-compile` or whatever.

(I've thought before about maintaining a simple template project, but I tend to the view that if you're starting something new, then you might prefer to read the GDS Way and then see some pointers to the existing Python projects that do this consistently/well. But sticking the script in a lib you can easily install would certainly be nice, for as long as we do something better than `pip-compile` does at least.)

galund: Sorry just to clarify, I was responding to the original post, not to @benvand...

I don't actually agree with ben on this so I think maybe we should take this to #Python :)

source/manuals/programming-languages/python/python.html.md.erb
theseanything: Maybe split this into two? The first part explains the *specificity* of dependencies for libraries vs repeatable installation of artifacts. Then a section dicussing *when* you actually decide on your dependency versions. (A definition would be good - I think linking out would be disruptive.)

benvand:
@MatMoore 
Had a realisation:
I think we need to tell them to use an already existing script.
The (various) scripts in GDS Python projects that exist already preserve the github links in the requirements-app IIRC.
Having every team that starts a python project coming up against that problem is, not desirable...

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: I've added some further text and a diagram, please review if you have a moment

source/manuals/programming-languages/python/python.html.md.erb
lfdebrux: I've changed this to use the term 'immediate dependencies' which is defined in the new introductory text

PR #212, opened 2018-11-16
Update Java style guide

alexbishop1:
The Java community meeting reviewed the [Java style guide](https://gds-way.cloudapps.digital/manuals/programming-languages/java.html) on 9 November 2018.

There was rough consensus on some changes, as detailed in the meeting notes at https://docs.google.com/document/d/1K1LHRfG3OLD771A81pe7nJAebdd62eCRA8u2dbgC-pI/edit (this document should be visible to anybody logged in to a GDS Google account).

Details of the individual changes can be found in the commit messages.

### DIFF
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -16,9 +16,7 @@ We favour consistency across our code, so when considering using a new or differ
 
 Variable and field names should match the class they are instantiating, or have a descriptive name in the context of their use.
 
-Use the default code indenting in IntelliJ.
-
-Ensure line feeds are present at the ends of files.  This can be enabled as an option in IntelliJ: `Settings->Editor->Other->Ensure line feed at file end on save.`
+Use the [GDS Way EditorConfig file](editorconfig), which has settings for things like code indentation. Place a copy of this file named `.editorconfig` in the root of your project to have IntelliJ IDEA and some other editors automatically apply the settings. If your editor does not support EditorConfig, manually configure its settings to match.
 
 ## Dependency injection (DI)
 
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -34,11 +34,9 @@ The use of static analysis is encouraged. Static analysis tools for Java include
 
 Try to minimise compiler warnings. If you cannot remove a warning, use an appropriate annotation to suppress it, preferably with a comment explaining why.
 
-## Build tools / external dependencies
+## Build tools
 
 Either Gradle or Maven should be used as the build tool.
-
-Different programmes have differing requirements about where dependencies should be sourced from.
 
 ## Web frameworks
 
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -20,7 +20,7 @@ Use the [GDS Way EditorConfig file](editorconfig), which has settings for things
 
 ## Dependency injection (DI)
 
-Some programmes use [Guice](https://github.com/google/guice) as a dependency injection framework, however, consider whether use of DI is appropriate for your project.
+Consider whether dependency injection is appropriate for your project before using it. Some programmes use the [Guice](https://github.com/google/guice) dependency injection framework.
 
 ## Imports
 
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Java style guide
-last_reviewed_on: 2017-07-14
+last_reviewed_on: 2018-11-09
 review_in: 6 months
 ---
 
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -32,7 +32,19 @@ The IntelliJ "Optimize imports" command helps to sorting them alphabetically and
 
 The use of static analysis is encouraged. Static analysis tools for Java include [SonarQube](https://www.sonarqube.org/), [Codacy](https://www.codacy.com/), [FindBugs](http://findbugs.sourceforge.net/) and [CheckStyle](http://checkstyle.sourceforge.net/). However, be aware that such tools can detect an overwhelming number of problems if applied to an existing project, which tends to result in their checks being ignored.
 
-Try to minimise compiler warnings. If you cannot remove a warning, use an appropriate annotation to suppress it, preferably with a comment explaining why.
+Try to minimise compiler warnings. If you cannot remove a warning, use an appropriate annotation to suppress it, preferably with a comment explaining why. For example:
+
+```java
+public void frobulateFoos() {
+    LOGGER.info("Frobulating foos");
+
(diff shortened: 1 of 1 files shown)


heathd:
🎉  nice improvements and clarifications. Thank you 🙏

source/manuals/programming-languages/java.html.md.erb
```diff
@@ -16,13 +16,11 @@ We favour consistency across our code, so when considering using a new or differ
...
 
-Use the default code indenting in IntelliJ.
-
-Ensure line feeds are present at the ends of files.  This can be enabled as an option in IntelliJ: `Settings->Editor->Other->Ensure line feed at file end on save.`
+Use the [GDS Way EditorConfig file](editorconfig), which has settings for things like code indentation. Place a copy of this file named `.editorconfig` in the root of your project to have IntelliJ IDEA and some other editors automatically apply the settings. If your editor does not support EditorConfig, manually configure its settings to match.
 
 ## Dependency injection (DI)
 
-Some programmes use [Guice](https://github.com/google/guice) as a dependency injection framework, however, consider whether use of DI is appropriate for your project.
+Consider whether dependency injection is appropriate for your project before using it. Some programmes use the [Guice](https://github.com/google/guice) dependency injection framework.
```
MatMoore: I think this kind of language is too vague for a styleguide: "consider if X is appropriate"

Is there anything in particular to consider, that you wouldn't have already thought about before reading the styleguide? If not I think we can drop this sentence.

source/manuals/programming-languages/java.html.md.erb
```diff
@@ -32,14 +30,16 @@ The IntelliJ "Optimize imports" command helps to sorting them alphabetically and
 
 ## Code checking
 
-Use of [Sonar](https://www.sonarqube.org/) for static analysis of code is encouraged.
+The use of static analysis is encouraged. Static analysis tools for Java include [SonarQube](https://www.sonarqube.org/), [Codacy](https://www.codacy.com/), [FindBugs](http://findbugs.sourceforge.net/) and [CheckStyle](http://checkstyle.sourceforge.net/). However, be aware that such tools can detect an overwhelming number of problems if applied to an existing project, which tends to result in their checks being ignored.
```
MatMoore: The last sentence here puts me off following this advice. It's almost suggesting I shouldn't bother if I'm not setting up a brand new project.

Do we have any templates or examples of builds that have these set up? Is it possible to set up CI so that static analysis runs on the changes but not the whole project?

source/manuals/programming-languages/java.html.md.erb
```diff
@@ -32,14 +30,16 @@ The IntelliJ "Optimize imports" command helps to sorting them alphabetically and
...
 
-## Build tools / external dependencies
+Try to minimise compiler warnings. If you cannot remove a warning, use an appropriate annotation to suppress it, preferably with a comment explaining why.
```
MatMoore: Would be good to have a code snippet showing a good example of this

source/manuals/programming-languages/java.html.md.erb
alexbishop1: I think the desire here was to stop people from blindly using DI without thinking about it

source/manuals/programming-languages/java.html.md.erb
alexbishop1: I’m not aware of any templates or examples.

We could lose the subordinate clause from the last sentence, or perhaps add some advice of how to deal with the deluge (e.g. configure the tools to find fewer problems or have a prioritisation plan) and what to focus on (e.g. newer code — problems in code that’s been out there for ages can’t be that bad or someone would have complained already! — and more serious problems).

source/manuals/programming-languages/java.html.md.erb
alexbishop1: Something like this?

```java
public void frobulate() {
    LOGGER.info("Frobulating foos");

    // LibFoo returns raw list but every element is always a Foo
    @SuppressWarnings("unchecked")
    List<Foo> foos = (List<Foo>) fooService.getFoos();

    foos.forEach(foo -> foo.frobulate());
}
```

MatMoore: Looks good to me although shouldn't that comment go in the argument there instead of "unchecked"?

source/manuals/programming-languages/java.html.md.erb
MatMoore: Hmm, I'm just skeptical it would actually deter anyone unless we give examples of what is and isn't appropriate.

source/manuals/programming-languages/java.html.md.erb
MatMoore: I think I'd find the latter quite helpful, as someone who's very unfamiliar with the tooling.

source/manuals/programming-languages/java.html.md.erb
alexbishop1: I think I’d have to defer to the people who were most vocal about wanting to put more emphasis on considering the usefulness of DI for examples

source/manuals/programming-languages/java.html.md.erb
alexbishop1: I have some thoughts on guidance for this but it’s not something we discussed in the meeting and I don’t feel qualified to write something unilaterally

source/manuals/programming-languages/java.html.md.erb
alexbishop1: I’m not sure I follow. Could you give a code snippet to illustrate what you mean?

source/manuals/programming-languages/java.html.md.erb
MatMoore: No problem, I wasn't at the meeting so am missing some context here

source/manuals/programming-languages/java.html.md.erb
MatMoore: Ok, maybe it's best to address this as a separate PR then? Happy to discuss further if I can help at all, but I'm speaking more as a user of the guidance than someone with an opinion on how to do this.

source/manuals/programming-languages/java.html.md.erb
MatMoore: as in:

```java
    @SuppressWarnings("LibFoo returns raw list but every element is always a Foo")
```

source/manuals/programming-languages/java.html.md.erb
alexbishop1: Yeah, perfect is the enemy of good and all that

source/manuals/programming-languages/java.html.md.erb
alexbishop1: Ah, you can’t do that. The compiler actually understands `"unchecked"` to mean something (for `@SuppressWarnings`, the Java Language Specification defines `"unchecked"` and `"deprecated"` but other tools can support other strings). There’s no way to put an explanatory comment inside the annotation; it has to go in a comment outside.

MatMoore: ahhh ok. I think I misunderstood this as our codebase does use other strings in places. Can we include the snippet you posted above then?

alexbishop1: Pushed another commit with the `@SuppressWarnings` example

Issue #186, opened 2018-09-24
Basic container guidance


willp-bl:
It would be good to re-use all/some of the guidance Laura wrote in this PR - https://github.com/alphagov/gds-way/pull/89

If it can capture the way things are working right now it can be amended/expanded later as needed.

soupdragon99:
There's a card in the GDS Way backlog for something on this. I do have the original guidance Laura put together, I'll raise a PR to open a discussion and people can update it as they see fit.

soupdragon99:
raised [PR23](https://github.com/alphagov/gds-way/pull/213) to open discussion.

PR #213, opened 2018-11-21
Add basic container guidance

soupdragon99:
Basic guidance about how to use containers based on the content Laura originally wrote in [PR - #89](https://github.com/alphagov/gds-way/pull/89).

This PR raised in response to [Basic container guidance #186](https://github.com/alphagov/gds-way/issues/186) for discussion and as a start point.


jonodrew:
I'm uncomfortable with the idea that the use of containers _specifically_ be a standard. I know certain teams are using serverless, and I don't think we should dictating which technologies should be used.

deanwilson:
This document will be re-written in light of the currently used container technologies and our new strategy.

PR #214, opened 2018-11-22
Add basic guidance about data stores

soupdragon99:
This PR is to add some initial basic guidance about data stores, prompted by a card in the [GDS Way backlog](https://trello.com/c/a5iVWpAg/9-databases).


timblair:
This feels a lot like broad guidance, rather than specific standards. For example, there are a few instances of "consider X if Y," rather than "use X if Y."

PR #215, opened 2018-11-22
Enable multipage

willp-bl:
This commit contains the pattern for the new tech docs multipage config pulled from the tech doc source repo. Currently published GDS Way documentation has been reviewed and amended to fit this new pattern.

This is for contribution and comment. I have taken the view of transferring the content as it currently stands in terms of heading and layout as far as possible. In this commit I have suggested three headings for content, which are:

* Build Software
* Operate Services
* Secure your service

These categories are open for comment. I think it would be good to add in placeholder titles for the backlog in the GDS Way trello board as I think it will encourage more contributions. But, I'd like input from people as to how we can structure this as there's no doubt several ways we can potentially structure the content beyond the above three categories.

See the [PaaS documentation](https://docs.cloud.service.gov.uk/#gov-uk-platform-as-a-service) as an example of this new pattern.


source/documentation/build-services/accounts-with-third-parties.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: How to manage access to your third-party service accounts
-last_reviewed_on: 2018-06-04
+title: Manage third party service accounts
+last_reviewed_on: 2018-22-11
```
willp-bl: this doesn't look quite right

willp-bl:
:shipit:

PR #227, opened 2018-12-06
How to review code - additional links

katstevens:
Pulls in some additional code review resources from the equivalent page on the tech learning pathway, minus one broken link (https://github.com/alphagov/gds-tech-learning-pathway/blob/master/source/resources/other/code-reviews.html.md).

I think the other code review advice from that page (as opposed to advice on writing PRs) is covered already - happy to add in anything missing?

### DIFF
--- a/source/manuals/code-review-guidelines.html.md.erb
+++ b/source/manuals/code-review-guidelines.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: How to review code
-last_reviewed_on: 2018-06-04
+last_reviewed_on: 2018-12-06
 review_in: 6 months
 ---
 
--- a/source/manuals/code-review-guidelines.html.md.erb
+++ b/source/manuals/code-review-guidelines.html.md.erb
@@ -98,7 +98,13 @@ You can find out more about how to review code by reading:
 * [the social aspects of code review][]
 * [phrasing review comments positively][]
 * [collated feedback and data from a developer survey][]
-
+* [the art of giving and receiving code reviews][]
+* [how to conduct effective code reviews][]
+* [The Art of Code Review][] (video, 23 minutes)
+* [Mozilla Science Lab: reviewing a contribution][]
+* [thoughtbot code review guide][]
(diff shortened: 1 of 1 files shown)


MatMoore:
@katstevens I think I might have given the wrong impression when we talked about this the other day - the learning pathways site is still intended to be current so I don't think it makes sense to maintain a list of blog posts and other links in both projects.

Both pages have a lot of overlap but (IMO) they each have a different purpose:

- The GDS way explains how we do things at GDS and is relevant to new staff, when forming new teams, or when changing the way an existing team does something
- The learning pathways is aimed at helping technologists at GDS develop and get promoted

So my suggestion for untangling this would be:
- this page should focus on processes, expectations and rules for code review, but not cover general best practice or link to non-GDS websites
- the learning pathways page should collect all these links that explain how to give a good code review and how to engage with code reviews, but it shouldn't prescribe any particular practice

Do you agree with this general approach? It would involve removing some content like the beginning section about good review practice. I'm happy to pair on this if you want, as I've already got a PR in progress on the other repo: https://github.com/alphagov/gds-tech-learning-pathway/pull/125

Also, I don't think this page actually explains the actual rule we follow here for code review:
- Every pull request must be approved by another member of staff
- If 2 or more people paired on a PR, they can approve it themselves

We might also want to consider in this guidance how we respond to external pull requests, as this requires more care than internal pull requests. As I understand it we don't have a well defined process for engaging with open source communities but past teams have done this well, for example the GDS operations team who documented their process here: https://gds-operations.github.io/contributing/

katstevens:
That all sounds good to me @MatMoore - I think pairing is a good idea to make sure everything's covered in one place or the other. Let's catch up next week.

> If 2 or more people paired on a PR, they can approve it themselves

Is this actually the case? I think it might vary between teams (also some teams require two approvals for certain categories of PR, etc). Perhaps some more generic guidance along the lines of 'agree this  process with your team', giving those rules as an example? Then we can mention that any agreed policy should cover external PRs too.

PR #257, opened 2019-04-03
Update tech debt tracking guidance

timblair:
Add information about changing impact scores over time.

### DIFF
--- a/source/standards/technical-debt.html.md.erb
+++ b/source/standards/technical-debt.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: How to track technical debt
-last_reviewed_on: 2018-10-03
+last_reviewed_on: 2019-04-03
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/standards/technical-debt.html.md.erb
```diff
@@ -63,6 +63,13 @@ rating based on a decision about the relative importance of the impact and
 likelihood. The higher the overall rating, the more the programme should be
 concerned about it.
 
+The impact of a piece of technical debt, and the effort required to remove the
+cause of it, will change over time. If a problematic system was not being
+actively worked on at the time a piece of debt was originally recorded, it may
+have a low impact rating. If it subsequently becomes worked on again, the
+impact should be updated. Recorded items of technical debt should be reviewed
```
SteveEvansGDS: Can we re-write these 2 sentences to make them active rather than passive? Something like this works:

"If it subsequently becomes worked on again, you should update the impact. You should review recorded items of technical debt at appropriate periods."

source/standards/technical-debt.html.md.erb
```diff
@@ -63,6 +63,13 @@ rating based on a decision about the relative importance of the impact and
...
+at appropriate periods.
+
```
SteveEvansGDS: Can we re-write the last 2 sentences to make them active rather than passive? Something like this works:

"If it subsequently becomes worked on again, you should update the impact. You should review recorded items of technical debt at appropriate periods."

PR #274, opened 2019-05-09
Update Git guidance

ChrisBAshton:
Encapsulates:

- Updated `git push -f` guidance (https://github.com/alphagov/styleguides/pull/121)
- Explanation of importance of good commit messages (https://github.com/alphagov/styleguides/pull/117)
- Explains why we line-break content longer than about 80 characters.

This last point is something I've picked up on through osmosis but if anyone has any more tangible reasons, or software that automates the line breaking for you, I'm happy to amend the line-break instructions 🙌

### DIFF
--- a/source/standards/git.html.md.erb
+++ b/source/standards/git.html.md.erb
@@ -18,6 +18,10 @@ developers on your project. This includes:
 * any future developers (including yourself) who want to see why a change was
   made
 
+Capturing context around a change allows people to understand why a particular
+implementation decision was made, much like an [architecture decision record](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions). We're
+being kinder to our future selves.
+
 ### Recommended blog posts on this topic
 
 * [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/)
@@ -57,7 +61,6 @@ history to understand why a change was made.
 If you are adding a link to a publicly viewable repository, ensure that the linked
 ticket is publicly viewable (and likely to remain so).
 
-
 ### Structure
 
 Commit messages should start with a one-line summary no longer than 50
@@ -71,7 +74,6 @@ For example:
 You should leave a blank line before the rest of the commit message, which you
 should wrap at around 72 characters: this makes it easier to view commit
 messages in a terminal.
-
 
 #### Example
 
--- a/source/standards/git.html.md.erb
+++ b/source/standards/git.html.md.erb
@@ -133,35 +133,23 @@ branch in the repository history. This advice may be freely ignored for smaller
 local feature branches for which a fast-forward merge will look like any other
 routine development work on `master`.
 
-## Never call `git push -f` without additional arguments
+## Don't use `git push -f`, use `--force-with-lease` instead
 
 Force pushing in git is a subject that attracts all kinds of religious
-battles. This advice is not about the merits of force pushing.
-
-This is about how to use `git push -f` if and when you do use it.
+battles. This advice is not about the merits of force pushing. This is about how
+to use `git push -f` if and when you do use it.
 
 Let's say you're working on a branch, 'foobar', and you decide to force push
 to the remote. So you do this:
 
     $ git push -f
 
-If anyone has committed anything to master[^1] since you last pulled -- and if
-you've been working on the branch for any length of time this is pretty likely
--- you will blow their changes away, because without arguments git will force
-push *all* remote-tracking branches.
+If anyone else has committed anything to your branch since you last pulled, you
+will blow their changes away.
 
-So, if you ever need to force push the 'foobar' branch, please instead do
+So for a bit of safety, if you ever need to force push please instead do:
 
-    $ git push --force-with-lease origin foobar
+  $ git push --force-with-lease
 
 `--force-with-lease` refuses to update a branch unless it is the state that we expect; i.e. nobody has updated the remote branch.
 See: https://developer.atlassian.com/blog/2015/04/force-with-lease/
-
-You can also change git's default behaviour of pushing all tracked branches by
-doing `git config --global push.default upstream`, but you should probably get
-into the habit of typing out your intentions in full when doing a destructive
-operation like force pushing, otherwise disaster will strike when you use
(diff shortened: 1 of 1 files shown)


source/standards/git.html.md.erb
```diff
@@ -18,6 +18,16 @@ developers on your project. This includes:
 * any future developers (including yourself) who want to see why a change was
   made
 
+  Taking the time to write a good commit message is important for a few reasons.
```
timblair: The double-space at the beginning of this line should be removed.

ChrisBAshton: Well spotted!

source/standards/git.html.md.erb
```diff
@@ -131,35 +142,37 @@ branch in the repository history. This advice may be freely ignored for smaller
...
 
-You can also change git's default behaviour of pushing all tracked branches by
-doing `git config --global push.default upstream`, but you should probably get
-into the habit of typing out your intentions in full when doing a destructive
-operation like force pushing, otherwise disaster will strike when you use
-someone else's differently-configured git, or miss a step when configuring a
-new machine.
+## File content conventions
+
+Where possible, it is a good idea to keep line length to around 80 characters.
```
galund: This varies by language community, I think, and is inconsistent with guidance elsewhere, for example https://gds-way.cloudapps.digital/manuals/programming-languages/python/style-guide.html

source/standards/git.html.md.erb
```diff
@@ -131,35 +142,37 @@ branch in the repository history. This advice may be freely ignored for smaller
...
+This ensures your colleagues can easily scan the contents of your file without
+having to set any word wrapping in their editor.
+
+This applies not just to code, but to documentation too. If, for example, you
```
galund: I used to think this, but actually having to re-flow paragraphs to 80 characters make it much _harder_ to follow the diff. Where a single word has been added, suddenly you have entirely lines changes.

So I actually think we should remove this guidance altogether.

ChrisBAshton: I totally agree to be honest! Just putting into guidance what I'm seeing, as it does seem to be a convention GOV.UK developers are following, but I think it's prone to inconsistency and solves a problem that doesn't really exist.

Happy to remove the guidance, though perhaps it should be put to some sort of vote/RFC 🤔

source/standards/git.html.md.erb
```diff
@@ -18,6 +18,16 @@ developers on your project. This includes:
 * any future developers (including yourself) who want to see why a change was
   made
 
+Taking the time to write a good commit message is important for a few reasons.
+
+It's being kinder to our future selves. Like an
+[architecture decision record][adr], capturing context around a change allows
+people to understand why a particular implementation decision was made.
+ 
+It's an advertisement to work with you. Its demonstrates that you are
```
philandstuff:
```suggestion
It's an advertisement to work with you. It demonstrates that you are
```

source/standards/git.html.md.erb
```diff
@@ -18,6 +18,16 @@ developers on your project. This includes:
...
+experienced enough to care about future maintainers. It's a palpable form of
```
philandstuff: "experienced enough" doesn't sound right - you can have lots of experience and not care, or you can be a new junior and get it straight away.  Maybe "It demonstrates that you care about future maintainers" is enough?

source/standards/git.html.md.erb
```diff
@@ -72,6 +81,8 @@ You should leave a blank line before the rest of the commit message, which you
 should wrap at around 72 characters: this makes it easier to view commit
 messages in a terminal.
 
+Ignoring the established community guidelines for structuring commit messages
+is mildy anti-social to other git users.
```
philandstuff: While I completely agree with this, I don't think it's necessary here. We have the motivation at the top of the document, the rest of the document should just be matter-of-fact statements about what is expected.

source/standards/git.html.md.erb
```diff
@@ -18,6 +18,16 @@ developers on your project. This includes:
...
 
+Taking the time to write a good commit message is important for a few reasons.
+
+It's being kinder to our future selves. Like an
+[architecture decision record][adr], capturing context around a change allows
+people to understand why a particular implementation decision was made.
+ 
+It's an advertisement to work with you. Its demonstrates that you are
+experienced enough to care about future maintainers. It's a palpable form of
+levelling up.
```
philandstuff: This last sentence feels unnecessary, and uses overly complicated language. (Can we get a tech writer to review?)

source/standards/git.html.md.erb
philandstuff: While I prefer reflowing, I mainly feel that this is not a common enough problem to need "solving" in the GDS way, so I agree we just shouldn't have this section

source/standards/git.html.md.erb
ChrisBAshton: Agreed - will edit. Thanks :)

source/standards/git.html.md.erb
ChrisBAshton: Agreed 👍

source/standards/git.html.md.erb
ChrisBAshton: Agreed, I think we can remove this paragraph tbh.

source/standards/git.html.md.erb
ChrisBAshton: Removed 👍

source/standards/git.html.md.erb
ChrisBAshton: Section removed 👍

source/standards/git.html.md.erb
```diff
@@ -18,6 +18,12 @@ developers on your project. This includes:
 * any future developers (including yourself) who want to see why a change was
   made
 
+Taking the time to write a good commit message is important for a few reasons.
```
galund: This still kinda duplicates the first sentence of the section, and the following para could arguably read better as an amendment to the last bullet point in the list of reasons?

ChrisBAshton: Agreed!

PR #282, opened 2019-05-20
Update CSS coding style

SteveEvansGDS:
Bumped the date and made some style changes. It'd be good if someone could check it from a technical point of view.

### DIFF
--- a/source/manuals/programming-languages/css.html.md.erb
+++ b/source/manuals/programming-languages/css.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: CSS coding style
-last_reviewed_on: 2016-11-18
+last_reviewed_on: 2019-05-20
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/css.html.md.erb
```diff
@@ -116,12 +114,12 @@ future when the vendor prefixed versions aren't needed.
...
   ```
 
-  **Why:** by putting parent reference selectors first it keeps the styling for
+  Putting parent reference selectors first keeps the styling for
   that element together. By putting styling for other elements in the middle you
   have to scroll around watching nesting levels to try and figure out what the
   `&` references.
 
-* Where possible rather than repeating selectors nest blocks together
+You should nest blocks together rather than repeating selectors where possible.
```
paroxp: Should there be a limit of nesting?

Imagine:
```
body {
  main {
    article {
      aside {
        div.links {
          p {
            a {
              &:hovered {
```

all of them will have things in it. But it suddenly get's unreadable?

Maybe just my opinion :)

fofr: Sass linting in govuk-lint used to warn if nested too deep:
https://github.com/alphagov/govuk-lint/blob/master/configs/scss_lint/gds-sass-styleguide.yml#L97-L100

I vaguely recall nesting being discouraged too. cc @nickcolley

nickcolley: We try to avoid nesting at all times since it makes it very difficult to search the codebase, and hides complexity.

nickcolley: It seems the whole guide is really out of date, so should probably have a review from the frontend community at some point. :eyes:

PR #287, opened 2019-05-24
Updated HTML styleguide

ChrisBAshton:
## Summary

- Reorder the guidance into sections: miscellaneous, document structure, element structure, individual elements
- Adds section on progressive enhancement and pre-HTML5 support.
- Removes dedicated ARIA section which contained a lot of repetition.
- Adds section on images and alt text.
- Adds section on buttons vs links.
- Adds section on visually hidden elements.
- Adds section on headings.

### DIFF
--- a/source/manuals/programming-languages/html.html.md.erb
+++ b/source/manuals/programming-languages/html.html.md.erb
@@ -6,80 +6,157 @@ review_in: 6 months
 
 # <%= current_page.data.title %>
 
-## HTML5
+## Browser support
 
-* Always quote attribute values - the decision tree for remember when to
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+The GOV.UK Service Manual lists the [minimum browsers that GOV.UK services are required
+to support](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in).
+Tools such as GOV.UK Frontend, that are used to help build services, may have
+[stricter standards of browser support](https://github.com/alphagov/govuk-frontend#browser-support).
+Therefore, you will need to test your code in different browsers depending on
+which part of GDS you are working in.
 
-## Use [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) roles appropriately
+Whatever your level of browser support, HTML should be written in such a way
+that older browsers see an appropriate fallback or warning message. For example,
+some browsers do not support the `<video>` element, so you should ensure that
+some contextual information is included (the `<p>` element in this example):
 
-### Guidance on using [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmarks
+```html
+<video width="320" height="240" controls>
+  <source src="movie.mp4" type="video/mp4">
+  <source src="movie.ogg" type="video/ogg">
+  <p>Unable to load video. Please upgrade your browser.</p>
+</video>
+```
 
-When you use HTML5 sectioning elements it's a good idea to include
-the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmark roles as well. The trick when you're
-using both HTML5 and [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is to put the landmark on the sectioning
-element itself because this avoids the information being duplicated by
-screen readers that support both technologies.
+When using HTML5 elements such as `<main>`, ensure that you include a
+[shim that allows them to be styled](https://github.com/aFarkas/html5shiv) in
+legacy browsers. Include the script inside a `<!--[if lt IE 9]><![endif]-->`
+conditional comment so that modern browsers do not download them.
+
+Your page should work with just HTML, before adding any CSS or JavaScript. Read
+[building a resilient frontend using progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement).
+
+## Document structure
+
+Use HTML5 to structure your page semantically. Use ARIA roles to help
+older assistive technologies map the sections correctly.
+
+Léonie Watson explains [why using ARIA landmark roles in this way delivers the
+best experience for screen reader users](http://tink.uk/screen-readers-aria-roles-html5-support/).
+
+This document doesn't prescribe whether to use single or double quotes in attributes,
+whether to omit values from attributes that don't need them or whether to add
+trailing slashes to void elements. However, the approach you adopt should be
+consistent throughout your code.
+
+### Header
+
+`<header role="banner">` should be used to contain your home link, branding, search,
+and any global navigation you may have.
+
+### Main content
+
+`<main role="main">` should be used to identify the main content of your page.
+
+A "Skip to main content" link should be added to help screen reader and keyboard
+users skip past your general site navigation and get straight to the content.
+You can use the [GOV.UK skip link component from the design system](https://design-system.service.gov.uk/components/skip-link/)
+for this.
+
+### Sectioned content
+
+Use the `<section>` tag to represent standalone, thematic groupings of content
+where a more specific semantic element would not be appropriate.
+
+Examples of sections would be chapters, or various tabbed pages in a tabbed dialog box,
+or a home page split into sections such as introduction, news items and contact information.
+
+You should add an `aria-labelledby` attribute linking to the id of the heading
+that identifies that section.
 
 Example:
 
-    <header role="banner"></header>
+```html
+<section aria-labelledby="introduction">
+    <h1 id="introduction">Introduction</h1>
+    <p>People have been catching fish for food since before recorded history...</p>
+</section>
 
-The following roles map directly to HTML5 sectioning elements:
+<section aria-labelledby="equipment">
+    <h1 id="equipment">Equipment</h1>
+    <p>The first thing you'll need is a fishing rod or pole that you find comfortable
+    and is strong enough for the kind of fish you're expecting to try to land...</p>
+</section>
+```
 
-* `role="main"` maps to the `<main>` element
-* `role="complementary"` maps to the `<aside>` element
-* `role="contentinfo"` maps to the `<footer>` element
-* `role="navigation"` maps to the `<nav>` element
+### Navigational elements
 
-[This article from Léonie Watson](http://tink.uk/screen-readers-aria-roles-html5-support/) explains why using ARIA landmark roles in this way delivers the best experience for screen reader users.
+Use `<nav role="navigation">` to wrap groups of links that aren't already in
+another context (e.g. `<footer>`). Common examples of navigation sections are
+menus, tables of contents, and indexes.
 
-For a more comprehensive guide to the mapping of roles to HTML5 sectioning elements read this [guide](http://blog.paciellogroup.com/2013/02/using-wai-aria-landmarks-2013/) from the Paciello Group.
+As with `<section>`, it is advisable to use `aria-labelledby` to label your `<nav>`.
+This helps to distinguish different `<nav>` blocks in the page.
 
-The following landmark roles don't map to HTML5 elements, but are
-still good to use:
+The GOV.UK blog has [more information regarding the use of the `<nav>` tag](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
 
-* `role="search"`
-* `role="application"`
+### Aside
 
-[Demo of how landmark roles help screen reader users](http://tink.co.uk/2011/07/how-do-aria-landmark-roles-help-screen-reader-users/)
+`<aside role="complementary">` should be used for content related to the primary content
+of the webpage, but which does not constitute the primary content of the page.
 
-## Main content
+Examples include author information, related links and related content.
 
-The `<main>` tag should be used to identify the main content of your page. Support for it is still marginal so ensure to add the `role="main"` attribute.
+### Footer
 
-## Sectioned content
+`<footer role="contentinfo">` should be used for the footer of the site.
 
-When using the `<section>` tag, add an `aria-labelledby` attribute linking to the id of the heading that identifies that section.
+## Individual element guidance
 
-Example:
+### Headings
 
-    <section aria-labelledby="advice">
-      <h1 id="advice">Help and advice</h1>
+Headings should be in order - for example `<h1>` then `<h2>`, but not `<h1>` then `<h3>`.
 
-## Navigational elements
+There should only be one `<h1>` element in a page.
 
-When using the `<nav>` tag, add the ARIA landmark `role="navigation"`.
+### Text emphasis
 
-The detail of using the `<nav>` tag is covered in this [blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
+Italics should be avoided according to the [GOV.UK content style guide](https://www.gov.uk/guidance/style-guide).
+Bold can be used sparingly, but can make large blocks of text difficult to read.
 
-### Emphasis
+Semantically, words voiced with emphasis can be marked up with `<em>`, whereas
+words conveying a sense or urgency or warning should be marked up with `<strong>`.
+In theory, screen readers could adopt a different tone of voice for this markup,
+though none of the major ones currently do. Browsers traditionally render these
+in italics and bold respectively, so you may need to override `em` to not use italics.
 
-If you want to give the current location emphasis within a navigational list, use a `<strong>` tag.
+As a rule, avoid using `<i>` or `<b>`, which are only useful in rare cases. It
+is usually better to use the `font-style` and `font-weight` CSS properties.
 
-Example:
+### Images
 
-    <nav role="navigation">
-      <ol>
-        <li><a href="/">Home</a></li>
-        <li><a href="/business">Business</a></li>
-        <li><strong>Business Rates</strong></li>
-      </ol>
-    </nav>
+Read the [Design System's Images section](https://design-system.service.gov.uk/styles/images/).
 
-It's semantically more accurate to use `<strong>` than `<em>` and this usage is a reasonably common convention.
+### Buttons vs Links
 
-When following this pattern ensure the text is of reasonable length. Too much text marked up in either tag can become difficult to read in some instances.
+Links should be used for navigating to another page. Buttons should be used for submitting forms
+or interactions within the page (for example, expanding an element). Empty links (`<a href="#">`)
+should therefore be avoided.
+
+Links should not open new tabs or windows by default, but if users choose to to so (via keyboard
+shortcuts or right-click context menu) then we should respect their choice.
+
+### Visually hidden elements
+
+Sometimes it can be helpful to provide additional content for screen reader users
+that might not be needed for sighted users. For example, the 'Skip to main content'
+link covered earlier.
+
+Do not use `display: none` on this text, as this will get ignored by screen readers.
+
+Instead, use CSS which ensures the text gets 'conceptually' rendered, even if the result
+is not visible on the screen. You can use the
+[Sass mixins from govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/master/package/helpers/_visually-hidden.scss).
+
+Visually hidden text is often a sign that something can be simplified or made visible to
+benefit sighted users too, so should be used sparingly.
--- a/source/manuals/programming-languages/html.html.md.erb
+++ b/source/manuals/programming-languages/html.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: HTML coding style
-last_reviewed_on: 2016-02-03
+last_reviewed_on: 2019-08-06
 review_in: 6 months
 ---
 


source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,223 @@
...
+### Self-closing tags
 
-* `role="main"` maps to the `<main>` element
-* `role="complementary"` maps to the `<aside>` element
-* `role="contentinfo"` maps to the `<footer>` element
-* `role="navigation"` maps to the `<nav>` element
+Always use well-formed HTML. Whilst [end slashes are optional in HTML5](https://www.w3.org/TR/html5/syntax.html#start-tags),
+some tags _must_ be closed, so for consistency and easier parsing, it's safer
+to always close your tags. Where it is possible to self-close a tag, use self
+closing, otherwise use normal closing.
```
alexbishop1: Not sure about this advice.

``, `<br>` and friends are void elements: they don’t have closing tags. I’m not sure the advice to self-close these tags is useful. You still have to know it’s a void element becauuse you can’t do `Foo</img>`. Similarly, you can’t have a self-closing non-void element like `<p/>` (or `<script src="foo.js"/>`, which is the one that catches lots of people out). 

The previous version of this guidance talked about [implicitly-closed elements](https://www.w3.org/TR/html5/syntax.html#optional-start-and-end-tags) like the `<dt>` and `<dd>` elements in `<dl><dt>kitten<dd>baby cat<dt>puppy<dd>baby dog</dl>`. The guidance to always close these elements with end tags is fairly useful because it reduced mental overhead.

(Also, this bit conflates tags and elements somewhat.)

ChrisBAshton: Thanks Alex, that's useful feedback 👍 I'll try and get my use of "tag" and "element" consistent.

You say you have to know whether or not an element is a void element, but that's only true of when we're _writing_ HTML. When we're reading it, we should be able to infer whether or not it's a void element through the presence of closing slashes.

For example, is `<made-up-element>` a void element or does its closing tag appear later in the document? It's clearer to do `<made-up-element />` if it's a void element, so that if we _do_ encounter `<another-made-up-element>` and there is no trailing slash, then we know it's a non-void element.

I think we're kind of saying the same thing, but I need to go back and write that portion of the guide more clearly.

ChrisBAshton: Have rewritten this section. I still advocate an explicit closing `/>`, hopefully the explanation is now clearer and the use of tag vs element now correct.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,228 @@
...
+that might not be needed for sighted users. For example, the 'Skip to main content'
+link covered earlier.
+
+Do **not** use `display: none` on this text, as this will get ignored by screen readers.
+
+Instead, use CSS which ensures the text gets 'conceptually' rendered, even if the result
+is not visible on the screen:
 
-It's semantically more accurate to use `<strong>` than `<em>` and this usage is a reasonably common convention.
+```css
```
injms: Whilst this is correct, I wonder if it would be better to point to the `.govuk-visually-hidden` and `.govuk-visually-hidden-focusable` classes in [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/package/helpers/_visually-hidden.scss)? Especially as I think using altering the `left` or `margin` means that some screenreader read things out of order.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,228 @@
...
 
-* Always quote attribute values - the decision tree for remember when to
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+HTML should be written in such a way that conforms with our [browser support model](https://github.com/alphagov/govuk-frontend#browser-support).
 
-## Use [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) roles appropriately
+For example, if IE8 is a supported browser, and it doesn't support the `<video>`
```
injms: ``
```suggestion
For example, Internet Explorer 8 doesn't support the `<video>` element. If you need to support IE8, make sure that an appropriate fallback is in place:
```

injms: This suggestion runs over two lines as you've kept to a set number characters wide - which is good for code, but not so good for content 😃

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,228 @@
...
+  Your browser does not support the video tag.
+</video>
+```
 
-When you use HTML5 sectioning elements it's a good idea to include
-the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmark roles as well. The trick when you're
-using both HTML5 and [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is to put the landmark on the sectioning
-element itself because this avoids the information being duplicated by
-screen readers that support both technologies.
+When using structual HTML5 elements, such as `<main>`, ensure that you include
```
injms: Is it worth mentioning the CSS that helps older browsers deal with HTML 5 elements?

For example:
```
article, aside, figure, footer, header, hgroup,
menu, nav, section { 
    display: block; 
}
```

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,228 @@
...
+<!-- Good -->
+<img height="30" width="30" src="foo.png" alt="logo" />
+<input type="checkbox" name="vehicle3" value="Boat" checked />
+```
 
-When using the `<section>` tag, add an `aria-labelledby` attribute linking to the id of the heading that identifies that section.
+## Document structure
 
-Example:
+We use HTML5 to structure our page semantically. Assistive technology support
```
injms: I think this statement is sweeping and should be amended; I'd suggest an alternative wording, since support isn't [marginal any more](https://tink.uk/screen-reader-support-for-html5-sections/) (and has hopefully improved since that article was published in 2013).

```suggestion
We use HTML5 to structure our page semantically. We also use ARIA roles to help older assistive technologies map the sections correctly.
```

It'd be interesting to run the [HTML5 Accessibility Page](http://www.html5accessibility.com/html5elements/) through the Empathy Lab to see how much has or hasn't improved.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,228 @@
...
 
-## HTML5
+## Progressive enhancement
 
-* Always quote attribute values - the decision tree for remember when to
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+HTML should be written in such a way that conforms with our [browser support model](https://github.com/alphagov/govuk-frontend#browser-support).
```
injms: Link to Service Manual, not to GOV.UK Frontend
```suggestion
HTML should be written in such a way that conforms with our [browser support model](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in/).
```

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,228 @@
...
+</video>
+```
 
-When you use HTML5 sectioning elements it's a good idea to include
-the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmark roles as well. The trick when you're
-using both HTML5 and [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is to put the landmark on the sectioning
-element itself because this avoids the information being duplicated by
-screen readers that support both technologies.
+When using structual HTML5 elements, such as `<main>`, ensure that you include
+a shiv that allows them to be styled in legacy browsers:
```
injms: Since a shiv is a knife, can we use a different phrase please?

```suggestion
a shim / patch that allows them to be styled in legacy browsers:
```

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,240 @@
...
+Headings should be in order, e.g. `<h1>` to `<h2>`, but not `<h1>` to `<h3>`.
+
+There should only be one `<h1>` element in a page.
+
+### Text emphasis
+
+`<b>` and `<i>` tags make text bold and italic but have no semantic meaning; it
+is purely for visual styling.
+
+`<strong>` and `<emph>` have the same effect visually, but have semantic meaning.
```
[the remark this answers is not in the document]
ChrisBAshton: Oops 🤦‍♂

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+
+Links should be used for navigating the user to a new page (hyperlinks), or to a different target
+on the same page (anchor links). `<a href="#">` should be avoided!
+
+Links should never open new tabs or windows. The user therefore retains control of their experience.
+
+Links to third-party websites should have an appropriate icon and off-screen text so that users
+know before they click.
+
+### Visually hidden elements
```
injms: On reflection, I'm wondering how much visually hidden elements are appropriate for a HTML style guide - it feels as if it'd sit better in a section on accessibility.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+presentational image. This explicitly tells the screen reader that there is no alternative text
+and the image gets skipped.
+
+Note that an empty `alt=""` is _not_ the same as having no alt attribute. If the attribute is
+missing, screen readers will read out the image filename instead, which is often not very useful.
+
+### Buttons
+
+Buttons should be used for client-side interactivity that doesn't change the destination of the
+page (e.g. a close button on a modal).
```
andysellick: If we're going to talk about buttons I'd do so in the wider context of guidance on how to build a form (I'm not sure this is covered yet?).

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+
+### Links
+
+Links should be used for navigating the user to a new page (hyperlinks), or to a different target
```
andysellick: I think this is a bit obvious.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+on the same page (anchor links). `<a href="#">` should be avoided!
+
+Links should never open new tabs or windows. The user therefore retains control of their experience.
+
+Links to third-party websites should have an appropriate icon and off-screen text so that users
```
andysellick: I'm not sure we do this.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+know before they click.
+
+### Visually hidden elements
+
+Sometimes it can be helpful to provide additional content for screen reader users
+that might not be needed for sighted users. For example, the 'Skip to main content'
+link covered earlier.
```
andysellick: I'd put this in a section on accessibility.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+width: 1px;
+height: 1px;
+margin: 0;
+padding: 0;
+overflow: hidden;
+clip: rect(0 0 0 0);
+clip-path: inset(50%);
+border: 0;
+white-space: nowrap;
+```
```
andysellick: You're already linking to the code above, might as well remove this code example - it'll save keeping it up to date if it changes.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+<nav role="navigation">
+  <ol>
+    <li><a href="/">Home</a></li>
+    <li><a href="/business">Business</a></li>
+    <li><strong>Business Rates</strong></li>
+  </ol>
+</nav>
+```
+
+### Images
```
andysellick: I'm uncertain about this section, because I think this document should both reflect what GDS does and also state best practice - but I'm not sure what GDS does with images (GOV.UK doesn't have a lot of them and I don't think we use anything much more complicated than basic img tags). Worth a discussion.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+For example, Internet Explorer 8 doesn't support the `<video>` element. If you
+need to support IE8, make sure that an appropriate fallback is in place:
 
-### Guidance on using [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmarks
+```html
+<video width="320" height="240" controls>
+  <source src="movie.mp4" type="video/mp4">
+  <source src="movie.ogg" type="video/ogg">
+  Your browser does not support the video tag.
+</video>
```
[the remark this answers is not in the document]
ChrisBAshton: Fair enough 👍this was written when I was under the impression we supported IE8, as per https://github.com/alphagov/govuk-frontend#browser-support

I now know that's out of date.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
-    <header role="banner"></header>
+## Element structure
 
-The following roles map directly to HTML5 sectioning elements:
+### Self-closing tags
 
-* `role="main"` maps to the `<main>` element
-* `role="complementary"` maps to the `<aside>` element
-* `role="contentinfo"` maps to the `<footer>` element
-* `role="navigation"` maps to the `<nav>` element
```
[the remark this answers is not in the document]
ChrisBAshton: I've moved it to each individual element in the Document Structure section.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+[End slashes are optional in HTML5](https://www.w3.org/TR/html5/syntax.html#optional-start-and-end-tags),
+meaning it is possible to write `<br>` as well as `<br />`. However, without
+web developer experience, it is impossible to know whether an element is a
+[void (self-closing)](https://www.w3.org/TR/html5/syntax.html#void-elements)
+element (e.g. `<br>`) or a container element (e.g. `<div></div>`). If you
+encounter a custom element in the DOM, e.g. `<foo>`, do you know if it is
+implicitly closed or should you be looking for a closing `</foo>` tag?
```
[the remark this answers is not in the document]
ChrisBAshton: Happy to cut it down, but isn't the point of a style guide that we outline these things? We've gone into similar levels of detail in, say, the [JavaScript guide](https://gds-way.cloudapps.digital/manuals/programming-languages/js.html).

Also happy to caveat it with "this is just a guide; consistency in your repo is more important than following the guide."

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
-The detail of using the `<nav>` tag is covered in this [blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
+### Sectioned content
 
-### Emphasis
+Use the `<section>` tag to represent standalone sections where a more specific
+semantic element would not be appropriate.
 
-If you want to give the current location emphasis within a navigational list, use a `<strong>` tag.
+You should add an `aria-labelledby` attribute linking to the id of the heading
+that identifies that section.
```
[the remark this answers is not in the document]
ChrisBAshton: I thought about that, and the previous version did have an Accessibility section, but that seems to be a throwback to the days where accessibility is tagged onto the end of a project rather than considered everywhere.

There's usually some aspect of a11y to consider at every point in our HTML, so having it in a separate section feels reductive to me 🤔

Agreed that we should talk this through at the frontend meetup.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+### Headings
+
+Headings should be in order - for example `<h1>` then `<h2>`, but not `<h1>` then `<h3>`.
+
+There should only be one `<h1>` element in a page.
+
+### Text emphasis
+
+`<b>` and `<i>` tags make text bold and italic but have no semantic meaning; it
+is purely for visual styling.
```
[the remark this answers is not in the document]
ChrisBAshton: Fair point, I'll condense it 👍

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+
+`<strong>` and `<em>` have the same effect visually, but have semantic meaning.
+In other words, the words they encapsulate are more important than the surrounding
+words.
+
+An example usage of `<strong>` is if we want to show the user their current location
+within a navigational list.
```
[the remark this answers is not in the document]
ChrisBAshton: Yeah it was a new one on me too. This was already in our style guide: https://github.com/alphagov/gds-way/pull/287/files#diff-870dd991fb2314bcc052a98935017b42L71

Happy to replace if we can think of a better example!

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Agreed, it felt a bit out of scope as I was writing it. Tried balancing best practice with what's actually required on GDS sites and this seemed a bit unnecessary.

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
injms: I don't know if it's out of date or not - best check on the Slack channel - but that looks like the browsers support by GOV.UK Frontend, rather than the browsers that the [Service Manual requires a service to support](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in).

source/manuals/programming-languages/html.html.md.erb
injms: It'd be worth seeing what @alex-ju thinks with his experience on Content Publisher and how images are used in that.

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
nickcolley: Indeed, GOV.UK Frontend has stricter standards for those teams that still need to support IE8

source/manuals/programming-languages/html.html.md.erb
nickcolley: https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
 
-Example:
+```html
+<!--[if lt IE 9]>
+  <script src="https://assets.publishing.service.gov.uk/static/ie-a4524544a53d57a7e259b4bb966b9c32557c98c920b77e52d09304642b68401a.js" crossorigin="anonymous" integrity="sha256-pFJFRKU9V6fiWbS7lmucMlV8mMkgt35S0JMEZCtoQBo="></script>
+  <style>
+    article, aside, figure, footer, header, hgroup, menu, nav, section {
+      display: block;
+    }
+  </style>
```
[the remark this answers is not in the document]
ChrisBAshton: Removed.

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
ChrisBAshton: Have cut it down and made it all about consistency, rather than prescriptive.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+be used quite sparingly.
+
+The detail of using the `<nav>` tag is covered in this [GOV.UK blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
+
+### Aside
+
+`<aside role="complementary">` should be used for content related to the primary content
+of the webpage, but which does not constitute the primary content of the page.
+
+Examples include author information, related links, related content, and advertisements.
```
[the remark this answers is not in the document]
ChrisBAshton: Done.

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
ChrisBAshton: Have condensed this section; example removed.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Thanks Nick! Useful blog post.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Done.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: This was more to labour the difference between links and buttons, as they're commonly misused (there's an open PR for [fixing this very thing in manuals-frontend](https://github.com/alphagov/manuals-frontend/pull/650))

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+    <p>The first thing you'll need is a fishing rod or pole that you find comfortable
+    and is strong enough for the kind of fish you're expecting to try to land...</p>
+</section>
+```
+
+### Navigational elements
+
+Use `<nav role="navigation">` to wrap groups of links that aren't already in
+another context (e.g. `<footer>`). As above, it is good to use `aria-labelledby`
+to disambiguate if you have multiple `<nav>` blocks, which should generally
```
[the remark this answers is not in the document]
ChrisBAshton: Done 👍

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+    <h1 id="equipment">Equipment</h1>
+    <p>The first thing you'll need is a fishing rod or pole that you find comfortable
+    and is strong enough for the kind of fish you're expecting to try to land...</p>
+</section>
+```
+
+### Navigational elements
+
+Use `<nav role="navigation">` to wrap groups of links that aren't already in
+another context (e.g. `<footer>`). As above, it is good to use `aria-labelledby`
```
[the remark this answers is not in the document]
ChrisBAshton: Reworded 👍

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
+users skip past your general site navigation and get straight to the content.
+You can use the [GOV.UK skip link component](https://design-system.service.gov.uk/components/skip-link/)
+for this.
 
-The detail of using the `<nav>` tag is covered in this [blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
+### Sectioned content
 
-### Emphasis
+Use the `<section>` tag to represent standalone sections where a more specific
+semantic element would not be appropriate.
```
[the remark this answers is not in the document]
ChrisBAshton: Done 👍

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
 title: HTML coding style
-last_reviewed_on: 2016-02-03
+last_reviewed_on: 2019-05-24
 review_in: 6 months
 ---
 
 # <%= current_page.data.title %>
 
-## HTML5
+## Progressive enhancement
```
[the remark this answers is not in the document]
ChrisBAshton: I've now made this a `Browser support` section, and added a `Progressive enhancement` subheading within that that covers JavaScript.

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
ChrisBAshton: In that case I think it's worth linking to both browser support tables, otherwise it sounds like we have one global set of supported browsers when really we do not. I've reworded this section 👍

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,263 @@
...
 
-* Always quote attribute values - the decision tree for remember when to
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+HTML should be written in such a way that conforms with our [browser support model](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in).
```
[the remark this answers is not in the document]
ChrisBAshton: Reworded 👍

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Please see https://github.com/alphagov/gds-way/pull/287/files#r289278778 - let's continue the discussion in that thread 👍

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
nickcolley: In this case, the sectioned elements similar to a `div` if it's not given a correct label. With it's label it'll become a useful landmark. So it works well where you have proposed @ChrisBAshton

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,172 @@
...
-screen readers that support both technologies.
+When using structual HTML5 elements, such as `<main>`, ensure that you include
+a shim that allows them to be styled in legacy browsers:
 
-Example:
+```html
+<!--[if lt IE 9]><script src="https://assets.publishing.service.gov.uk/static/ie-a4524544a53d57a7e259b4bb966b9c32557c98c920b77e52d09304642b68401a.js" crossorigin="anonymous" integrity="sha256-pFJFRKU9V6fiWbS7lmucMlV8mMkgt35S0JMEZCtoQBo="></script><![endif]-->
+```
+
+### Progressive enhancement
```
ChrisBAshton: TODO - remove this section, link to https://www.gov.uk/service-manual/technology/using-progressive-enhancement

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: I've cut this section right down, but may be worth cutting down even further if we don't want to go into the merits of webp and srcset in this document. Could you give this a look @alex-ju ?

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,168 @@
...
+When using structual HTML5 elements, such as `<main>`, ensure that you include
+a shim that allows them to be styled in legacy browsers:
 
-Example:
+```html
+<!--[if lt IE 9]><script src="https://assets.publishing.service.gov.uk/static/ie-a4524544a53d57a7e259b4bb966b9c32557c98c920b77e52d09304642b68401a.js" crossorigin="anonymous" integrity="sha256-pFJFRKU9V6fiWbS7lmucMlV8mMkgt35S0JMEZCtoQBo="></script><![endif]-->
+```
+
+Your page should work with just HTML, before adding anything else like CSS and
+JavaScript. Read [building a resilient frontend using progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement).
```
andysellick: Minor reword: `before adding anything else like CSS and JavaScript` I'd just write `before adding any CSS or JavaScript`.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,168 @@
...
+
+## Element structure
+
+We don't prescribe whether to use single or double quotes in attributes,
+whether to omit values from attributes that don't need them, or whether
+to add [trailing end slashes to void elements](https://www.w3.org/TR/html5/syntax.html#start-tags).
 
-    <header role="banner"></header>
+What's important is that the approach you adopt is consistent throughout
+your repository, or at the very least consistent throughout your file.
```
andysellick: This is much better and I appreciate you taking the time to rewrite. However, I'm left wondering if this section is needed at all? Thoughts?

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,168 @@
...
+### Navigational elements
 
-The detail of using the `<nav>` tag is covered in this [blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
+Use `<nav role="navigation">` to wrap groups of links that aren't already in
+another context (e.g. `<footer>`). Common examples of navigation sections are
+menus, tables of contents, and indexes.
 
-### Emphasis
+As with `<section>`, it is advisable to use `aria-labelledby` to label your `<nav>`.
+This is helpful in disambiguating where there are different `<nav>` blocks in the page.
```
andysellick: This wording feels too complex, maybe something simpler like `This helps to identify different <nav> blocks in the page.` or `This helps to distinguish different <nav> blocks in the page.` ?

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,168 @@
...
 
-If you want to give the current location emphasis within a navigational list, use a `<strong>` tag.
+The detail of using the `<nav>` tag is covered in this [GOV.UK blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
```
andysellick: Is there a way of rewording this sentence so that the link text is more descriptive? e.g. `The GOV.UK blog has some [more information on the use of the nav tag]` ?

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,168 @@
...
+
+Small, fixed size images can be implemented using the `img[src]` tag. Larger images should
+come in different sizes optimised for different devices, and [implemented responsively
+using `srcset`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images).
+
+Modern formats such as [webp](https://developers.google.com/speed/webp/) offer the best
+compression methods (faster to download, better for our users), but JPG/PNG fallbacks should
+be provided.
+
+An `alt` tag _must_ be provided, and should be used to describe the image to visually impaired
```
andysellick: An alt `attribute` must be provided. Maybe reword this to `Images must have an alt attribute`.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,168 @@
...
+users. However, it is perfectly acceptable to have an empty `alt=""` attribute if the image is
```
andysellick: Nitpick: the word 'perfectly' is unnecessary.

source/manuals/programming-languages/html.html.md.erb
andysellick: Okay, then maybe say that. `Links should be used for navigating to another page. Buttons should be used for submitting forms or interactions within the page e.g. expanding an element.`

ChrisBAshton: Done 👍

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Fair point. I've removed the section and condensed this all into a single paragraph and slipped it into the `Document structure` section. Think it's worth noting for avoidance of doubt.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Have had a think and have now removed the guidance on format and responsiveness - it's something we can add in future if we need it.

Image section now clarifies `alt` text responsibility and nothing else.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,155 @@
...
-## HTML5
+## Browser support
 
-* Always quote attribute values - the decision tree for remember when to
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+The GOV.UK Service Manual lists [browsers that it is required to support](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in). Specific
+services, such as GOV.UK Frontend, may have [stricter standards of browser
```
36degrees: GOV.UK Frontend is not a service, but is used to build services.

The guidance in the service manual sets a _minimum_ standard that services should adhere to in terms of browser support. GOV.UK Frontend has a different browser support matrix (in that we still support IE8 - 10) because we know that there are some services out there that still need to support those browsers (because it's still used by their users) and if GOV.UK Frontend didn't enable them to do that they would not be able to use it.

It might be worth thinking about what we want the user to do with this information. Why is it important to know that GOV.UK Frontend has a different browser support matrix? I don't think this is clear at the minute.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,155 @@
...
-the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmark roles as well. The trick when you're
-using both HTML5 and [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is to put the landmark on the sectioning
-element itself because this avoids the information being duplicated by
-screen readers that support both technologies.
+When using structual HTML5 elements, such as `<main>`, ensure that you include
+a shim that allows them to be styled in legacy browsers:
 
-Example:
+```html
+<!--[if lt IE 9]><script src="https://assets.publishing.service.gov.uk/static/ie-a4524544a53d57a7e259b4bb966b9c32557c98c920b77e52d09304642b68401a.js" crossorigin="anonymous" integrity="sha256-pFJFRKU9V6fiWbS7lmucMlV8mMkgt35S0JMEZCtoQBo="></script><![endif]-->
```
36degrees: This is very specific to GOV.UK. Unless you want other teams within GDS hot-linking to your assets (which I don't think you do) I'd be tempted to provide a link to something like https://github.com/aFarkas/html5shiv

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,155 @@
...
+### Main content
 
-* `role="search"`
-* `role="application"`
+`<main role="main">` should be used to identify the main content of your page.
 
-[Demo of how landmark roles help screen reader users](http://tink.co.uk/2011/07/how-do-aria-landmark-roles-help-screen-reader-users/)
+A "Skip to main content" link should be added to help screen reader and keyboard
+users skip past your general site navigation and get straight to the content.
+You can use the [GOV.UK skip link component](https://design-system.service.gov.uk/components/skip-link/)
```
36degrees: Suggest either:

```suggestion
You can use the [skip link component](https://design-system.service.gov.uk/components/skip-link/)
```

or:

```suggestion
You can use the [skip link component from the Design System](https://design-system.service.gov.uk/components/skip-link/)
```

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,155 @@
...
+There should only be one `<h1>` element in a page.
+
+### Text emphasis
+
+Use `<strong>` and `<em>` when styling text that should be read with _emphasis_.
+Use of `<b>` and `<i>` is acceptable if used purely for visual styling.
+
+### Images
+
+Images _must_ have an `alt` attribute, which is used to describe the image to visually impaired
```
36degrees: This is duplicating information [that's currently in the Design System](https://design-system.service.gov.uk/styles/images/#alternative-text) – I wonder if that can be avoided.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Agreed. 👍 thanks

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Agreed 👍 I wanted a section on browser support because when I joined GOV.UK, I wasn't sure what the browser support expectations were for the service I was building.

Initially I was under the impression that the minimum support levels of the GOV.UK Frontend were what GOV.UK supported as a whole, then someone informed me about the service standard minimum support levels and how they differ.

I've now amended this section, hopefully it is now clearer. Want to make it clear that the service manual defines the minimum browser support a service should adhere to, but that if it uses, say, the GOV.UK Frontend, any PRs to that must continue to be compatible with its own minimum standards.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
 
-## HTML5
+## Browser support
 
-* Always quote attribute values - the decision tree for remember when to
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+The GOV.UK Service Manual lists the [minimum browsers that the GDS is required
```
SteveEvansGDS: I don't think we brand GDS as 'the GDS' if it's just the acronym.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
+link covered earlier.
 
-    <nav role="navigation">
-      <ol>
-        <li><a href="/">Home</a></li>
-        <li><a href="/business">Business</a></li>
-        <li><strong>Business Rates</strong></li>
-      </ol>
-    </nav>
+Do **not** use `display: none` on this text, as this will get ignored by screen readers.
```
SteveEvansGDS: We shouldn't use bold for emphasis. Using 'Do not' is fine.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
+### Headings
+
+Headings should be in order - for example `<h1>` then `<h2>`, but not `<h1>` then `<h3>`.
+
+There should only be one `<h1>` element in a page.
+
+### Text emphasis
+
+Use `<strong>` and `<em>` when styling text that should be read with _emphasis_.
+Use of `<b>` and `<i>` is acceptable if used purely for visual styling.
```
andysellick: I don't agree with this - if you're using a tag for visual styling only on a word or phrase, then it should be done with a span element and a specific class.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
+
+### Headings
+
+Headings should be in order - for example `<h1>` then `<h2>`, but not `<h1>` then `<h3>`.
+
+There should only be one `<h1>` element in a page.
+
+### Text emphasis
+
+Use `<strong>` and `<em>` when styling text that should be read with _emphasis_.
```
andysellick: The `em` element defaults to italicised text in most browsers and I think we generally avoid that because it makes it harder for some people to read. Same with `i`.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
+```
 
-When you use HTML5 sectioning elements it's a good idea to include
-the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmark roles as well. The trick when you're
-using both HTML5 and [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is to put the landmark on the sectioning
-element itself because this avoids the information being duplicated by
-screen readers that support both technologies.
+When using structual HTML5 elements, such as `<main>`, ensure that you include
+a [shim that allows them to be styled](https://github.com/aFarkas/html5shiv) in
+legacy browsers. You can include the script inside a `<!--[if lt IE 9]><![endif]-->`
```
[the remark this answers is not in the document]
injms: Are we sure that that the the html5 shim is currently used anywhere? I haven't come across it on the projects that I've worked on, so happy to be proved wrong. It also seems a little prescriptive - if we do need to include a section on shims, it feels like a line on why they might be useful could be better.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
-  selectively quote is too much mental overhead, so just quote them
-* Always use well-formed HTML - implicitly closed elements such as
-  `<p>` cause unnecessary parsing overhead and can cause issues in
-  accessibility
+The GOV.UK Service Manual lists the [minimum browsers that GDS is required
+to support](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in).
+Tools such as GOV.UK Frontend, that are used to help build services, may have
+[stricter standards of browser support](https://github.com/alphagov/govuk-frontend#browser-support).
+Therefore, you will need to test your code in different browsers depending on
+which part of GDS you are working on.
```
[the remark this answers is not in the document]
injms: I'd suggest that lines 13 to 16 can be removed without any adverse effect.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
 
-The following roles map directly to HTML5 sectioning elements:
+Léonie Watson explains [why using ARIA landmark roles in this way delivers the
+best experience for screen reader users](http://tink.uk/screen-readers-aria-roles-html5-support/).
 
-* `role="main"` maps to the `<main>` element
-* `role="complementary"` maps to the `<aside>` element
-* `role="contentinfo"` maps to the `<footer>` element
-* `role="navigation"` maps to the `<nav>` element
+This document doesn't prescribe whether to use single or double quotes in attributes,
```
injms: I think that lines 47 to 50 are a bit too fussy and could be replaced with 'The project should be consistent in its code style'.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,153 @@
...
+Use of `<b>` and `<i>` is acceptable if used purely for visual styling.
+
+### Images
+
+Read the [Design System's Images section](https://design-system.service.gov.uk/styles/images/).
+
+### Buttons vs Links
+
+Links should be used for navigating to another page. Buttons should be used for submitting forms
+or interactions within the page (for example, expanding an element). `<a href="#">` should therefore
```
injms:
```suggestion
or interactions within the page (for example, expanding an element). Empty links, for example `<a href="#">`, should
```

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: The example in that doc suggests using it for drawing visual attention to individual keywords, e.g.

> The two most popular science courses offered by the school are `<b class="term">`chemistry`</b>` (the study of chemicals and the composition of substances) and `<b class="term">`physics`</b>` (the study of the nature and properties of matter and energy).

But as you say, discourages using `<b>` for general styling:

> However, you should not use `<b>` for styling text; instead, you should use the CSS `font-weight` property to create boldface text, or the `<strong>` element to indicate that text is of special importance.

I should rephrase the sentence to make it clear that it's only about drawing visual attention to individual words 👍

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
ChrisBAshton: The `aFarkas/html5shiv` isn't being used on GOV.UK, but we do have an internal one:

```html
<!--[if lt IE 9]><script src="https://assets.publishing.service.gov.uk/static/ie-a4524544a53d57a7e259b4bb966b9c32557c98c920b77e52d09304642b68401a.js" crossorigin="anonymous" integrity="sha256-pFJFRKU9V6fiWbS7lmucMlV8mMkgt35S0JMEZCtoQBo="></script><![endif]-->
```

This was in the original PR but was removed because [we don't want other teams hot-linking to our assets](https://github.com/alphagov/gds-way/pull/287/files/a3555155d27173262e110a19f86311c1194f853c#diff-870dd991fb2314bcc052a98935017b42) (I also don't know how that asset is generated or how long it's expected to exist - in my previous company, old assets would expire as new versions are released).

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Have gone down a rabbit hole re-educating myself about `<strong>` and `<em>` and `<b>` and `<i>` and have boiled it down to this:

```md
You should avoid bold or italics for blocks of text as it can make it more
difficult to read.

Use `<em>` when styling words that should be read with _emphasis_. The emphasised
words is read in a different tone of voice by screen readers. Browsers traditionally
render English `<em>` text in italics, but other languages may denote emphasis
differently.

Use `<strong>` for content of strong importance or urgency (such as warnings). Again,
`<strong>` text is read in a different tone of voice by screen readers, and again,
browsers traditionally render English `<strong>` text in a bold style, but other
languages may denote importance differently.

As a general rule, avoid using `<i>` or `<b>`. They are purely to apply visual
styling without additional meaning, which is better achieved using the `font-style`
and `font-weight` CSS properties. Keeping presentation information separate from
your HTML content is considered best practice.
```

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: I like that the current version is unambiguous - when I first joined, I wondered what the house style was in terms of quotes and self-closing elements and so on. I think "should be consistent in its code style" doesn't necessarily reassure me that, say, it's ok to not apply trailing slashes.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Have now added a warning in this section that too much italicised text is hard to read.

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
ChrisBAshton: I'm kind of on the fence on that one - I see where you're coming from, but also this is the section on Browser Support and I think it's helpful to know that your milage may vary. (Before I go and make a PR to govuk-frontend which is only compatible in IE9+, thinking that's the global baseline of browser support across GDS, hopefully I'd read this guidance and realise it's service-specific).

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,165 @@
...
+
+## Individual element guidance
+
+### Headings
+
+Headings should be in order - for example `<h1>` then `<h2>`, but not `<h1>` then `<h3>`.
+
+There should only be one `<h1>` element in a page.
+
+### Text emphasis
```
andysellick: I appreciate that you took the time to rewrite this section but I think it's unnecessary. I'm sorry, I should have been clearer about that from the start.

ChrisBAshton: Thanks for looking. I think there's value in having this section, since this document is about HTML markup and it's so common to see em/strong/b/i misused. Worth a discussion in the next frontend meeting?

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,165 @@
...
 
-When you use HTML5 sectioning elements it's a good idea to include
-the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) landmark roles as well. The trick when you're
-using both HTML5 and [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is to put the landmark on the sectioning
-element itself because this avoids the information being duplicated by
-screen readers that support both technologies.
+When using HTML5 elements such as `<main>`, ensure that you include a
+[shim that allows them to be styled](https://github.com/aFarkas/html5shiv) in
+legacy browsers. Include the script inside a `<!--[if lt IE 9]><![endif]-->`
+special comment so that modern browsers do not download them.
```
Nooshu: "conditional comment" rather than "special comment"

source/manuals/programming-languages/html.html.md.erb
tombye: I wasn't in the frontend meeting mentioned so not sure about what was decided but don't the deleted points that started on line 11 define a style? Excuse if their deletion was a decision already made.

Also worth noting that we use a fair amount of testing tools which parse our HTML using XML parsing libraries so not quoting attributes or closing elements (making them valid XML) would limit our use of them.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: In meeting:

- don't use italics - not supported by our font
- screen reader spec says items read with emphasis, but not implemented
- link to content style guide for GOV.UK
- @selfthinker to review guide

ChrisBAshton: Have now pushed up the changes.

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: @tombye I followed the original guide when writing some HTML and was wondering whether to use single or double quotes, and there were other files in the project that used no quotes at all (so apparently not following the guidance).

In trying to get an answer, it was discussed about 2 months ago at the frontend meeting and the collective decision was that consistency is the most important thing and that we shouldn't prescribe the style.

You make a good point about tooling. I suppose if we're not prescribing the style then the developer is free to adopt whatever style that is demanded by their tooling.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+`<main role="main">` should be used to identify the main content of your page.
 
-[Demo of how landmark roles help screen reader users](http://tink.co.uk/2011/07/how-do-aria-landmark-roles-help-screen-reader-users/)
+A "Skip to main content" link should be added to help screen reader and keyboard
+users skip past your general site navigation and get straight to the content.
+You can use the [GOV.UK skip link component from the design system](https://design-system.service.gov.uk/components/skip-link/)
+for this.
 
-## Main content
+### Sectioned content
```
selfthinker: I wonder if this might confuse people because GOV.UK doesn't use any `<section>`s anywhere.
Not a blocker at all, especially as sections have been in the text before, but something worth discussing. If we tell people to use it, why don't we use it ourselves?

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+Use `<nav role="navigation">` to wrap groups of links that aren't already in
+another context (e.g. `<footer>`). Common examples of navigation sections are
+menus, tables of contents, and indexes.
 
-### Emphasis
+As with `<section>`, it is advisable to use `aria-labelledby` to label your `<nav>`.
+This helps to distinguish different `<nav>` blocks in the page.
 
-If you want to give the current location emphasis within a navigational list, use a `<strong>` tag.
+The GOV.UK blog has [more information regarding the use of the `<nav>` tag](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/).
```
selfthinker: I wonder if the main gist of that article could be added:
> But don't use too many `<nav>` elements per page.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
 
-Example:
+### Aside
+
+`<aside role="complementary">` should be used for content related to the primary content
+of the webpage, but which does not constitute the primary content of the page.
```
selfthinker: The phrasing of this is a bit awkward and difficult to understand. Suggestion:
> `<aside role="complementary">` should be used for blocks of content that are separate from the main content.
(Not a blocker.)

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+
+Examples include author information, related links and related content.
```
selfthinker: The examples should be as clear as possible, especially as when to use `aside` is confusing to a lot of people. The most typical example is a sidebar. Something like "author information" could lead to people using it wrongly.
(Not a blocker.)

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+should therefore be avoided.
+
+Links should not open new tabs or windows by default, but if users choose to to so (via keyboard
+shortcuts or right-click context menu) then we should respect their choice.
+
+### Visually hidden elements
+
+Sometimes it can be helpful to provide additional content for screen reader users
+that might not be needed for sighted users. For example, the 'Skip to main content'
+link covered earlier.
```
selfthinker: The skip link is not the best example as it's mostly for keyboard only users although some screen reader users use it as well. So, this particular example is actually needed for sighted users, it's only hidden until it has focus.
(Not a blocker.)

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+Do not use `display: none` on this text, as this will get ignored by screen readers.
 
-It's semantically more accurate to use `<strong>` than `<em>` and this usage is a reasonably common convention.
+Instead, use CSS which ensures the text gets 'conceptually' rendered, even if the result
+is not visible on the screen. You can use the
+[Sass mixins from govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/master/package/helpers/_visually-hidden.scss).
 
-When following this pattern ensure the text is of reasonable length. Too much text marked up in either tag can become difficult to read in some instances.
+Visually hidden text is often a sign that something can be simplified or made visible to
+benefit sighted users too, so should be used sparingly.
```
selfthinker: Good point!

source/manuals/programming-languages/html.html.md.erb
selfthinker: We have a big problem with alt texts, most people don't understand how it should be written. (And many experts don't even agree on rules.) But the GOV.UK content style guide and/or the Design System are probably the better place to have that kind of information. In the end it's much more related to content than to code.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+### Text emphasis
+
+Italics should be avoided according to the [GOV.UK content style guide](https://www.gov.uk/guidance/style-guide).
+Bold can be used sparingly, but can make large blocks of text difficult to read.
+
+Semantically, words voiced with emphasis can be marked up with `<em>`, whereas
+words conveying a sense or urgency or warning should be marked up with `<strong>`.
+Browsers traditionally render these in italics and bold respectively, so you may
+need to override `em` to not use italics. In theory, screen readers could adopt
+a different tone of voice for this markup, though none currently do.
```
selfthinker: I cannot say for certain that none support this. Maybe saying "none of the major ones currently do" makes that less tenuous. Unless @accessiblewebuk knows more for certain. (He tested this in a couple of screen readers a year or two ago.)
(Not a blocker.)

source/manuals/programming-languages/html.html.md.erb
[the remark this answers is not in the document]
selfthinker: A proper fallback could be a link to the video so a user could download it and open in a video player of their choice.

source/manuals/programming-languages/html.html.md.erb
```diff
@@ -1,85 +1,162 @@
...
+
+## Document structure
 
-    <header role="banner"></header>
+Use HTML5 to structure your page semantically. Use ARIA roles to help
+older assistive technologies map the sections correctly.
 
-The following roles map directly to HTML5 sectioning elements:
+Léonie Watson explains [why using ARIA landmark roles in this way delivers the
+best experience for screen reader users](http://tink.uk/screen-readers-aria-roles-html5-support/).
```
selfthinker: Off-topic for this PR, but it would be good to know which browser/AT combination needs that extra support so we know when we can remove it.

source/manuals/programming-languages/html.html.md.erb
selfthinker: I have reviewed everything, left some comments, some about accessibility, some not. But none of what I raised is a blocker. :+1:

source/manuals/programming-languages/html.html.md.erb
selfthinker: I have tried to find out if there are any which support it. It seems like some of the major ones only support it in a roundabout way (e.g. saying "bold" for `strong` as well as `b` but only on inspection, so not by default). It seems some non-major screen readers might support it (although it's not clear which ones), but they would treat strong/b/em/i all the same.

More info in these articles:
* [Screen Readers lack emphasis](https://developer.paciellogroup.com/blog/2008/02/screen-readers-lack-emphasis/) from 2008
* [Screenreader support for text-level semantics](https://www.brucelawson.co.uk/2018/screenreader-support-for-text-level-semantics/) from 2018
* [Thread: bold vs. strong, italics vs. emphasis](https://webaim.org/discussion/mail_thread?thread=5641) from 2012

andysellick: If we can't be sure of this it might be better to remove this paragraph?

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: `aside` _is_ confusing. It can be used for sidebars, but should only be used if the content of the sidebar is "tangentially related to the content of the parenting sectioning content". So if it's just a sidebar of adverts or newsletter signup forms - as sidebars often are - then `aside` would probably not be the most appropriate choice.

Think this might be worth following up in a separate issue. 👍

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Separate _but related_ - see https://github.com/alphagov/gds-way/pull/287#discussion_r310963175

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Fair enough. Though hard to think of a concise example otherwise - perhaps something like:

```
Sometimes it can be helpful to provide additional content for screen reader users
that might not be needed for sighted users. For example, contextual descriptions
for links, so that they can be understood in isolation:

`<a href="/fishing">Read more <span class="sr-only">about fishing</span></a>".
```

ChrisBAshton: Fair enough. Though hard to think of a concise example otherwise - perhaps something like:

```
Sometimes it can be helpful to provide additional content for screen reader users
that might not be needed for sighted users. For example, contextual descriptions
for links, so that they can be understood in isolation:

`<a href="/fishing">Read more <span class="sr-only">about fishing</span></a>".
```

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Agreed 👍

source/manuals/programming-languages/html.html.md.erb
ChrisBAshton: Have now changed to "none of the major ones currently do". Agreed that it's worth revisiting later whether or not this is worth advocating for, given the lack of support in screen readers.

Nooshu:
Nice work @ChrisBAshton. Looks good to merge. Any future issues can be raised as a separate issues / PR's.

PR #291, opened 2019-06-11
Update guidance to re-affirm LTS as default.

stephenharker:
The guidance here did say it was ok to use standard Ubuntu releases but
I think this has to be avoided if possible and instead use backports and
PPA sources to update only those parts of the LTS release that need it.


source/standards/operating-systems.html.md.erb
```diff
@@ -16,6 +16,6 @@ You must also make sure your OS is:
...
+GDS recommends LTS as the default choice for your VM. If you are running the latest LTS release and there are newer features not available in that release, for instance newer kernels, consider using the official backports or trustworthy PPA sources rather than upgrading the whole OS to a standard (non-LTS) release. Care must be taken to ensure that these backports and other sources receive regular and timely security updates but this should be less burden than having to continually (every 6-9 months) update the OS on VMs.
 
-If you use standard Ubuntu, we recommend you migrate to the latest release within 3 months to ensure you stay within the support period, and receive security updates.
+If you do use standard Ubuntu, it is essential to migrate to the latest release within 3 months to ensure you stay within the support perid and receive security updates. This must be considered carefully when proposing adopting standard Ubuntu releases as it can affect the stability of your running service. Standard releases only receive security updates for 9 months after their initial release whereas LTS receives support for 5 years (or longer with paid-for extended support).
```
rhowe-gds:
```suggestion
If you do use standard Ubuntu, it is essential to migrate to the latest release within 3 months to ensure you stay within the support period and receive security updates. This must be considered carefully when proposing adopting standard Ubuntu releases as it can affect the stability of your running service. Standard releases only receive security updates for 9 months after their initial release whereas LTS receives support for 5 years (or longer with paid-for extended support).
```

PR #298, opened 2019-06-14
Update Ruby style guide

SteveEvansGDS:
Added a line about testing with RSpec

### DIFF
--- a/source/manuals/programming-languages/ruby.html.md.erb
+++ b/source/manuals/programming-languages/ruby.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Ruby style guide
-last_reviewed_on: 2018-11-07
+last_reviewed_on: 2019-06-14
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


kevindew:
> Looks good to me, assuming there aren't projects using minitest

Unfortunately there are a bunch of them: [Whitehall](https://github.com/alphagov/whitehall), [Govspeak](https://github.com/alphagov/govspeak), [Government-frontend](https://github.com/alphagov/government-frontend). These are mostly older projects as new projects would almost certainly be started with RSpec, though these are not considered tech debt either.

If anything I'm not sure the page we have on [RSpec](https://gds-way.cloudapps.digital/standards/testing-with-rspec.html#what-do-we-use-rspec-for) really belongs in the gds-way. I don't think we have other pages on language testing frameworks; it references GOV.UK common patterns rather than standards; and has a recommendation that is [uncommon and unconventional](https://gds-way.cloudapps.digital/standards/testing-with-rspec.html#how-to-configure-rspec).

PR #316, opened 2019-08-15
Update JavaScript documentation to bring it up-to-date

nickcolley:
Updates the guidance to reflect the current state of JavaScript at GDS.

Encourages moving away from jQuery.

See individual commits for details.

### DIFF
--- a/source/manuals/programming-languages/js.html.md.erb
+++ b/source/manuals/programming-languages/js.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: JavaScript coding style
-last_reviewed_on: 2017-01-01
+last_reviewed_on: 2019-08-15
 review_in: 6 months
 ---
 
--- a/source/manuals/programming-languages/js.html.md.erb
+++ b/source/manuals/programming-languages/js.html.md.erb
@@ -19,6 +19,7 @@ review_in: 6 months
 * [Modules](#modules)
 * [Module structure](#module-structure)
 * [jQuery](#jquery)
+* [Supporting older browsers](#supporting-older-browsers)
 * [Method arguments](#method-arguments)
 * [Let the project define the style](#let-the-project-define-the-style)
 
@@ -249,6 +250,12 @@ In older projects put together a plan to migrate away from jQuery.
 
 **Why:** jQuery has been used to provide browser support for older browsers.
 The older versions of jQuery that we use have security vulnerabilities and are no longer maintained by the jQuery team.
+
+## Supporting older browsers
+
+Use native web APIs where possible.
+
+Use [feature detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection) before [polyfilling](https://developer.mozilla.org/en-US/docs/Glossary/Polyfill), to support older browsers.
 
 ## Method arguments
 
--- a/source/manuals/programming-languages/js.html.md.erb
+++ b/source/manuals/programming-languages/js.html.md.erb
@@ -228,7 +228,8 @@ Avoid jQuery in new projects.
 
 In older projects put together a plan to migrate away from jQuery.
 
-**Why:** jQuery has been used to provide browser support for older browsers.
+**Why:** jQuery had been used to provide browser support for older browsers.
+However, browser support for ES5 JavaScript is now widespread enough that a library like jQuery is unnecessary.
 The older versions of jQuery that we use have security vulnerabilities and are no longer maintained by the jQuery team.
 
 ## Supporting older browsers
--- a/source/manuals/programming-languages/js.html.md.erb
+++ b/source/manuals/programming-languages/js.html.md.erb
@@ -169,14 +169,16 @@ code before. This can cause people to misunderstand what a line is doing.
 
 ## Modules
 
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/js.html.md.erb
```diff
@@ -241,36 +215,18 @@ makes it possible to test those functions in isolation.
 
 ## jQuery
 
-* Prefix jQuery objects with a `$`.
-
-  ```javascript
-  // Bad
-  var list = $('.list')
+Avoid jQuery in new project.
```
nickcolley:
```suggestion
Avoid jQuery in new projects.
```

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -241,36 +215,18 @@ makes it possible to test those functions in isolation.
...
-  var $list = $('list')
-  $list.click(...)
-  $list.addClass(...)
-  ```
+Use native web APIs where possible.
 
-  **Why:** DOM queries are slow operations, especially if they are complicated.
-  By caching the result of a jQuery object in a variable it reduces the number
-  of queries you have to perform on the document.
+Use [feature detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection) before [polyfilling](https://developer.mozilla.org/en-US/docs/Glossary/Polyfill), to support older browsers.
```
nickcolley: Maybe this needs to be expanded more?

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -110,14 +110,6 @@ Use soft tabs with a two space indent.
   sets the appropriate affordances. It also follows the conventions of the
   standard library.
 
-## CoffeeScript
```
tijmenb: I guess this was in here because it's still the [default thing in Rails](https://guides.rubyonrails.org/working_with_javascript_in_rails.html). Could we replace it with a generalised section on using compiles-to-js languages? (For example, what do we think about TypeScript?)

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -232,36 +232,12 @@ makes it possible to test those functions in isolation.
...
-  var $list = $('list')
-  $list.click(...)
-  $list.addClass(...)
-  ```
+In older projects put together a plan to migrate away from jQuery.
 
-  **Why:** DOM queries are slow operations, especially if they are complicated.
-  By caching the result of a jQuery object in a variable it reduces the number
-  of queries you have to perform on the document.
+**Why:** jQuery has been used to provide browser support for older browsers.
```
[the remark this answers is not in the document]
nickcolley: Good point about ES5, worth clarifying this for sure.

source/manuals/programming-languages/js.html.md.erb
nickcolley: Hmm maybe we just leave the section in I wasnt aware it's still a default.

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -110,14 +110,6 @@ Use soft tabs with a two space indent.
...
-
-Don't use CoffeeScript.
-
-**Why:** It's an extra abstraction and introduces another
-language for developers to learn. Using JavaScript gives us guaranteed
-performance characteristics and more well known support paths.
-
 ## HTML class hooks
 
 When attaching JavaScript to the DOM use a `.js-` prefix for the HTML classes.
```
injms: With v3 of Frontend using the namespace '.govuk-js', is it sensible to also change the JavaScript prefix advice changed to use '.app-js-' or '.gem-js'?

nickcolley: I'm trying to keep his pull request is limited to factual changes only, additions like that I think can be done as a separate pull request in the future.

Think it's important to merge smaller things rather than having very big pull requests open for long periods of time.

injms: Sure - make sense. But the flip side is that it's highly likely it won't happen until the next review is due.

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -241,36 +224,19 @@ makes it possible to test those functions in isolation.
...
-  // Bad
-  $('.list').click(...)
-  $('.list').addClass(...)
-
-  // Good
-  var $list = $('list')
-  $list.click(...)
-  $list.addClass(...)
-  ```
+Use native web APIs where possible.
```
alex-ju: ❤️

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -241,36 +224,19 @@ makes it possible to test those functions in isolation.
...
 
-  **Why:** DOM queries are slow operations, especially if they are complicated.
-  By caching the result of a jQuery object in a variable it reduces the number
-  of queries you have to perform on the document.
+Use [feature detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection) before [polyfilling](https://developer.mozilla.org/en-US/docs/Glossary/Polyfill), to support older browsers.
```
alex-ju: ❤️

PR #318, opened 2019-08-29
Update page on 'Linting Python at GDS'

lfdebrux:
Update the page on Python linting, specifically the advice on using
Flake8:

- Digital Marketplace no longer pins Flake8 or uses Flake8-Putty
- Per-file-ignores are now built in to Flake8

### DIFF
--- a/source/manuals/programming-languages/python/linting.html.md.erb
+++ b/source/manuals/programming-languages/python/linting.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Linting Python at GDS
-last_reviewed_on: 2019-02-08
+last_reviewed_on: 2019-08-29
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/python/linting.html.md.erb
```diff
@@ -73,22 +73,23 @@ or warnings for upcoming deprecations.
...
+[per-file-ignores][], you can see an example [here][DMAPI-flake8-config].
 
 
 ### Common Configuration
 
-Digital Marketplace is already running [Flake8][] on all of it's repos pinned at v2.6.2 with
-[Flake8 Putty][flake8-putty]. You can find an example of their configuration in the root of any repo in the
-[`.flake8` file][DMAPI-flake8-config].
+Digital Marketplace is already running the latest version of [Flake8][] on all
+of it's repos.  You can find an example of their configuration in the root of
```
jonodrew: I think this should be `its`

lfdebrux: Do'h!

PR #321, opened 2019-08-30
Remove nginx.conf from gds-way app

lfdebrux:
We are using the Staticfile buildpack, which although it does support a custom nginx.conf, complains if you do, so this PR removes it.

This also means that the custom redirects are gone:

- The redirect to force host to be gds-way is no longer there. This might actually be beneficial, as it will make it easier to deploy alternate versions of the app to PaaS.

- The redirect to the RE Logit.io manual is probably no longer needed, as there are no links within the manual to the offending page

@richardtowers thinks that removing these redirects should be okay (https://gds.slack.com/archives/CACFQBL0Y/p1567176214007200).

I've tested this by deploying to PaaS at https://gds-way-ldeb-pr-deploys.cloudapps.digital.


lfdebrux:
For future reference, if we want to restore or add redirects, middleman can do this https://middlemanapp.com/basics/redirects/

PR #328, opened 2019-09-10
Date bump only, no content changes

SteveEvansGDS:
Is the info about RE running a beta on Prometheus still correct?


deanwilson:
> Is the info about RE running a beta on Prometheus still correct?

The answer is complicated and we're going to bump it until the next review period.

PR #352, opened 2019-11-11
Revert "Add guidance on writing Dockerfiles"

cmcnallygds:
Reverts alphagov/gds-way#335

reverting while clarifying content rules re pre-election period


rhowe:
This is hilarious

PR #361, opened 2019-12-17
Update dependencies

pritchyspritch:
Add more information regarding security advisories and dependabot usage


source/standards/tracking-dependencies.html.md.erb
```diff
@@ -36,9 +41,14 @@ You should monitor for potential vulnerabilities in every layer of your technolo
 
 Most of the tooling that helps you stay on top of dependency updates will also highlight vulnerabilities. Additional tooling includes:
 
-* [GitHub security alerts][] - This can be configured per repository in the settings tab of the GitHub UI. When GitHub discovers or is informed about a vulnerability, it will email an alert to the repository owner and users with admin access
+* [GitHub security alerts][] - This can be configured per repository in the settings tab of the GitHub UI. When GitHub discovers or is informed about a vulnerability, it will email an alert to the repository owner and users with admin access.
+* [GitHub security alerts][] are turned on by default on public repositories and are being automatically turned on once a day for every Alphagov repository. If services wish to opt out of security advisories on their repository, they can [contact cyber security](https://gds.slack.com/archives/CCMPJKFDK) and then add the `no-security-advisories` tag to their repository.
```
louzoid-gds: > and are being automatically turned on once a day for every Alphagov repository

To be clear, does this mean that notifications will be sent once a day? I'm unclear on precisely what that means.
Also, should this bit be its own bullet point? May be best included in the one above. Just a preference that is not strongly held :)

pritchyspritch: We can make it a bullet point for sure. It means that we are making sure that security alerts are turned on for every repository in alphagov, whether it's private/public/forked. For example, when users create a new repository that's private they would normally need to turn on security advisories automatically, but this will now be turned on automatically. For all existing repositories, they will not receive any more notifications than they already do. The functionality allows us to audit the whole organisation for security issues in our dependencies.

louzoid-gds: Oh right I get you now. Thanks for clarifying! In that case, I'm wondering if we could compress both bullets points into just one saying something like:

> GitHub security alerts - When GitHub discovers or is informed about a vulnerability, it will email an alert to the repository owner and users with admin access. GitHub security alerts will be turned on for all Alphagov repositories. If services wish to opt out of security advisories on their repository, they can contact cyber security and then add the no-security-advisories tag to their repository.

PR #379, opened 2020-01-08
Update production-readiness.html.md.erb

cmcnallygds:
Date bump and minor style changes - needs technical review

### DIFF
--- a/source/standards/production-readiness.html.md.erb
+++ b/source/standards/production-readiness.html.md.erb
@@ -1,12 +1,12 @@
 ---
 title: Check if your service is ready for production
-last_reviewed_on: 2019-05-01
+last_reviewed_on: 2020-01-08
 review_in: 6 months
 ---
 
 # <%= current_page.data.title %>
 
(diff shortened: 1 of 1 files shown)


source/standards/production-readiness.html.md.erb
```diff
@@ -16,11 +16,11 @@ Read the [Service Manual][] for more information about building and running a li
 
 ## Use a checklist
 
-To avoid problems when your service goes into production use a checklist to confirm if your service is ready. Each check should reference a standard that’s already in place before you can put a service into production. An example is, [documenting architecture decisions].
+To avoid problems when your service goes into production, use a checklist. Each check should reference a standard that’s already in place before you can put a service into production. An example is [documenting architecture decisions].
```
cmcnallygds: Not entirely sure what this means: 'Each check should reference a standard that’s already in place before you can put a service into production.' Can it be reworded?

erino:
We don't need this content, and will be removed in https://github.com/alphagov/gds-way/pull/405

PR #381, opened 2020-01-09
Update java.html.md.erb

cmcnallygds:
Some minor style changes and date bump. Needs technical review to check if content is still accurate.


alexbishop1:
Reviewing the content in the Java style guide is on the agenda for the next Java community meeting, which is scheduled for Tuesday next week (we intended to do it in the November meeting but it was cancelled).

The style updates look mostly fine but I’m not sure about some of the changes from “that” to “which” in restrictive clauses.

PR #391, opened 2020-02-05
accounts-with-third-parties is still valid.

deanwilson:


source/standards/accounts-with-third-parties.html.md.erb
```diff
@@ -1,14 +1,14 @@
 ---
 title: How to manage access to your third-party service accounts
-last_reviewed_on: 2019-08-01
+last_reviewed_on: 20120-02-05
```
Krenair:
```suggestion
last_reviewed_on: 2020-02-05
```
Suspect Concourse will have better support for showing PR build failures in the 202nd century

PR #408, opened 2020-03-04
Clarify what needs release notes

heathd:


source/manuals/release-notes.html.md.erb
```diff
@@ -6,7 +6,16 @@ review_in: 6 months
 
 # <%= current_page.data.title %>
 
-When you [publish a new version of your code](/standards/publish-opensource-code.html), you should also publish a release note that tells users what’s changed and how to update from their current version.
+For software systems which third parties depend on it's important to communicate clearly about changes to the functionality, semantics or syntax of your service.
```
m-green: Hi @heathd - thanks for this great suggestion! I’ve rejigged it for style below - I hope I’ve still captured everything you wanted to. Happy to revisit it if you’d like me to.

---

When you [publish a new version of your service](/standards/publish-opensource-code.html), you should also publish a release note that clearly tells your users what’s changed.

You should publish release notes if your service is either:

- a public API that third parties use
- published as a downloadable package, for example an API client library

You can also publish release notes if your service is used only internally, or used by people rather than through an API.

## Writing a release note

heathd: As a reader I prefer to be able to scan some text and know whether or not it applies to me

Saying “You should do X”.... “You only need to do X in circumstance Y” makes that much harder.

I prefer to say "In circumstance Y you should do X" then when I'm reading it I can quickly stop reading if I know it's not relevant to my situation. This seems to meet user needs for readers much better. I also think that for this kind of guidance documentation it's important to talk about WHY as well as WHAT. I disagree with the aim of eliminating all WHY from the conversation I think this makes for much poorer guidance.

I think the statement "For software systems which third parties depend on it's important to communicate clearly about changes to the functionality, semantics or syntax of your service. " communicates clearly and succinctly both the why and the scope of this guidance.

m-green:
```suggestion
When you [publish a new version of your code](/standards/publish-opensource-code.html), you should also publish a release note that clearly explains what’s changed about your service’s functionality, semantics or syntax.

This is so that third parties that depend on your service can find out how the changes affect them.
```

Hi @heathd - definitely not trying to eliminate all ‘why’ from the content! Sorry if it came across that way.

I’ve had another go at the first couple of paras. Sorry I couldn’t use your suggested statement as is - I’ve aimed to use as much as possible while keeping to GOV.UK style and active voice. Also I wasn’t sure whether your suggested statement had a ‘why’ (beyond “it’s important”) so I’ve had a go at fleshing it out a bit. Hope it looks ok.

I do appreciate it might seem intuitive that "In circumstance Y you should do X" would meet user needs better or be easier to scan, but the evidence we have doesn’t agree I’m afraid. Especially where there are multiple circumstances as in this example, because they make for a complicated sentence if you lead with them. I can share more research if it helps (including some recent findings from GOV.UK Frontend research with developers)?

heathd: On reflection I think my problem is that the wording proposed ("When you publish a new version of your code, you should also publish a release note...") orients the content towards a generalist audience, wheras the actual audience for this guidance is engineering staff in GDS. 

I think the act of "publishing" is also generally not what an engineer would think that they do most of the time when they make changes to code. Engineers are making commits to the code on a daily basis which introduce changes. Most of the time the code they are changing is not depended on by third parties and so release notes are not needed and would be a significant burden to maintain.

How about:

"When you make changes to code that will affect third parties, you should also publish a release note that clearly explains what’s changed about your service’s functionality, semantics or syntax."

There are very few GDS projects which have external APIs, and even fewer which provide client libraries or other software distributed for execution on a third party system. So placing the least common case first seems unhelpful for the intended audience.

Saying "You can also publish release notes if your service is used only internally, or used by people rather than through an API." is a truism which seems pointless. 

It would be more helpful to say "You _don't need to_ publish release notes if your service is used only internally, or used by people rather than through an API".

m-green: Your suggestions sounds good!

On the second sentence, "You don't need to..." might sound too much like you should *not* publish release notes in those instances? Whereas "You can..." is usual GOV.UK style for "This is optional". So would recommend that tweak.

So we'd be left with:
"When you make changes to code that will affect third parties, you should also publish a release note that clearly explains what’s changed about your service’s functionality, semantics or syntax."
"You can also publish release notes if your service is used only internally, or used by people rather than through an API."

(The only other thing I'd say is that "explain what's changed about your service’s functionality, semantics or syntax" seems tautological. Would just "explain what's changed." do the job? We can leave that discussion for another day though...)

I can update the PR and send this through 2i if you're happy with the above?

heathd: Hi Mark, 

sorry I don't agree at all that "You can..." is a good way to express "You don't need to..."

it's too ambiguous

Let's stick with "You don't need to..."

m-green:
Updated following discussion - needs 2i before merging.

SteveEvansGDS (approved):

PR #428, opened 2020-05-15
Add extra notes that detail how we use Dependabot

benthorner:
Previously we had regular discussion on GOV.UK about how we can
make Dependabot less burdensome. This adds and consolidates the
notes on Dependabot to record the outcome of those discussions,
so we don't keep having them again, and again, and...

This PR was at one point a discussion area for a potential change for 
an account setting. However, we decided to leave it as-is, and instead 
add a note to explain why.

> At the time of raising this PR, the note about the setting to "Treat PR 
approval as a request to merge" is still under discussion. This PR is also 
serves as a place to continue that discussion. Some useful facts about 
this setting, from previous discussions:
>
> - Will the commits be signed? [Yup, by GitHub](https://github.com/alphagov/frontend/commit/f455693b034274815e73c6296e905d9a4ff16079). So that shouldn’t break any pipelines that only permit commits from specific entities.
> - Will the commit message be the same? [Yup](https://dependabot.com/docs/config-file/#commit_message), so this shouldn’t break any pipelines that are parsing the commit messages.

### DIFF
--- a/source/standards/tracking-dependencies.html.md.erb
+++ b/source/standards/tracking-dependencies.html.md.erb
@@ -20,13 +20,18 @@ Update your dependencies frequently rather than in ‘big bang’ batches. This 
 
 There are tools which scan GitHub repositories and raise PRs when they find dependency updates. Teams at GDS are using:
 
-* [Dependabot][] - used by GOV.UK, GOV.UK Pay and GovWifi. The GOV.UK team manual contains [guidance on using Dependabot][] and [how the PRs raised should be reviewed][]
+* [Dependabot][] - used by GOV.UK, GOV.UK Pay, GovWifi and Digital Marketplace. The GOV.UK docs contain [guidance on using Dependabot][] and [how the PRs raised should be reviewed][]
+    > Note: this is separate from the [security-only updates provided automatically by GitHub Dependabot].
+
+    > Note: repos requiring at least one approving review for PRs cannot, and should not, use [Dependabot's auto-approve-and-merge facility].
+
+    > Note: we have not enabled "Treat PR approval as a request to merge", as this would lead to a surprising behaviour at the point of approval.
+
 * [PyUp][] - a Python dependency checker. Used by Digital Marketplace and GOV.UK Notify, PyUp will monitor for updates and vulnerabilities
 * [Greenkeeper][] - an npm dependency checker used by the GOV.UK Verify team on the [Node.js client for the Verify Service Provider][]
 
 All the above tools are free to use on public repositories.
 
-GitHub has turned on Dependabot for all repositories which are active, public and have not been forked.
 
 The Cyber Security team will review the repositories that do not have dependency management in use and will turn on Dependabot where required. Service teams are free to use a different tool such as [Snyk](https://snyk.io/), but will need to add a `no-dependabot` tag to their repository for monitoring purposes. You can [contact Cyber Security](https://gds.slack.com/archives/CCMPJKFDK) if you have any questions or need help.
 
@@ -70,6 +75,8 @@ Also consider managed solutions where possible. For example:
 
 This guidance is in line with the GDS Reliability Engineering strategic principle of [use fully managed cloud services by default][].
 
+[Dependabot's auto-approve-and-merge facility]: https://dependabot.com/blog/automatic-pull-request-merging/
+[security-only updates provided automatically by GitHub Dependabot]: https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies#alerts-and-automated-security-updates-for-vulnerable-dependencies
 [GDS supported programming languages]: /standards/programming-languages.html#content
 [managing software dependencies in the Service Manual]: https://www.gov.uk/service-manual/technology/managing-software-dependencies
 [programming language style guides]: /manuals/programming-languages.html


benthorner:
Comment from @36degrees 

> On the Design System team, we’ve restricted it so that only members of the team can push to the master branch. This is so that we can grant write access to all frontend devs at GDS (so that their PR approvals ’count') but two FE devs couldn’t ship a change without our team being involved.

> I’d be keen to check that Dependabot doesn’t trip up on this, as in theory it shouldn’t be able to merge to master, but it may try anyway!

> Other than that, sounds sane to me.

My understanding is that this setting _should_ be compatible with that way of working: Dependabot has **write** access to the repo, but not the required **admin** or **maintainer** that you've got protecting the master branch. Note: most of our repos don't have this protection.

source/standards/tracking-dependencies.html.md.erb
```diff
@@ -20,13 +20,18 @@ Update your dependencies frequently rather than in ‘big bang’ batches. This
 
 There are tools which scan GitHub repositories and raise PRs when they find dependency updates. Teams at GDS are using:
 
-* [Dependabot][] - used by GOV.UK, GOV.UK Pay and GovWifi. The GOV.UK team manual contains [guidance on using Dependabot][] and [how the PRs raised should be reviewed][]
+* [Dependabot][] - used by GOV.UK, GOV.UK Pay, GovWifi and Digital Marketplace. The GOV.UK docs contain [guidance on using Dependabot][] and [how the PRs raised should be reviewed][]
```
richardTowers, line 23: Could add GOV.UK PaaS to this too - dependabot's used for the nodeJS frontend.

heathd:
it would be good to know specifically what configuration settings you propose to use. 

also if you can link to the official documentation of what each option does

Screenshot 2020-05-18 at 10 31 38

the "enable auto-merging to be enabled on projects" sounds as though it can be enabled on a per-project basis?

benthorner:
@heathd this PR makes a few tweaks to the notes to clarify the guidance we already have. The only actual change proposed is to enable: "Treat PR approval as a request to PR".

Although we've checked "Allow auto-merging to be enabled on projects" for individual repos, this conflicts with the guidance about requiring at least one approving review.

I've been unable to find any documentation about this feature, apart from informal discussion that it's equivalent to `@dependabot merge`, and that it [can only be enabled org-wide](https://github.com/dependabot/feedback/issues/190#issuecomment-535491375).

source/standards/tracking-dependencies.html.md.erb
```diff
@@ -20,13 +20,18 @@ Update your dependencies frequently rather than in ‘big bang’ batches. This
...
+    > Note: this is separate from the [security-only updates provided automatically by GitHub Dependabot].
+
+    > Note: repos requiring at least one approving review for PRs cannot, and should not, use [Dependabot's auto-approve-and-merge facility].
+
+    > Note: our account is setup to "Treat PR approval as a request to merge", which helps to reduce some of the day-to-day, manual overhead.
```
richardTowers, line 28: I've been thinking about this, and I think I'm against it because:

* It provides very limited productivity benefits
* It adds surprising behaviour to GitHub's user interface
* I can imagine a scenario where this confusing behaviour could cause someone who didn't understand the setup to deploy something they didn't mean to

## It provides very limited productivity benefits

As far as I can tell, all this setting does for us is save you from having to click merge after you click approve. We could just train developers to click a couple more buttons, and get all the same benefits without any of the weirdness.

## It adds surprising behaviour to GitHub's user interface

"Approving" a PR in most cases signals that this user is happy for the PR to be merged (perhaps by another user), but having this automated is still surprising. New developers who've worked with Dependabot before will probably not expect this behaviour. People who are used to approving non-dependabot PRs (which won't auto-merge) will have to learn that dependabot PRs behave slightly differently.

If we learn anything from aeroplane crash investigations, it's that surprising user interfaces can cause unexpected bad things to happen.

## I can imagine a scenario where this confusing behaviour could cause someone who didn't understand the setup to deploy something they didn't mean to

Scenario:

> This version bump is fine to go out in principle, but it shouldn't be released until after this [other condition is met], or [something terrible] will happen. I'll approve the PR with a comment saying that it should only be merged once [other condition is met].

Consequence:

> [something terrible] will happen

It's easy to say that this hypothetical user is not using the "approvals" feature properly, but in a blameless incident retro for this hypothetical incident I would point at the approvals user interface as a cause.

richardTowers, line 28: I've been thinking about this, and I think I'm against it because:

* It provides very limited productivity benefits
* It adds surprising behaviour to GitHub's user interface
* I can imagine a scenario where this confusing behaviour could cause someone who didn't understand the setup to deploy something they didn't mean to

## It provides very limited productivity benefits

As far as I can tell, all this setting does for us is save you from having to click merge after you click approve. We could just train developers to click a couple more buttons, and get all the same benefits without any of the weirdness.

## It adds surprising behaviour to GitHub's user interface

"Approving" a PR in most cases signals that this user is happy for the PR to be merged (perhaps by another user), but having this automated is still surprising. New developers who've worked with Dependabot before will probably not expect this behaviour. People who are used to approving non-dependabot PRs (which won't auto-merge) will have to learn that dependabot PRs behave slightly differently.

If we learn anything from aeroplane crash investigations, it's that surprising user interfaces can cause unexpected bad things to happen.

## I can imagine a scenario where this confusing behaviour could cause someone who didn't understand the setup to deploy something they didn't mean to

Scenario:

> This version bump is fine to go out in principle, but it shouldn't be released until after this [other condition is met], or [something terrible] will happen. I'll approve the PR with a comment saying that it should only be merged once [other condition is met].

Consequence:

> [something terrible] will happen

It's easy to say that this hypothetical user is not using the approvals feature properly, but in a blameless incident retro for this hypothetical incident I would point at the approvals user interface as a cause.

heathd:
ah ok, so the "Allow auto-merging to be enabled on projects" is a different behaviour to the "Treat PR approval as a request to PR" function? 

If I understand correctly:

1. **Allow auto-merging to be enabled on projects** would allow a project to enable auto-merging. Auto-merging is where a PR gets merged even if it doesn't have approval from a reviewer?
2. **Treat PR approval as a request to PR** means that once a PR has been approved it gets merged

did I understand correctly?

source/standards/tracking-dependencies.html.md.erb
benthorner: Thanks for your comments @richardTowers - so good, you posted them twice 😆. While I do think we should avoid training already skilled developers as "manual mouse clickers" where possible, I also agree the productivity benefits of this are minimal, since manual intervention is already required to approve these PRs in the first place.

We have issues with our use of Dependabot - it's a large and not noticeably valuable pull on our time - but this would not address those. The other pull on our time has been repeating this discussion and never getting to this point, where we have a documentable conclusion. How about this for a (much less significant) change:

- I update this PR to document that we should not use this setting, because it's surprising.
- I seek a much quicker review from yourself and @heathd (because you've seen the PR).
- I put something in the next tech community weeknote to point people to the outcome.

heathd: just to say I think that Richard has articulated in a much clearer way than I was able to my nervousness about this change.

richardTowers: I'm pretty sure I posted it twice because of _surprising GitHub UI_, so if anything that goes to support my point.

👍  - I think that plan is fine. I'd be happy to 🚢  a small update to this page with a short explanation.

PR #442, opened 2020-06-26
feat(java-style-guide): Add guidance for when to use `var`

mrwilson:
From Java 10, you can use `var` instead of the full class name for local variables.

This can improve readability when there's unnecessary line noise caused by duplication between variable class and variable name.

### DIFF
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -18,6 +18,16 @@ We favour consistency across our code, so make sure that you have the agreement 
 
 Variable and field names should match the class they are instantiating, or have a descriptive name reflecting the context of their use.
 
+You can use `var` in Java 10 onwards with local variables to remove the duplication between class name and variable name.
+
+```java
+// Java 9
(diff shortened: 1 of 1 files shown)


alexbishop1:
I think we do have to apply some judgement to when to use `var`. It’s a no-brainer for most constructors:

```java
// Ugh
ByteArrayOutputStream output = new ByteArrayOutputStream();

// So much better
var output = new ByteArrayOutputStream();
```

And the same applies for a lot of getters, as exemplified by the examples in this PR.

However, there are some situations where `var` doesn’t add clarity. For example, imagine we have a method that does some complex operation and a returns a structure containing some results:

```java
// It’s clear what we’re getting back here
Map<String, List<String>> results = frobulator.frobulate();

// Not so clear what’s been returned here
var results = frobulator.frobulate();
```

There’s a couple of good articles with guidance that we might want to link to:

- https://developer.oracle.com/java/jdk-10-local-variable-type-inference.html
- https://openjdk.java.net/projects/amber/LVTIstyle.html

alexbishop1:
This got incorporated into https://github.com/alphagov/gds-way/pull/452.

PR #444, opened 2020-06-30
Deprecate Logit and add Splunk info

OllieJC:
Draft change to deprecate Logit and add information about Splunk, the now preferred tool for infrastructure, app and audit logging.

### DIFF
--- a/source/standards/logging.html.md.erb
+++ b/source/standards/logging.html.md.erb
@@ -109,7 +109,8 @@ requesting, ask in the `#cyber-security-help` Slack channel.
 ### Dropwizard
 
 The [dropwizard-logstash] library will set things up for you using the
-standard names. Splunk will be able to ingest this format.
+standard names. Splunk will be able to ingest this format but the ingestion
+route may differ as Splunk Cloud does not support syslog inputs.
 
 ### GOV.UK PaaS (Cloud Foundry)
 
@@ -135,6 +136,3 @@ drain logs into it from your app.
 [Reliability Engineering documentation]: https://reliability-engineering.cloudapps.digital/#logging
 [Payment Card Industry Data Security Standard (PCI-DSS)]: https://www.pcisecuritystandards.org/pci_security/
 [Amazon CloudWatch]: https://aws.amazon.com/cloudwatch/
-[Logstash HTTP output]: https://www.elastic.co/guide/en/logstash/current/plugins-outputs-http.html
-[HEC]: https://docs.splunk.com/Documentation/Splunk/latest/Data/HECExamples
-[Fluentd to HEC]: https://github.com/splunk/fluent-plugin-splunk-hec
--- a/source/standards/logging.html.md.erb
+++ b/source/standards/logging.html.md.erb
@@ -109,8 +109,7 @@ requesting, ask in the `#cyber-security-help` Slack channel.
 ### Dropwizard
 
 The [dropwizard-logstash] library will set things up for you using the
-standard names. Splunk will be able to ingest this format but the ingestion
-route may differ as Splunk Cloud does not support syslog inputs.
+standard names. Splunk will be able to ingest this format.
 
 ### GOV.UK PaaS (Cloud Foundry)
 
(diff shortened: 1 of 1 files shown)


source/standards/logging.html.md.erb
```diff
@@ -145,23 +175,30 @@ Example:
...
 ```
 
-### Advice for particular frameworks or platforms
+## Advice for particular frameworks or platforms
+
+### Dropwizard
 
-#### Dropwizard
+The [dropwizard-logstash] library will set things up for you using the
+standard names.
```
gidsg, line 183: Are we sure this advice is applicable to Splunk. As Logstash format is specific to Kibana (used by Logit)

OllieJC: Good point. I _think_ Splunk would handle the Logstash format fine, so isn't _incorrect_. There is some Splunk guidance for Java logging but I'm not sure I'd want to recommend proprietary libraries: https://dev.splunk.com/enterprise/docs/java/logging-java

source/standards/logging.html.md.erb
```diff
@@ -1,53 +1,84 @@
...
+
 ## Long-term storage
 
-When storing logs, you should focus on long-term durability for compliance reasons or to identify long-term performance trends.
+When storing logs, you should focus on long-term durability for compliance
+reasons or to identify long-term performance trends.
 
-ELK is not appropriate for long-term storage. There are other services which are less expensive and more reliable, for example if you use [Amazon Web Services (AWS)][] you could use [Amazon CloudWatch][] to run basic queries on older logs which may be a suitable solution.
+[Splunk] is appropriate for up to a year of storage, if you require more,
+consider archiving to [Amazon Web Services (AWS)] S3.
```
sivvo, line 49: can we simplify this by providing longer term archival within Splunk?

source/standards/logging.html.md.erb
```diff
@@ -1,172 +1,130 @@
...
+[helpdesk] to request access. If you're unsure what role you should be
+requesting, ask in the `#cyber-security-help` Slack channel.
+
+## Advice for particular frameworks or platforms
+
+### Dropwizard
+
+The [dropwizard-logstash] library will set things up for you using the
+standard names. Splunk will be able to ingest this format but the ingestion
+route may differ as Splunk Cloud does not support syslog inputs.
```
gidsg, line 104: I think this might be wrong as `dropwizard-logstash` is for formatting logs in a logstash format. Nothing to do with syslog. Which is a transport protocol not a log format.

OllieJC: @gidsg - I've amended this paragraph and moved the syslog note elsewhere.

source/standards/logging.html.md.erb
```diff
@@ -1,172 +1,141 @@
...
+## Advice for particular frameworks or platforms
+
+### Dropwizard
+
+The [dropwizard-logstash] library will set things up for you using the
+standard names. Splunk will be able to ingest this format.
+
+### Cloud Foundry
+
+There is [broker documentation] describing how drain logs to Splunk via
```
gidsg, line 116: This is a bit confusing I would rather mention PaaS in this section as Cloudfoundry is the abstraction tier the users shouldn't need to know about.  e.g. _drain logs from PaaS to Splunk_

source/standards/logging.html.md.erb
```diff
@@ -1,172 +1,141 @@
...
+files in a proprietary format so consider archiving to S3 in human-readable
+files using your own process.
 
 ## Log shipping
 
 If you have set up log shipping, you should consider:
 
-* how your logs will get to short and long-term stores - we recommend using the [Elastic Beats family of tools][] as a platform to collect and ship your data.
+* how your logs will get to short and long-term stores - we recommend using
+[Amazon CloudWatch] to collect logs and subscriping to the Cyber Security
```
lfdebrux, line 60:
```suggestion
[Amazon CloudWatch] to collect logs and subscribing to the Cyber Security
```

PR #445, opened 2020-07-01
Add async await to page on using NodeJS

tombye:
Adds a bit on using async functions to the section on asynchronous code.

They've been available for a while and have been widely adopted because they make asynchronous code easier to read and write.

### DIFF
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
 };
 ```
 
+#### Async functions
+
+If your version of Node is above 7.10.0 you should use async functions, and the await keyword, to organise your asynchronous code. This will make it easier to read and write.
+
+```
+// without async/await
+const getUserFromCreds = function (username, password) {
+  return getUserIdFromAuth(username, password)
+    .then(userId => {
+      return getUserFromApi(userId);
+    })
+    .then(user => {
+      return {
+        'id': user.id,
+        'name': user.name,
+        'age': user.ageInYears      
+      };
+    })
+    .catch(e => {
+      console.log(e);
+    });
+}
+
+// with async/await
+const getUserFromCreds = async function (username, password) {
+  try {
+    const userId = await getUserIdFromAuth(username, password);
+    const user = await getUserFromApi(userId);
+
+    return {
+      'id': user.id,
+      'name': user.name,
+      'age': user.ageInYears
+    };
+  } catch(e) {
+    console.log(e);
+  }
+}
+```
+
+The `await` keyword will pause execution until its promise fulfills. This means a sequence of awaited promises can only execute in the order they're declared.
+
+If you need your promises to execute non-sequentially, you can instantiate them separately from when you await them (see [the downsides of async/await on MDN](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await#The_downsides_of_asyncawait) for more on this pattern):
+
+```
+const getUserFromCreds = async function (username, password) {
+  try {
+    // use await to make this promise blocking
+    await userId = getUserIdFromAuth(username, password);
+
+    // start other promises running before awaiting
+    const avatarPromise = getAvatarForUser(userId);
+    const userPromise = getUserFromApi(userId);
+
+    const avatarImgUrl = await avatarPromise;
+    const user = await userPromise;
+
+    return {
+      'id': userId,
+      'name': user.name,
+      'age': user.ageInYears,
+      'avatar': avatarImgUrl
+    };
+  } catch(e) {
+    console.log(e);
+  }
+}
+```
+
 
 ### Functional programming
 
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -255,7 +255,7 @@ const callback = filename => (err, data) => {
 
 #### Use of `async`/`await`
 
-If your version of Node is above 7.10.0 you should use async functions, and the await keyword, to organise your asynchronous code. This will make it easier to read and write.
+If your version of Node is above 7.10.0 you should use async functions, and the await keyword, to organise your asynchronous code. This will make your code easier to read and write.
 
 ```js
 // without async/await
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -271,25 +271,18 @@ const getUserFromCreds = function (username, password) {
         'age': user.ageInYears      
       };
     })
-    .catch(e => {
-      console.log(e);
-    });
 }
 
 // with async/await
 const getUserFromCreds = async function (username, password) {
-  try {
-    const userId = await getUserIdFromAuth(username, password);
-    const user = await getUserFromApi(userId);
+  const userId = await getUserIdFromAuth(username, password);
+  const user = await getUserFromApi(userId);
 
-    return {
-      'id': user.id,
-      'name': user.name,
-      'age': user.ageInYears
-    };
-  } catch(e) {
-    console.log(e);
-  }
+  return {
+    'id': user.id,
+    'name': user.name,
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
 };
 ```
 
+#### Async functions
+
+If your version of Node is above 7.10.0 you should use async functions, and the await keyword, to organise your asynchronous code. This will make it easier to read and write.
+
+```
```
ChrisBAshton, line 260: This should be ```js to get free code highlighting.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
...
+    };
+  } catch(e) {
+    console.log(e);
+  }
+}
+```
+
+The `await` keyword will pause execution until its promise fulfills. This means a sequence of awaited promises can only execute in the order they're declared.
+
+If you need your promises to execute non-sequentially, you can instantiate them separately from when you await them (see [the downsides of async/await on MDN](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await#The_downsides_of_asyncawait) for more on this pattern):
```
ChrisBAshton, line 298: I think we should make it a suggestion rather than a 'you could do this'.

```suggestion
If you have multiple promises, you should try to [instantiate them separately from when you await them](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await#The_downsides_of_asyncawai), otherwise each promise will be forced to wait for the previous one. Example:
```

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
...
+
+```
```
ChrisBAshton, line 300: And this: ```js

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
 };
 ```
 
+#### Async functions
```
ChrisBAshton, line 256: I think this would be a more descriptive heading for people scanning the page:

```suggestion
#### Use of `async`/`await`
```

tombye: I like the new heading but it doesn't look like the font for code in headings adds much. What do you think about leaving at the backticks?

With backticks:
image

Without backticks:
image

DilwoarH: I think without backticks is probably bit better design wise

ChrisBAshton: Agreed 👍  thanks both.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
...
+
+The `await` keyword will pause execution until its promise fulfills. This means a sequence of awaited promises can only execute in the order they're declared.
+
+If you have multiple promises, you should try to [instantiate them separately from when you await them](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await#The_downsides_of_asyncawait), otherwise each promise will be forced to wait for the previous one. Example:
+
+```js
+const getUserFromCreds = async function (username, password) {
+  try {
+    // use await to make this promise blocking
+    await userId = getUserIdFromAuth(username, password);
```
sfount, line 304: I don't think this would be valid syntax
```suggestion
    const userId = await getUserIdFromAuth(username, password);
```

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
...
+
+    // start other promises running before awaiting
+    const avatarPromise = getAvatarForUser(userId);
+    const userPromise = getUserFromApi(userId);
+
+    const avatarImgUrl = await avatarPromise;
```
sfount, line 310: Another option if you're waiting on multiple promises and want to make that clear is wrapping them in `Promise.all`. I think what you've got here is a clear example so not necessarily suggesting replacing it, just a useful pattern I've found in the past.

```js
const avatarRequest = getAvatarForUser(userId)
const userRequest = getUserFromApi(userId)

const [ avatarImgUrl, user ] = await Promise.all([ avatarRequest, userRequest ])
// return { ... }
```

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
 };
 ```
 
+#### Use of `async`/`await`
+
+If your version of Node is above 7.10.0 you should use async functions, and the await keyword, to organise your asynchronous code. This will make it easier to read and write.
```
sfount, line 258:
```suggestion
If your version of Node is above 7.10.0 you should use async functions, and the await keyword, to organise your asynchronous code. This will make your code easier to read and write.
```

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -253,6 +253,75 @@ const callback = filename => (err, data) => {
...
+      return getUserFromApi(userId);
+    })
+    .then(user => {
+      return {
+        'id': user.id,
+        'name': user.name,
+        'age': user.ageInYears      
+      };
+    })
+    .catch(e => {
```
chrisfarms, line 274: I don't think having the `.catch` and `try/catch` logic in these examples is required to demonstrate the main differences here

having a try/catch with an console.log would probably be considered bad practice within a async function (should always let the error propagate up to the caller) so the code ends up looking a artificial in the second example.

If you want talk about how error handling changes, maybe it deserves it's own example

source/manuals/programming-languages/nodejs/index.html.md.erb
tombye: Good spot, I corrected this in my test files but not here 👍

source/manuals/programming-languages/nodejs/index.html.md.erb
tombye: Makes sense. I added it to ape the examples in https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await but I see your point. Now removed.

Would you be interested in raising another PR to add a section on error handling in asynchronous JS?

source/manuals/programming-languages/nodejs/index.html.md.erb
tombye: It makes a lot of sense to use what's available from the promises API first. In your example, do the promises returned by `getAvatarForUser` and `getUserFromApi` need to be stored in variables are could they be passed straight into `Promise.all`?

tombye: @sfount I've had a go at rewriting that section in https://github.com/alphagov/gds-way/pull/445/commits/af00770b77430343bf1b80b4c922785de18d7f4e, be interested in your opinion.

tombye:
Github seems to think the work still going on in response to https://github.com/alphagov/gds-way/pull/445#discussion_r448241546 means I dismissed @ChrisBAshton 's approval. Not true, just tidying up the last few comments 🙂

source/manuals/programming-languages/nodejs/index.html.md.erb
sfount: @tombye  looks great 👍 no need to store anything so passing the calls directly into the `await Promise.all` is totally valid. 

The link through to the static methods on Promises seems super helpful too.

tombye:
@ChrisBAshton @chrisfarms @DilwoarH @sfount 

I made @ChrisBAshton 's approval invalid by making a further change based on a comment from @sfount. I feel I've got all the bits of approval needed to merge but need someone to push the 👍 button, would one of you mind doing that?

PR #460, opened 2020-07-30
Add some terraform specific guidance.

deanwilson:
Once we have another team using it this should be updated to include
conftest.

### DIFF
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -20,6 +20,27 @@ Use [Terraform][] to configure third party cloud infrastructure like Amazon Web 
 
 Terraform supports a [large number of providers][], and you can configure it to support multiple environments with different parameters. See the [govuk-aws][] repository as an example.
 
+### Versioning
+
+Due to the high rate of change in many cloud provider offerings we
+recommend you keep your Terraform versions and codebases up to date. A
+version manager such as [tfenv](https://github.com/tfutils/tfenv),
+already used by a number of GDS teams, can help you with supporting
+multiple versions.
+
+### Code analysis
+
+There are a number of Terraform focused static analysis tools in use at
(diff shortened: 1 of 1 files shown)


source/standards/configuration-management.html.md.erb
```diff
@@ -20,6 +20,27 @@ Use [Terraform][] to configure third party cloud infrastructure like Amazon Web
 
 Terraform supports a [large number of providers][], and you can configure it to support multiple environments with different parameters. See the [govuk-aws][] repository as an example.
 
+### Versioning
+
+Due to the high rate of change in many cloud provider offerings we
+recommend you keep your terraform versions and codebases up to date. A
+version manager such as [tfenv](https://github.com/tfutils/tfenv),
+already used by a number of GDS teams, can help you with supporting
+multiple versions.
```
issyl0, line 29: Is it worth mentioning `.terraform-version` files here too?

source/standards/configuration-management.html.md.erb
```diff
@@ -20,6 +20,27 @@ Use [Terraform][] to configure third party cloud infrastructure like Amazon Web
 
 Terraform supports a [large number of providers][], and you can configure it to support multiple environments with different parameters. See the [govuk-aws][] repository as an example.
 
+### Versioning
+
+Due to the high rate of change in many cloud provider offerings we
+recommend you keep your terraform versions and codebases up to date. A
```
issyl0, line 26:
```suggestion
recommend you keep your Terraform versions and codebases up to date. A
```

source/standards/configuration-management.html.md.erb
deanwilson: That's an implementation detail. Once they look at the tool they'll see it

source/standards/configuration-management.html.md.erb
deanwilson: Fix incoming.

PR #478, opened 2020-08-27
Bumps review date on naming software page

louzoid-gds:
Naming things is hard but this guidance is still in good shape :)


OllieJC:
Nit: https://github.com/alphagov/gds-way/blob/5d1ada91c6df379f86da5975d9c5324ed7b02150/source/standards/naming-software-products.html.md.erb#L29 is now `search-api` is it worth calling that out as a good thing?

louzoid-gds:
Good spot @OllieJC. I've deleted that example

arnau (approved):

PR #487, opened 2020-09-10
Update slis.html.md.erb

emilygds:
Added text to explain how the team decided where to scrape metrics for their SLIs. This should hopefully make the last remaining diagram accessible.

Also restructured slightly so that the image of a formula has explanatory text above.

This doc could do with a general style review, but from an accessibility perspective I think the diagrams are now covered.


cmcnallygds (commented):
Looks good to me!

cmcnallygds (approved):
LGTM

PR #495, opened 2020-09-29
Document tools we use to test accessibility, browser support

36degrees:
Tell people about the tools that we have access to that we can use to test services with assistive technologies, and with different browsers. I've checked with Matt that he's happy for people to be pointed in his direction to get accounts…

I've split the browser testing section out from the 'building accessible services' guidance. This results in a very short page, but it doesn't feel right to have it buried in the accessibility guidance when it's a thing in its own right.

I haven't bumped the date on the accessibility page because I haven't reviewed the rest of the content.

We don't actually have access to Assistiv Labs (it's a work in progress…) so this should remain a draft until we do. We might also need to document how to activate the JAWS license.

### DIFF
--- a/source/manuals/accessibility.html.md.erb
+++ b/source/manuals/accessibility.html.md.erb
@@ -44,6 +44,8 @@ You can use [Assistiv Labs](https://assistivlabs.com/) to test your service with
 
 Read more about [using JAWS](https://webaim.org/articles/jaws/), [using NVDA](https://webaim.org/articles/nvda/) and [using VoiceOver](https://webaim.org/articles/voiceover/).
 
+You can run Voiceover training on macOS by going to System Preferences > Accessibility > VoiceOver > Open VoiceOver Training…
+
 ### Test content served by third party systems
 
 If a service is built using an existing or third party system, such as a content management system or JavaScript framework, it should be tested for accessibility. Accessibility compliance cannot be guaranteed by any system, as even small changes can introduce accessibility barriers.
--- a/source/manuals/accessibility.html.md.erb
+++ b/source/manuals/accessibility.html.md.erb
@@ -36,11 +36,19 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
 
 Accessibility testing tools such as [aXe](https://www.deque.com/axe/) and [Lighthouse](https://developers.google.com/web/tools/lighthouse/) are useful for finding issues, but [automated testing can only find around 30% of likely accessibility problems](https://alphagov.github.io/accessibility-tool-audit/). Automated testing shouldn’t be the only accessibility testing carried out on a service, it should also be tested with real users.
 
+### Testing with assistive technologies
+
+Test your service with assistive technologies throughout development, especially when you introduce a significant feature or make any major change. You should test in [the assistive technology and browser combinations listed in the service manual](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies).
+
+You can use [Assistiv Labs](https://assistivlabs.com/) to test your service with JAWS and NVDA. Contact Matt Hobbs (Head of Frontend) to request an account.
+
+Read more about [using JAWS](https://webaim.org/articles/jaws/), [using NVDA](https://webaim.org/articles/nvda/) and [using VoiceOver](https://webaim.org/articles/voiceover/).
+
 ### Browser compatibility testing
 
 Services should be thoroughly tested in all browsers they are required to be compatible with, including mobile devices and accessibility tools such as screen readers. Emulated browser testing services are available, but some testing on real mobile devices is also recommended.
 
-For reference, these are the [browsers that GOV.UK is required to be compatible with](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) and a service should also work with these [assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#which-assistive-technologies-to-test-with).
+For reference, these are the [browsers that GOV.UK is required to be compatible with](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
 
 ### Test content served by third party systems
 
--- a/source/manuals/accessibility.html.md.erb
+++ b/source/manuals/accessibility.html.md.erb
@@ -44,12 +44,6 @@ You can use [Assistiv Labs](https://assistivlabs.com/) to test your service with
 
 Read more about [using JAWS](https://webaim.org/articles/jaws/), [using NVDA](https://webaim.org/articles/nvda/) and [using VoiceOver](https://webaim.org/articles/voiceover/).
 
-### Browser compatibility testing
-
-Services should be thoroughly tested in all browsers they are required to be compatible with, including mobile devices and accessibility tools such as screen readers. Emulated browser testing services are available, but some testing on real mobile devices is also recommended.
-
-For reference, these are the [browsers that GOV.UK is required to be compatible with](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
-
 ### Test content served by third party systems
 
 If a service is built using an existing or third party system, such as a content management system or JavaScript framework, it should be tested for accessibility. Accessibility compliance cannot be guaranteed by any system, as even small changes can introduce accessibility barriers.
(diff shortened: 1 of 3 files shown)


Nooshu (dismissed):
Looks good to me, thanks Ollie.

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,13 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
...
 
-### Browser compatibility testing
+### Testing with assistive technologies
 
-Services should be thoroughly tested in all browsers they are required to be compatible with, including mobile devices and accessibility tools such as screen readers. Emulated browser testing services are available, but some testing on real mobile devices is also recommended.
+Test your service with assistive technologies throughout development, especially when you introduce a significant feature or make any major change. You should test in [the assistive technology and browser combinations listed in the service manual](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies).
 
-For reference, these are the [browsers that GOV.UK is required to be compatible with](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) and a service should also work with these [assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#what-to-test).
+You can use [Assistiv Labs](https://assistivlabs.com/) to test your service with JAWS and NVDA. Contact Matt Hobbs (Head of Frontend) to request an account.
```
chrisfarms, line 43: Could we expand the acronyms on first use. I got a picture of a shark when I googled for JAWS :wink:

36degrees: Happy to, but not sure that it helps the user much… we don't expand them in the service manual either?

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,29 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
...
+### Testing with assistive technologies
 
-Services should be thoroughly tested in all browsers they are required to be compatible with, including mobile devices and accessibility tools such as screen readers. Emulated browser testing services are available, but some testing on real mobile devices is also recommended.
+Test your service with assistive technologies throughout development, especially when you introduce a significant feature or make any major change. You should test in [the assistive technology and browser combinations listed in the service manual](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies).
 
-For reference, these are the [browsers that GOV.UK is required to be compatible with](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) and a service should also work with these [assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#what-to-test).
+Read more about [using JAWS](https://webaim.org/articles/jaws/), [using NVDA](https://webaim.org/articles/nvda/) and [using VoiceOver](https://webaim.org/articles/voiceover/).
```
cmcnallygds, line 43: We should write out the acronyms the first time we use them

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,29 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
...
+
+You can run Voiceover training on macOS by going to System Preferences > Accessibility > VoiceOver > Open VoiceOver Training…
```
cmcnallygds, line 45: Should be 'VoiceOver'

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,29 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
...
+
+#### Using Assistiv Labs to test with assistive technologies
+
+You can use [Assistiv Labs](https://assistivlabs.com/) to test your service with JAWS and NVDA.
+
+Contact Matt Hobbs (Head of Frontend) to request an account.
```
cmcnallygds, line 51: We usually don't name people because the content becomes out-of-date when they move on. Could we just say 'Head of Frontend'?

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,29 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
...
+
+We have been assured by Assistiv Labs that the Virtual Machine (VM) is deleted when you exit, however in the interests of security and ensuring we do not leak any personal data or sensitive material:
```
cmcnallygds, line 53: We should use 'make sure' rather than 'ensure' according to the style guide

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,29 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
...
+
+- treat the VM you are using as if it was a shared device – like using a public computer in an internet cafe or library
+- do not use it to test any confidential or secret service, or services containing content that is considered sensitive
+- do not use any credentials from live systems with access to "real data" during your testing
+- do not sign in to any accounts, unless those accounts are used solely for the purposes of testing (for example, a test account in an integration or staging environment)
+- make configuration changes to the system through your own device rather than the VM
+
+You should also make sure you understand how using the service affects any other requirements for your service, such as PCI compliance, and assess any additional risks. You can speak to your service's information assurance lead or the GDS privacy team for specific guidance.
```
cmcnallygds, line 61: We should probably expand the PCI acronym

source/manuals/accessibility.html.md.erb
```diff
@@ -36,11 +36,29 @@ This is why [semantic markup](https://html.com/semantic-markup/) and good headin
 
 Accessibility testing tools such as [aXe](https://www.deque.com/axe/) and [Lighthouse](https://developers.google.com/web/tools/lighthouse/) are useful for finding issues, but [automated testing can only find around 30% of likely accessibility problems](https://alphagov.github.io/accessibility-tool-audit/). Automated testing shouldn’t be the only accessibility testing carried out on a service, it should also be tested with real users.
 
-### Browser compatibility testing
+### Testing with assistive technologies
 
-Services should be thoroughly tested in all browsers they are required to be compatible with, including mobile devices and accessibility tools such as screen readers. Emulated browser testing services are available, but some testing on real mobile devices is also recommended.
+Test your service with assistive technologies throughout development, especially when you introduce a significant feature or make any major change. You should test in [the assistive technology and browser combinations listed in the service manual](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies).
```
cmcnallygds, line 41: Service Manual should be capitalised

source/manuals/browser-support.html.md.erb
```diff
@@ -0,0 +1,13 @@
+---
+title: Supporting different browsers
+last_reviewed_on: 2020-09-29
+review_in: 3 months
+---
+
+# <%= current_page.data.title %>
+
+Test your service in [the browsers listed in the service manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
```
cmcnallygds, line 9: As above, Service Manual should be capitalised

source/manuals/browser-support.html.md.erb
```diff
@@ -0,0 +1,13 @@
...
+
+You can use [BrowserStack](https://www.browserstack.com/) to test browsers that run on Windows, mobile phones and tablets. Contact Matt Hobbs (Head of Frontend) to request an account. You should also test with physical devices whenever possible.
```
cmcnallygds, line 11: As above, we should try to avoid naming individuals

source/manuals/accessibility.html.md.erb
36degrees: JAWS and NVDA are acronyms, but are generally never referred to in their expanded form. For example, [NVDA's own product page](https://www.nvaccess.org/about-nvda/) only ever uses the acronym.

Do we still think it's useful to expand them?

36degrees:
@cmcnallygds Updated with your suggested changes

source/manuals/accessibility.html.md.erb
36degrees: Talked about this elsewhere and realised the main issue was that when JAWS and NVDA are 'introduced' there's no context, so split them out into a list of resources with a lead in that makes it clear they're screen readers (b18016a )

36degrees:
We've now got access to AssistivLabs and have started rolling it out, so this is ready to be reviewed now.

cmcnallygds (approved):

PR #498, opened 2020-09-29
Review code review guidelines page

lfdebrux:
I find this page useful, the content looks to me to be still good.

There is a dead link at the bottom (to the GDS Tech Learning Pathway), but I understand this to be a temporary issue that will be resolved one way or another later down the line, so I've left it in.


deanwilson (approved):
Thank you for the review. Merge away

PR #500, opened 2020-09-30
Update alt text guidance

injms:
## What

The image alt text and caption guidance now better aligns with the updated [content guidance](https://www.gov.uk/guidance/content-design/images).

## Why

The content guidance the the developer guidance were contradictory - the content guidance was updated recently, so this needed to also be updated.


deanwilson (approved):
Looks good and I've learned something.

Merge away

PR #502, opened 2020-09-30
Minor amends to the hosting page

louzoid-gds:
And also a little bit of **strong** action to further highlight that this guidance applies to GDS only


tomnatt (approved):
LGTM :shipit:

PR #506, opened 2020-10-01
Bump CSS coding style with v minor tweaks

injms:
Bumped the date for review, plus two very minor changes:

 - Change Sass comment (`//`) to CSS comment (`/* ... */`) in CSS example
 - Removed reference to the deprecated frontend toolkit
 - Updated heading about using `rem` units with deprecated tools
 - Shortened review time, since we'll need to review this sooner than 6 months

### DIFF
--- a/source/manuals/programming-languages/css.html.md.erb
+++ b/source/manuals/programming-languages/css.html.md.erb
@@ -1,7 +1,7 @@
 ---
 title: CSS coding style
-last_reviewed_on: 2020-02-05
-review_in: 6 months
+last_reviewed_on: 2020-10-01
+review_in: 3 months
 ---
 
 # <%= current_page.data.title %>
@@ -23,10 +23,7 @@ You should use the [GOV.UK Design System spacing scale](https://design-system.se
 
 ### Use with deprecated libraries
 
-If your project is not using the GOV.UK Design System, you should use the
-[variables from the frontend_toolkit][measurements.scss] or use pixel values in multiples of `5px`.
-
-[measurements.scss]: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_measurements.scss
+If your project is not using the GOV.UK Design System, you should use pixel values in multiples of `5px`.
 
 ### Include fallbacks for rem spacing
 
@@ -40,10 +37,10 @@ If you do need to use `rem` units, include a fallback for browsers that do not s
 
 ```css
 font-size: 12px;
-font-size: 1.2rem; // This line is ignored by older browsers
+font-size: 1.2rem; /* This line is ignored by older browsers */
 ```
 
-#### Use with deprecated libraries
+#### Using `rem` units with deprecated libraries
 
 You should avoid using `rem` units with [GOV.UK Template](https://github.com/alphagov/govuk_template).
 


source/manuals/programming-languages/css.html.md.erb
```diff
@@ -23,8 +23,7 @@ You should use the [GOV.UK Design System spacing scale](https://design-system.se
 
 ### Use with deprecated libraries
 
-If your project is not using the GOV.UK Design System, you should use the
-[variables from the frontend_toolkit][measurements.scss] or use pixel values in multiples of `5px`.
+If your project is not using the GOV.UK Design System, you should use the [variables from the frontend_toolkit][measurements.scss] or use pixel values in multiples of `5px`.
```
andysellick, line 26: Should we be talking about using toolkit at all? At the very least we should be clearly recommending using govuk-frontend instead.

andysellick (approved):
Only one comment, non-blocking. I've realised to address it would require rewriting quite a bit of this file, maybe there could be a general warning at the top for it instead?

source/manuals/programming-languages/css.html.md.erb
injms: Or it could be
```suggestion
If your project is not using the GOV.UK Design System you should use pixel values in multiples of `5px`.
```

andysellick: Yeah, that seems good 👍

injms: Done!

source/manuals/programming-languages/css.html.md.erb
```diff
@@ -23,8 +23,7 @@ You should use the [GOV.UK Design System spacing scale](https://design-system.se
 
 ### Use with deprecated libraries
 
-If your project is not using the GOV.UK Design System, you should use the
-[variables from the frontend_toolkit][measurements.scss] or use pixel values in multiples of `5px`.
+If your project is not using the GOV.UK Design System, you should use pixel values in multiples of `5px`.
 
 [measurements.scss]: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_measurements.scss
```
andysellick, line 28: If we're not talking about toolkit anymore should this be removed?

andysellick (commented):
Sorry, I've been a bit more thorough this time and actually read the whole file. I think we need to add something in the `Use with deprecated libraries` section around line 45 about not using deprecated libraries and using the Design system instead. Or remove that section - it feels very out of date.

injms:
> Sorry, I've been a bit more thorough this time and actually read the whole file. I think we need to add something in the `Use with deprecated libraries` section around line 45 about not using deprecated libraries and using the Design system instead. Or remove that section - it feels very out of date.

I took that section to be specifically talking about not using `rem` units with govuk_template, not spacing in general - I've updated the heading to make that clearer. 

I think that it's still just about relevant, so I've shortened the review date to be three months so that we can review it sooner and remove the section if things change.

I'd rather not add in another repetitive reference to needing to use the Design System, since it's already mentioned at the start of the section on line 22 and I think that covers it.

andysellick (approved):

PR #508, opened 2020-10-01
Update HTML coding style page

injms:
* Removed section about fallbacks for unsupported elements because that content sits in the service manual section on progressive enhancement

 - Removed section on using a JavaScript HTML5 shiv because (to the best of my knowledge) that's not something that appears to be done at GDS

 - Added link to MDN article on ARIA roles

 - Updated dead link to Léonie Watson's article about ARIA roles - the old article has been removed, but Léonie has published another article about how to use HTML5 and ARIA roles together which is relevant and useful

 - Changes two `h1`s to two `h2` in example code, as elsewhere in the article it says not to have multiple `h1`s on the same page 🤦‍♂️

 - Tried to make it clearer that browsers render `em` in italics and `strong` in bold by default

 - Mentioned visually hidden classes instead of only mentioning the Sass mixin

 - Ran through the alex checker - replaced a 'just' with 'only'

 - A couple of minor indentation and spacing changes

### DIFF
--- a/source/manuals/programming-languages/html.html.md.erb
+++ b/source/manuals/programming-languages/html.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: HTML coding style
-last_reviewed_on: 2020-03-06
+last_reviewed_on: 2020-10-01
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
All the changes look good to me 👍 Thanks Ian.

PR #509, opened 2020-10-02
Date bump - this document is still valid

deanwilson:


stephenharker (approved):
Ok for now, valid for procurement.

PR #511, opened 2020-10-13
Renamed master branch to main branch

jeffmaher:


paroxp (approved):
This may be confusing to teams which haven't switched yet. But we can handle that as we know the direction.

PR #514, opened 2020-10-19
Change master to main in docs

lfdebrux:
We want to encourage GDSers to have their main branch named 'main'
instead of 'master', so we should use that language when talking about
branches.


deanwilson (approved):
Looks good and helps us get to where we want to be, thank you.

Merge when green.

PR #516, opened 2020-10-19
BAU: Fix broken links

mrwilson:
This PR is a collection of commits that fix outbound links that no longer exist _or_ no longer accurately represent the link name.

## Implementation
```
# Find most non-relative links in erb templates and attempt to resolve them
for url in $(
  find . -name '*.erb' -exec grep -e "^\[.*\]: http.*$" {} \; | sed -e 's/^\[.*\]: \(.*\)/\1/g'
); do
   curl -s -o /dev/null -w "%{http_code} - ${url}\n" $url;
done >> urls.txt

# Ignore all urls which resolved with 200 OK
fgrep -v "200 -" urls.txt
```


deanwilson (approved):
Exceptionally diligent work! Merge away.

PR #519, opened 2020-10-23
Bump date. No content change

deanwilson:
This has been unreviewed for a very long time due to no one being sure of the
exact nature of the phrase:

"Reliability Engineering is running a beta on using Prometheus as the operational metrics service for GDS."

This can be discussed by the newly formed CDIO CnD pillar as they will be the new owners of this. So in the meantime the rest of the doc is as valid as ever.

I've not renamed the owners as we're waiting on official branding and it'll be easier to do a mass RE find and replace.


louzoid-gds (approved):
LGTM

PR #523, opened 2020-11-25
Document standardx as a JavaScript linting approach

kevindew:
This change documents the usage of [standardx](https://github.com/standard/standardx) as an appropriate JavaScript linting tool when the usage of [standard](https://standardjs.com) conflicts with GDS approaches to browser compatibility.

It also fixes a couple of issues with the JS programming language doc and bumps the review date.

A much more wordy explanation follows:

GDS encountered a tricky problem with their standard choice when standard 16 was released, this release pushed a new rule, [no-var](https://github.com/standard/standard/issues/633), which issued warnings when projects used `var` over `let` or `const`. This caused problems for a number of GDS projects where `var` is still used as it required for JS to operate in IE < 11.

In order to resolve this problem the suggestion is to use standardx which operates the same as standard but allows individual project rules. This means that a project can continue using an up-to-date rule set for standard while disabling any rules that are problematic (this is expected to be `no-var` for most projects, but future problematic rules may
occur). This change sets about documenting this decision and to communicate the usage of standardx as a reasonable alternative to standard.

When considering how to approach this problem we experimented with the suggestion in this file of switching to a generic `.eslintrc` setup. Doing this is actually rather complicated. We found that if we wanted to share a rule set that built upon standard but with GDS customisations then we had a rather arduous installation and configuration process (for example the installation process went from `npm install --save-dev standard` to `npm install --save-dev eslint eslint-config-gds eslint-plugin-node eslint-plugin-import eslint-plugin-promise`) which seemed strange for a one rule change, then there were a bunch of curiosities: should we disable all rules that allow > ES5 functionality for consistency? why do browser projects need to install various, seemingly unnecessary, plugins? should we recommend one rule set for modern projects (standard) and another for legacy (eslint-config-gds)?

Given these thoughts we decided that standardx was a simple pragmatic choice that allowed us to move forward without being stuck on an old version. It's quite conceivable that this problem could increase if standard adopts further rules that have browser compatibility issues, which may force us to reconsider if we need to share a config. Hopefully, though, GDS will adopt the newer practices in JS and keep up with the recommendations standard tries to enforce.

### DIFF
--- a/source/manuals/programming-languages/js.html.md.erb
+++ b/source/manuals/programming-languages/js.html.md.erb
@@ -23,9 +23,11 @@ review_in: 3 months
 
 ## Linting
 
-We follow [standardjs](http://standardjs.com/), an opinionated JavaScript linter.
+We follow [standardjs](http://standardjs.com/), an opinionated JavaScript linter. In cases where [standard conflicts with a service's browser support](#when-to-use-standardx) we use [standardx][].
 
 All JavaScript files follow its conventions, and it typically runs on CI to ensure that new pull requests are in line with them.
+
+[standardx]: https://github.com/standard/standardx
 
 ## Running standard manually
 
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
This looks good to me.

So long as nobody tries to use semicolons, I'm happy :trollface:.

source/manuals/programming-languages/js.html.md.erb
```diff
@@ -23,23 +23,37 @@ review_in: 3 months
 
 ## Linting
 
-We follow [standardjs](http://standardjs.com/), an opinionated JavaScript linter.
+We follow [standardjs](http://standardjs.com/), an opinionated JavaScript linter. In cases where standard conflicts with a services browser support we use [standardx][].
```
injms, line 26: Could this part include a link to part of this page which explains why standardx is being used? I was thinking about the "Using standardx" section, as this bit could be a bit clearer about which rules should be overridden.

kevindew: Thanks for the suggestion Ian. I've reworked the "Using standardx" section to be more of a "When to use standardx" to try increase this clarity. How does that look to you?

injms (approved):
Super - I think this makes it clear why and when standardx should be used 👍

huwd (approved):
Looks good, clear explaination

PR #527, opened 2020-12-15
Change target dates for accessibility fixes

cmcnallygds:
Updated the date for fixing the Tech Docs Template accessibility issues from “end of December 2020” to “end of March 2021”. As a result of resourcing pressures and re-prioritisation this year, it’s taking longer than we hoped to arrange developer and designer resources to fix the accessibility issues.


venusbb:
Looks good.

venusbb (approved):
Changes look OK.

PR #532, opened 2021-01-05
Remove the TechOps page and add a new one for the parts that moved to CDIO

deanwilson:
Most of the text has been moved from the old techops page to the new CDIO one. I've also tried to rename `TechOps Cyber` to `CDIO Security` to reflect the pillar names.

This is only an initial draft and can be updated and added to in the future as the other pillars participate in the GDSWay.

PR #540, opened 2021-01-21
Remove the TechOps page and add a new one for the parts that moved to…

deanwilson:
… CDIO

### DIFF
--- a/config/tech-docs.yml
+++ b/config/tech-docs.yml
@@ -38,6 +38,6 @@ owner_slack_workspace: gds
 redirects:
   /standards/publish-opensource-code.html: /standards/source-code.html#publish-open-source-code
   /standards/git.html: /standards/source-code.html#working-with-git
-  /standards/cyber-security-overview.html: /standards/technical-operations.html#cyber-security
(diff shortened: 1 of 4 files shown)


louzoid-gds:
It looks like the Technical Operations page will remain (https://gds-way.cloudapps.digital/standards/technical-operations.html)
Is that intended? I'm not sure we need both and deleting it might be the way to go.
Everything else is looking good.

timblair:
Is this a duplicate of #532?

tomnatt:
I've updated a few minor style points and rebased (note the force push). I believe this is ready to go - can someone who hasn't committed code in this PR do a formal review please.

timblair (changes requested):
Could you use a [redirect](https://github.com/alphagov/gds-way/blob/9ec1034f/config/tech-docs.yml#L38) instead of leaving the "look here instead" TechOps page?

timblair (approved):
But let's not block the merge. That redirect can be added as a next step.

deanwilson:
I'll change the placeholder page to a redirect on the next review date. TechOps is still in use in a lot of places.

PR #543, opened 2021-01-22
Bumps review date on quarterly planning page

louzoid-gds:
Some programmes have moved away from quarterly planning but a large part of GDS and CDIO C&D still do hence guidance still valid.
Also sets review date to 3 months so we can review sooner than 6.


deanwilson (approved):
Merge when green

PR #546, opened 2021-01-28
Review Go style guide

LeePorte:
There is no need to change this guide at this point in time. Review
again in 6 months.


louzoid-gds (approved):
LGTM

PR #549, opened 2021-01-28
Remove inaccessible image and minor style change

cmcnallygds:
Removes the image as it's in an inaccessible format and unnecessary, as its contents are explained in the text.


louzoid-gds (approved):
👍

PR #556, opened 2021-02-24
Commit deployment pipeline to this repository instead

Krenair:
This used to be part of tech-ops.git reliability-engineering/pipelines/internal-apps.yml - which was the `internal-apps` pipeline in the autom8 team. This was weird and none of our other pipelines really group mostly unrelated apps together like that. Ideally pipelines should live with the code they deploy, to make it easy for developers to see how a change will get deployed.

### DIFF
--- a/ci/pipeline.yaml
+++ b/ci/pipeline.yaml
@@ -1,0 +1,67 @@
+resources:
+  - name: gds-way-git
+    type: git
+    source:
+      uri: https://github.com/alphagov/gds-way.git
+      branch: main
+
+  - name: deploy-to-paas-docs-space
+    type: cf
+    source:
+      api: https://api.cloud.service.gov.uk
+      username: ((cf_user))
+      password: ((cf_password))
+      organization: gds-tech-ops
+      space: docs
+jobs:
+  - name: self-update
+    serial: true
+    plan:
+    - get: gds-way-git
+      trigger: true
+    - set_pipeline: gds-way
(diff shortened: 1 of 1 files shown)


ci/pipeline.yaml
```diff
@@ -0,0 +1,51 @@
...
+            type: docker-image
+            source:
+              repository: ghcr.io/alphagov/aws-ruby
+              tag: 2.6.1
+          inputs:
+            - name: gds-way-git
+              path: repo
+          outputs:
+            - name: bundled
+          run: *bundle
```
risicle, line 45: Is this an unresolved reference?

ci/pipeline.yaml
```diff
@@ -0,0 +1,51 @@
...
+      password: ((cf_password))
+      organization: gds-tech-ops
+      space: docs
+jobs:
+  - name: self-update
+    serial: true
+    plan:
+    - get: gds-way-git
+      trigger: true
+    - set_pipeline: gds-way
```
risicle, line 22: Again, could use `self`

ci/pipeline.yaml
Krenair: Yep, well spotted

risicle (approved):
:roll_eyes: Leave it pointing at the literal pipeline if you like, but I bet if you manufactured ICBMs you'd ship them set to target the coordinates of your house by default, too...

PR #565, opened 2021-03-18
Review Continuous Delivery section

philandstuff:
I have removed some sections that don't reflect what GDS does in
practice:

 - continuous integration: we don't do trunk-based development, and we
   don't enforce or encourage keeping branches shorter-lived than a
   day or two
 - measuring continous delivery: we don't consistently measure the
   three named metrics across the organisation

I've also done some rewording and reformatting.

### DIFF
--- a/source/standards/continuous-delivery.html.md.erb
+++ b/source/standards/continuous-delivery.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Use continuous delivery
-last_reviewed_on: 2020-07-06
+last_reviewed_on: 2021-03-18
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
This looks like a good improvement to me. More concise, and closer to what GDS actually do in practice (e.g. not many teams are measuring lead time to change or mean time to recovery, however much they might like to).

source/standards/continuous-delivery.html.md.erb
```diff
@@ -1,42 +1,39 @@
...
-
-### Frequent integrations with main branch
-
-With [continuous integration][] you integrate with a main branch at least once a day, for example using source-control branching models like [Trunk-Based Development][].
+To achieve continuous delivery you need:
 
-If your team’s environment is regulated, and you need evidence that you review all code changes, raise and merge frequent small pull requests. For example, [Fourth Wall Helpful][], a client-side pull request and build status monitor for Github repositories, could help you build a team culture to support this practice.
-
-Use approaches like [feature-flagging][] or [modular architectures][] to deploy partially complete features. This reduces the size of changes going to production and encourages your team to build modular, configurable systems. 
+- automatic build promotion
```
galund, line 35: Maybe something like
```
 - minimum viable pull requests, and a team process to discourage long-lived branches (eg [Fourth Wall Helpful][])
```

source/standards/continuous-delivery.html.md.erb
```diff
@@ -1,42 +1,39 @@
...
-With continuous delivery use frequent integration with the main branch, automatic build promotion and production monitoring.
-
-### Frequent integrations with main branch
-
-With [continuous integration][] you integrate with a main branch at least once a day, for example using source-control branching models like [Trunk-Based Development][].
+To achieve continuous delivery you need:
 
-If your team’s environment is regulated, and you need evidence that you review all code changes, raise and merge frequent small pull requests. For example, [Fourth Wall Helpful][], a client-side pull request and build status monitor for Github repositories, could help you build a team culture to support this practice.
-
-Use approaches like [feature-flagging][] or [modular architectures][] to deploy partially complete features. This reduces the size of changes going to production and encourages your team to build modular, configurable systems.
```
galund, line 39: I quite liked this, too. Feature-flagging goes alongside A/B testing in a lot of cases, so maybe we could link off to something about that.

philandstuff: I've added this back in

source/standards/continuous-delivery.html.md.erb
philandstuff: I wrote some words, let me know what you think

galund (approved):

PR #576, opened 2021-04-06
Bump Review date and interval

deanwilson:
This document is quite static and relfects an area of
strategy that will not change quickly so while it's still valid
I'm also increasing the review intervals.

Also clarify that it applies to CDIO as well as GDS to ensure we're
all in agreement on where to host.


louzoid-gds (approved):
LGTM

PR #580, opened 2021-04-08
Template current year in MIT licence example

mrwilson:
Anyone coming to this page and using the example copyright notice will not have to update to the current year.


philandstuff (approved):
nice

PR #582, opened 2021-04-09
Use `main`, not `master`

philandstuff:
This makes it clear that we use `main`, not `master`, for the default
branch name.

This is open as a draft while I seek feedback on ways to make the migration easier.

### DIFF
--- a/source/standards/source-code.html.md.erb
+++ b/source/standards/source-code.html.md.erb
@@ -87,7 +87,7 @@ Before you rename your default branch, you should consider the impact this will 
 You can configure some CI systems to check out the default branch rather than hardcoding a branch name:
 
  - [GitHub Actions](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/) supports a `$default-branch` macro
- - [concourse-git-resource](https://github.com/concourse/git-resource): if you don't set `branch` then it will check out the default branch
+ - [concourse-git-resource](https://github.com/concourse/git-resource): if you do not set `branch`, it will check out the default branch
 
 ### Commit messages
 
--- a/source/standards/source-code.html.md.erb
+++ b/source/standards/source-code.html.md.erb
@@ -70,7 +70,7 @@ Update the application's README to explain why the repository has been archived,
 
 Your repository's default branch should be called `main`.
 
-To configure git to use `main` for new repositories, first ensure you are using git 2.28 or later, then run:
+To configure git to use `main` for new repositories, first make sure you are using git 2.28 or later, then run:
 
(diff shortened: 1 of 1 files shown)


thomasleese:
It's very specific to GOV.UK, but we've created some documentation to help with the transition https://docs.publishing.service.gov.uk/manual/howto-switch-default-branch.html that could be a helpful template or resource?

source/standards/source-code.html.md.erb
```diff
@@ -66,6 +66,29 @@ Update the application's README to explain why the repository has been archived,
 
 ## Working with Git
 
+### Default branch name
+
+Your repository's default branch should be called `main`.
+
+To configure git to use `main` for new repositories, first ensure you are using git 2.28 or later, then run:
```
cmcnallygds, line 73:
```suggestion
To configure git to use `main` for new repositories, first make sure you are using git 2.28 or later, then run:
```

source/standards/source-code.html.md.erb
```diff
@@ -66,6 +66,29 @@ Update the application's README to explain why the repository has been archived,
...
+
+To migrate existing repositories from `master` to `main`, you can use
+[GitHub's branch renaming
+tool](https://github.com/github/renaming#renaming-existing-branches).
+
+Before you rename your default branch, you should consider the impact this will have on:
+
+ - internal users
+ - external users
+ - your CI system (for example, Concourse pipelines or GitHub actions)
```
cmcnallygds, line 85: We should probably write out what CI stands for the first time we use it

source/standards/source-code.html.md.erb
```diff
@@ -66,6 +66,29 @@ Update the application's README to explain why the repository has been archived,
...
+
+You can configure some CI systems to check out the default branch rather than hardcoding a branch name:
+
+ - [GitHub Actions](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/) supports a `$default-branch` macro
+ - [concourse-git-resource](https://github.com/concourse/git-resource): if you don't set `branch` then it will check out the default branch
```
cmcnallygds, line 90:
```suggestion
 - [concourse-git-resource](https://github.com/concourse/git-resource): if you do not set `branch`, it will check out the default branch
```

source/standards/source-code.html.md.erb
philandstuff: I will do this but CI feels like one of those acronyms where writing it out doesn't help much. CI no longer means "Continuous Integration", it means "running tests on a server"

cmcnallygds (approved):

PR #583, opened 2021-04-09
Fix a bunch of links

philandstuff:
These were all links that had `/master/` in them, but the fix was not
the same in each case.  Some were dangling references that could be
removed, some were dead links whose content had moved elsewhere, and
some just needed `s/master/main/`.


louzoid-gds (approved):
LGTM. Thanks Phil

PR #603, opened 2021-04-29
Update to clarify that not all teams use quarterly planning

timblair:
Also extend the review period: this rarely (if ever) changes, so doesn't need reviewing every three months.


deanwilson (approved):
Merge away

PR #606, opened 2021-05-11
Update nodejs guidance

tombye:
The NodeJS guidance is overdue reviewing.

These changes are the result of a review and of comments on this pull request from when it was a draft.

Slightly contentious changes include:
- removal of references to Node boilerplate due to its deprecation
- now suggesting Koa as the framework to use for web apps instead of Express

No doubt needs some content love from @mrwilson and the review date updating prior to merge. The later will be done when the rest is approved.

### DIFF
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -496,10 +496,6 @@ versions of Node.js.
 - [I’m harvesting credit card numbers and passwords from your site. Here’s how.](https://hackernoon.com/im-harvesting-credit-card-numbers-and-passwords-from-your-site-here-s-how-9a8cb347c5b5)
 - [you might not need gulp or grunt](https://medium.freecodecamp.org/why-i-left-gulp-and-grunt-for-npm-scripts-3d6853dd22b8).
 
-## Starting a new project
-
-We recommend starting any new project with the [GDS Node.JS boilerplate](https://github.com/alphagov/gds-nodejs-boilerplate). It makes it easier to create new serves and follows the advice given here. Is also includes the GOV.UK styles and templates, StandardJS, grunt and snyk.
-
 ## Further reading
 
 In the GDS context, the guidelines provided in the links below are advisory
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -469,7 +469,7 @@ your best to reduce the risks involved:
   its popularity, author reputation and so on.
 - Use tools like [GitHub security alerts](https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies) or [Snyk](https://snyk.io/) to check packages for vulnerabilities
 
-> Use `npm init`, `npm start`, `npm script` and so on.
+> Use `npm init`, `npm start`, `npm ci`, NPM scripts and so on
 
 Making full use of NPM's features will simplify your continuous integration and deployment.
 
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -255,7 +255,7 @@ const callback = filename => (err, data) => {
 
 #### Use of `async`/`await`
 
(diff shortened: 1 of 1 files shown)


tombye:
> Now that we're all for async/await... Should we stop recommending express and start recommending Koa instead?

I agree that my change makes it a real question now but I'll defer to @sfount and the rest of the Pay team on that as:
- I have no production experience of Koa and the internet seems full of articles debating this question
- from what I can tell, the Pay team run the most actively maintained app using Express at GDS

paroxp:
PaaS team also runs (perhaps less active) app based on Express.

The reason I've mentioned that, is that we've found a need to create a wrapping middleware around each async handler we have to make the interaction easier.

We don't have any production application written in Koa, though any spike/firebreak projects we're doing, or the future project is likely to be based on it.

tombye:
After a bit more thinking and reading on the Koa question, I **think** that:
- it's well documented, has a good ecosystem and support and seems pretty [trustworthy](https://gds-way.cloudapps.digital/manuals/programming-languages/nodejs/index.html#node-package-manager-npm)
- it doesn't seem like a massive jump to work with it if you're used to ExpressJS (which is one implication of our making this change
- it was created by the same team who did ExpressJS as a better way to do similar things
- it follows these bits of advice from our NodeJS docs possibly better than Express does:
  - 'Use asynchronous versions of the Node.js API functions whenever possible'
  - 'avoid relying on packages for functionality you can easily implement yourself, especially if a package has a lot of functionality you do not need'

In short, I'd still be interested in what the Pay team think, mainly about whether anything they do with ExpressJS would be harder with Koa, but changing the advice to Koa gets my 👍🏻 .

tombye:
If @sfount or @iqbalgds from the Pay team have time it would be very useful to know your opinions on these changes and if any other updates are needed to this page 🙂

sfount:
Changes all look good to me!

On the Express vs. Koa front, I wonder if it doesn't have to be mutually exclusive. 

Until a project using Koa is proved as an easy-to-contribute-to, production ready codebase I think it makes sense to continue advising use of Express, with Koa as an acceptable alternative given all of the well articulated points in the above commit message (spiritual successor, well maintained, etc.).

There are some sharp edges (actually quite reasonable but things to consider) with moving over. Koa is definitionally less opinionated and code bases using the Koa framework can read quite differently depending on who is implementing them. 

In order to replace the Express framework core library, for example, you would need
https://github.com/koajs/koa
https://github.com/koajs/router
https://github.com/koajs/bodyparser
https://github.com/koajs/session

This isn't an issue in particular but does feel like something we should be able to easily answer before explicitly advising developers to use a particular framework.

To ➕ @paroxp 's experience, I've also used Koa in spin-off projects and may well have chosen it for new Pay frontends a few years ago if the guidance wasn't so clear on Express.

---

It's also worth noting that the next version of the Express framework's router will actually be Promise-y by default and that's already implemented https://github.com/expressjs/express/issues/2259. However the release timeline on a project depended on as much as Express shouldn't be relied on in our decision making.

tombye:
@sfount those are great points, some of which I think I was trying to say in previous comments but you put much better :)

I like the solution of Express as a default with an option to use Koa if it suits your problem, like with @paroxp 's future situation. @chao-xian also asked outside of here if suggesting Koa would produce a requirement for existing Express projects to migrate which would rarely make sense unless the substantial work involved would unblock existing issues. I think this is interpretation is a another potential problem your solution dodges nicely.

Thinking about your point on the greater freedom Koa allows around how to do core web app features, I wonder if a note saying developers deciding to use Koa should follow the patterns, and use any associated middleware/modules, in [the koa examples repo](https://github.com/koajs/examples) would be enough guidance to ensure the consistency needed?

tombye:
@paroxp @sfount @chao-xian from the discussion so far it seemed to me that:
- the present guidance could block legitimate uses of Koa
- we could do with enough experience using Koa to advise on the best way to approach common web app problems before we suggest it over Express

That being the case, I:
1. created [an issue on our guidance around using Koa](https://github.com/alphagov/gds-way/issues/618) so that discussion has somewhere to live and doesn't block the other changes here
2. changed the frameworks section back to suggesting Express but with a caveat that Koa could be used, with the right understanding of the differences in its design with Express, to unblock the kind of uses you both referenced in your comments

Does this seem sensible and, if so, could we merged these changes?

paroxp (approved):

PR #622, opened 2021-07-02
Document still valid so Bump Review Date

deanwilson:
This is document was valid and while there is work happening to
update it it's been awaiting review for six months so we'll accept
what we have and add the new details as they are ready.


louzoid-gds (approved):
LGTM

PR #623, opened 2021-07-02
Add information on GitHub Actions and workflows

katstevens:
The GOV.UK Pay team have been trialling using Github Actions for continuous integration and image builds. We've captured some of the security concerns, and come up with a list of best practices around permissions (and the use of third party actions) for our own docs. 

This feels like a useful thing to share more widely, especially given the [potential issues around unversioned third party actions](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions). Version tags like `dodgyaction@v3` are mutable: it's recommended to pin to a Git SHA instead so that if the provider releases a compromised version under that tag, the workflow will not automatically pick up the bad action. 

The other point that may not be obvious for new Github Actions users is that **the output of workflow runs is public on public repos**. A traceback from a failed unit test run is probably fine, but if a team is trying to do something more complex or sensitive, they should consider carefully whether Github Actions is the right tool for the job. 

I've added this information to the existing section on GitHub - happy to move if there's a better place for it.

### DIFF
--- a/source/standards/source-code.html.md.erb
+++ b/source/standards/source-code.html.md.erb
@@ -88,6 +88,7 @@ Third-party actions should only be used if:
 
 Note that for public repositories, the output of workflow runs is visible to everyone. Do not use workflows if this output could be considered sensitive.
 
+Consider creating a Workflow Template in the [alphagov workflow folder](https://github.com/alphagov/.github/tree/main/workflow-templates) if you need to share a similar workflow between many repositories.
 
 ## Working with Git
 
--- a/source/standards/source-code.html.md.erb
+++ b/source/standards/source-code.html.md.erb
@@ -64,6 +64,31 @@ If an application is no longer used in production, you should [archive its repos
 
 Update the application's README to explain why the repository has been archived, and link to a new location if the application has been superseded.
 
+### Using Github Actions and workflows
+
+Alphagov repositories can be configured to use [GitHub Actions](https://docs.github.com/en/actions) for CI/CD jobs, for example running unit tests or deploying a static site.
+
+Actions and workflows can be powerful, so take care to follow best security practice.
+
+Ensure the repository permissions follow [the principle of least privilege](/standards/principle-least-access.html).
+Set workflow permissions to read-only by default, and disable all actions on repositories that are not using them (this will hide the 'Actions' tab from the repository menu).
+
+If your repository has external contributors, [ensure they do not have permissions to add workflows or trigger workflow runs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).
+
+If your workflow interacts with another resource (for example AWS or DockerHub), consider setting up a dedicated account or role, with permissions limited to the scope of the action.
+
+[Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible. If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions).
+Third-party actions should only be used if:
+
+- The provider is verified by GitHub (for example, [aws-actions](https://github.com/aws-actions))
+- The action is complex enough that you cannot write your own local action
+- You have fully reviewed the code in the version of the third-party action you will be using
+- You have pinned the specific version in your workflow and in the repository settings, using a Git commit SHA
+- The third-party action is actively maintained, well-documented and tested ([follow the guidance on third party dependencies](/standards/tracking-dependencies.html)).
+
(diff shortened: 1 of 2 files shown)


gidsg:
Also worth adding Github Actions to this list? https://github.com/alphagov/gds-way/blob/3fde1dffacad589e3583025105e8542030167d42/source/standards/source-code.html.md.erb#L40

katstevens:
> Also worth adding Github Actions to this list?
> 
> https://github.com/alphagov/gds-way/blob/3fde1dffacad589e3583025105e8542030167d42/source/standards/source-code.html.md.erb#L40

Good spot! I've tacked it onto the last commit.

gidsg:
We have found it quite useful, where we have many repos with similar workflows to use [Workflow Templates](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template)

How about adding something like

> Consider creating a [Workflow Template](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template) in https://github.com/alphagov/.github/tree/main/workflow-templates if you need to share a similar workflow between many repositories.

source/standards/source-code.html.md.erb
```diff
@@ -64,6 +64,31 @@ If an application is no longer used in production, you should [archive its repos
...
+Alphagov repositories can be configured to use [GitHub Actions](https://docs.github.com/en/actions) for CI/CD jobs, for example running unit tests or deploying a static site.
+
+Actions and workflows can be powerful, so take care to follow best security practice.
+
+Ensure the repository permissions follow [the principle of least privilege](/standards/principle-least-access.html).
+Set workflow permissions to read-only by default, and disable all actions on repositories that are not using them (this will hide the 'Actions' tab from the repository menu).
+
+If your repository has external contributors, [ensure they do not have permissions to add workflows or trigger workflow runs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).
+
+If your workflow interacts with another resource (for example AWS or DockerHub), consider setting up a dedicated account or role, with permissions limited to the scope of the action.
```
richardTowers, line 78: Maybe work for a future PR, but I'd like to see some discussion of how to store secrets.

When should they be stored at the repository level? When should they be stored at the organisation level?

If you have lots of repositories that access the same resource (say DockerHub), presumably you should avoid duplicating the credentials for that service across all the repositories (as that would be a headache if you needed to rotate the token). Org level secrets seem to solve this problem (you can scope them to specific repos, I think). But I'm not sure how much GitHub permission you need to set those - only org owners perhaps?

On the subject of org owners, is it worth mentioning that org owners from other teams in alphagov may be able to extract your credentials?

gidsg: @richardTowers yeah, I had a quick look at org secrets and it says:
> Only alphagov administrators may add organization secrets.

gidsg: @richardTowers out of interest do you know how other org owners may extract credentials. It seems from my limited testing that repo level secrets are not viewable by others once set i.e. similar to AWS secret keys in that way, although I suppose someone with permission to alter a workflow could extract a secret by executing a job.

katstevens: Yep - secrets and tokens are a whole 'nother can of worms (agree it might be better in another PR). 

That reminds me of another permissions wrinkle I forgot to mention: Pay were looking at using a workflow to create a tag. To do that, we would have needed to change the workflow permissions on the repo to [`write: contents`](https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-contents), which basically would grant any third party action in that workflow a whole bunch of extra permissions 🙀 This is [a long-running issue with Github's permission model](https://github.com/dear-github/dear-github/issues/113), but if someone was using a potentially dodgy third party action, the attack surface would be much bigger.

richardTowers: @gidsg - yeah, that's my understanding too. I was thinking of an org owner modifying a workflow, or exposing an org level secret to another repo under their control and then echo'ing it in their own workflow. I don't think they could do that without leaving some audit trail, but they could probably do it without being spotted.

katstevens:
I've added a line about Workflow Templates, thanks @gidsg.

As per the conversation above, managing secrets is probably a section in itself. I'm not 100% sure what's best to recommend there for the organisation as a whole yet, so I'll leave that out for this PR. @louzoid-gds I think we're ready for review 👍

louzoid-gds (approved):
Thanks again for writing this up Kat 👍

katstevens:
Of course, the other thing we should really do is review the Github Actions workflows and settings on this repository, and follow our own advice 😹

PR #625, opened 2021-07-13
Bump a bunch of needing-review guidance

timblair:
I also ran the lice site through http://validator.w3.org/checklink and attempted to fix any issues it raised, including updating links that worked but resulted in a permanent redirect.

PR #626, opened 2021-07-14
Fix various broken and redirecting links

timblair:
Following on from #625, fixing a few more link issues highlighted by http://validator.w3.org/.


deanwilson (approved):
Merge when green

PR #628, opened 2021-07-21
Update Java style guide (July 2021 revisions)

alexbishop1:
The Java community meeting reviewed the [Java style guide](https://gds-way.cloudapps.digital/manuals/programming-languages/java.html) on 21 July 2021.

There was rough consensus on some changes, which are detailed in the individual commit messages.

### DIFF
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -168,7 +168,7 @@ You should use either Gradle or Maven as the build tool. Use recent versions if 
 
 ## Web frameworks
 
-We use [Dropwizard](https://www.dropwizard.io/) as our web framework of choice.
+The [Dropwizard](https://www.dropwizard.io/) web framework is used widely within GDS.
 
 Dropwizard 2.0 was released at the end of 2019. There is a [guide for migrating from Dropwizard 1.3.x to 2.0.x](https://github.com/dropwizard/dropwizard/wiki/Upgrade-guide-1.3.x-to-2.0.x). Teams within GDS have found the upgrade straightforward.
 
--- a/source/manuals/programming-languages/java.html.md.erb
+++ b/source/manuals/programming-languages/java.html.md.erb
@@ -192,13 +192,13 @@ We recommend publishing artifacts (for which the source is already public) to [M
 
 You should _not_ use Maven Central to publish artifacts for which the source is closed or contains other proprietary assets, as it is a public repository with anonymous access.
 
-### Claiming group ids
+### Claiming group IDs
 
-You will need to claim one or more [group ids](https://maven.apache.org/guides/mini/guide-naming-conventions.html) in order to publish an artifact to Maven Central. The central repository is run and operated by Sonatype, and you will need to register for an account on their JIRA instance to request control of a group id.
+You will need to claim one or more [group IDs](https://maven.apache.org/guides/mini/guide-naming-conventions.html) in order to publish an artifact to Maven Central. The central repository is run and operated by Sonatype, and you will need to register for an account on their Jira instance to request control of a group ID.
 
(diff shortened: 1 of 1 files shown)


j-maynard (approved):
🦁

SandorArpa (dismissed):
LGTM

mrwilson (approved):

source/manuals/programming-languages/java.html.md.erb
```diff
@@ -91,6 +93,16 @@ var checkResponse = service.getCheckResponse();
 
 Be mindful that `var` hides the type of the variable. If the variable name makes the type obvious, this usually isn’t a problem. But if it is not clear from either the variable name or the right-hand side of the assignment, it might be better to explicitly write the type.
 
+If you use `var` in place of a generic type, this will usually prevent you from using the so-called diamond operator (itself a form of type inference) on the right-hand side of the assignment:
```
willp-bl, line 96: This seems to be implying that var example below won't work

alexbishop1: Does it? The intention is to first show a typical use of the diamond operator, then show how if we replace the type with `var` we can no longer use the diamond operator and have to specify the generic type on the right-hand side.

willp-bl: it might be clearer perhaps to say that when using var you have to specify the type on the right hand side of the assignment, rather than on the left with a diamond operator on the right

JohnWattsGDS:
```suggestion
If you use `var` in place of a generic type you should specify the generic type inside the diamond operator on the right hand side.  If you don't, the compiler will give you a `Foo<Object>` which you probably don't want (and if you do want that, you should make that clear by using `<Object>` instead of `<>`):
```

source/manuals/programming-languages/java.html.md.erb
```diff
@@ -91,6 +93,16 @@ var checkResponse = service.getCheckResponse();
...
+
+```java
+// Java 9
+Set<String> usernames = new HashSet<>();
+
+// Java 10+
+var usernames = new HashSet<String>();
```
JohnWattsGDS, line 103:
```suggestion
var usernames = new HashSet<String>();
// Note that:
var usernames = new HashSet<>(); // will compile but is equivalent to var usernames = new HashSet<Object>();
```

source/manuals/programming-languages/java.html.md.erb
alexbishop1: I’ve reworded it. How’s it read now?

source/manuals/programming-languages/java.html.md.erb
```diff
@@ -101,8 +113,10 @@ Keep in mind that improvements to the Java standard library mean that some exter
 
 When using external libraries, favour those which complement the Java standard library. For example, be wary of any library which introduces a new type that replicates the functionality of an existing type in the Java standard library without implementing the same interfaces (for example, a list type which does not implement `java.util.List`).
 
+Make sure any external library you use is appropriate for your purposes and avoid relying on internal implementation details of external libraries. If your IDE’s code completion suggests a method from an external library, make sure it’s a supported part of the library’s defined API.
```
willp-bl, line 116: thoughts on adding in some detail about accidentally relying on transitive dependencies?

alexbishop1: I did half-write a sentence about this but since it wasn’t actually something we discussed in the meeting, I left it out for now.

source/manuals/programming-languages/java.html.md.erb
alexbishop1: Updated the example to have something similar.

JohnWattsGDS (approved):
🚀

mrwilson (approved):

willp-bl (approved):
LGTM

Issue #617, opened 2021-06-08
Update HTML guidance


chris-gds:
After a short review we noted there are various aspects that are out of date and could go with an update especially around 
`aria-labelledby=` guidance and `<section>` use.

https://gds-way.cloudapps.digital/manuals/programming-languages/html.html

timblair:
@chris-gds Thank for looking. Is there any more detail you can provide on what aspects need updating? Or even better, a PR...?

chris-gds:
👋🏼  @timblair - Sure, more details below - initial thoughts centre around [`section`](https://gds-way.cloudapps.digital/manuals/programming-languages/html.html#sectioned-content) use.  This most likely could be tested further across assistive tech.

```
<section aria-labelledby="introduction">
  <h2 id="introduction">Introduction</h2>
  <p>People have been catching fish for food since before recorded history...</p>
</section>

<section aria-labelledby="equipment">
  <h2 id="equipment">Equipment</h2>
  <p>The first thing you'll need is a fishing rod or pole that you find comfortable
    and is strong enough for the kind of fish you're expecting to try to land...</p>
</section>
```

- I don't think this pattern is used commonly within GOV.UK nor is it present in the [Design System](https://design-system.service.gov.uk/)
- After testing it might be not needed
- Or the reverse might be true and many parent containers that contain headings should be updated accordingly as well as the relevant publishing apps.

(I've also requested time to see if I can review / update in lot more detail).

timblair:
@chris-gds Did you manage to carve out any more time to to a full review of the HTML guidance?

tombye:
I noticed the ARIA in HTML spec' states that the technique of using `aria-labelledby` to assign an accessible name to `<section>`s gives them the role of `region`. Without it, they have no role:

https://w3c.github.io/html-aria/#el-section

Should be a consideration if this is added in this guidance.

tombye:
Myself and @chris-gds had a very long discussion about this outside of here and decided that the current guidance on use of `<section>` needs to change and that I'll create a PR to that end. When I do, I'll update this comment and close this issue.

tombye:
I haven't found time to put together a PR for the proposed changes and, after discussing it with others, I intend to consult with the frontend community first. I'll update the status of that here when it happens.

PR #630, opened 2021-07-23
Update last-review date on the HTML style guide

timblair:
#617 notes the need for some parts of this guidance to be reviewed, but this has been raised in community meetings a few times and the position is that it's still broadly correct and there are no significant changes required.

Marking this as reviewed, but reducing the review time to three months.


Nooshu (approved):
Approved, will look to incorporate #617 in the coming months.

PR #631, opened 2021-07-26
New page on Disaster Recover added to the operating a service section

louzoid-gds:
Big thanks to Claire and the tech writers for the reviews. Other comments/amends very welcome.

### DIFF
--- a/source/standards/disaster-recovery.html.md.erb
+++ b/source/standards/disaster-recovery.html.md.erb
@@ -35,7 +35,7 @@ build pipelines
 
 ### Back up your data
 
-You should back up your data. How often you back up will depend on your [Recovery Point Objective (RPO)](https://cloud.google.com/architecture/dr-scenarios-planning-guide#basics_of_dr_planning). An RPO determines what you consider an acceptable loss of data between the last backup and the point of data loss. For example, if you back up your data nightly, your RPO is 24 hours, the length of time between the backups. In practice, many modern database service cloud providers provide [point-in-time recovery](https://docs.aws.amazon.com/aws-backup/latest/devguide/point-in-time-recovery.html) from backups. This can result in an RPO of around 5 minutes or less.
+You should back up your data. How often you back up will depend on your [Recovery Point Objective (RPO)](https://cloud.google.com/architecture/dr-scenarios-planning-guide#basics_of_dr_planning). An RPO determines what you consider an acceptable loss of data between the last backup and the point of data loss. For example, if you determine that up to 24 hours of data loss is an acceptable risk, you will only need to backup your data once daily. For lower RPOs (minutes or seconds), you should use [continuous backups and point-in-time recovery](https://docs.aws.amazon.com/aws-backup/latest/devguide/point-in-time-recovery.html) services that most modern cloud providers offer.
 
 ### Have cold or offline backups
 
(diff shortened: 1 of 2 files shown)


source/standards/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,77 @@
...
+- Domain Name System (DNS) provider
+- Content Delivery Network (CDN)
+- any Virtual Private Networks (VPNs) your development team might use to access production systems
+build pipelines
+- external library and container registries (for example, RubyGems, PyPI or Docker Hub)
+- secrets management providers
+
+### Back up your data
+
+You should back up your data. How often you back up will depend on your [Recovery Point Objective (RPO)](https://cloud.google.com/architecture/dr-scenarios-planning-guide#basics_of_dr_planning). An RPO determines what you consider an acceptable loss of data between the last backup and the point of data loss. For example, if you back up your data nightly, your RPO is 24 hours, the length of time between the backups. In practice, many modern database service cloud providers provide [point-in-time recovery](https://docs.aws.amazon.com/aws-backup/latest/devguide/point-in-time-recovery.html) from backups. This can result in an RPO of around 5 minutes or less.
```
timblair, line 38: The description of RPOs is sensible, but the examples are backwards (cause vs. effect). An RPO is an objective that you want to meet, and you make choices based on that. So rather than "if you back up your data nightly, your RPO is 24 hours" it should be "if your RPO is 24 hours, you might choose to take nightly backups." Likewise,  point-in-time recovery ... can result in an RPO of around 5 minutes or less" should be "if you determine that you need a 5 minute RPO, then point-in-time snapshotting and recovery might be necessary."

source/standards/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,77 @@
...
+- the backups are encrypted and versioned
+- the account you are copying the backups from only has write permissions
+
+### Decide where to store your backups
+
+Where you are storing your backups is also important for disaster recovery. Most cloud storage solutions have durability guarantees. However, an outage could occur in the region containing your live service and backups. If your service has very high availability requirements, you may need copies of your data in a different region, or with a different service provider. Your conversations with IA and service owners can help you determine your risk tolerance for these hopefully rare scenarios.
+
+The National Cyber Security Centre (NCSC) article [Offline backups in an online world](https://www.ncsc.gov.uk/blog-post/offline-backups-in-an-online-world) further explains why offline backups are important and suggests some approaches to use in a cloud environment.
+
+### Decide on manual or automatic recovery in the event of a disaster
```
timblair, line 55: This section feels like it's bundling a few different things together:

* Setting RTOs to drive decisions about how you design your recovery mechanisms
* The fact that different elements of a service (or the same elements under different conditions) might require different RTOs
* Different uses of manual vs. automatic recovery, and the fact that these also have different cost and complexity trade-offs
* That risk and service owners are the ones who ultimately own (set) the RTOs and RPOs

Possibly because of this mixture, the current section title feels very reductive, and not particularly descriptive of what you're trying to convey.

timblair (changes requested):
Nice work! A couple of sections that I think still need some work:

* RPOs to drive decisions on data backups
* RTOs to drive decisions on recovery mechanisms

I've left separate in-line comments on each of those points.

source/standards/disaster-recovery.html.md.erb
louzoid-gds: Ah yes! I totally forgot about the thread on this in Slack. I'll have a go at a reword

source/standards/disaster-recovery.html.md.erb
louzoid-gds: Yep you're right. I'll change it

louzoid-gds:
Amends based on Tim's comments made. More reviews welcome

timblair (approved):

PR #632, opened 2021-07-28
Update govuk_tech_docs to the latest version

timblair:
This includes a number of accessibility fixes, but one a11y issue remains, so the accessibility statement has been updated to reflect this.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,8 +8,8 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
-    autoprefixer-rails (10.2.5.0)
-      execjs (< 2.8.0)
+    autoprefixer-rails (10.2.5.1)
+      execjs (> 0)
     backports (3.21.0)
     chronic (0.10.2)
     chunky_png (1.4.0)
@@ -17,8 +17,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.12.2)
-    commonmarker (0.21.2)
-      ruby-enum (~> 0.5)
+    commonmarker (0.22.0)
     compass (1.0.3)
       chunky_png (~> 1.2)
       compass-core (~> 1.0.2)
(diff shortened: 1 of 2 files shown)


tombye:
I don't really do much ruby so might be wrong here but shouldn't the first commit include the `Gemfile` too?

timblair:
Nope, the Gemfile just says `gem 'govuk_tech_docs'`. It's the Gemfile.lock that sets the version.

tombye (changes requested):
The changes to the statement look good. I assume this will just pull in version `2.4.2` of `govuk_tech_docs`? it didn't for me locally running `bundle install` but testing with that version looked good.

The only change I'd like is for the `Gemfile` to lock `ffi` to `1.12`, (like https://github.com/alphagov/datagovuk-tech-docs/pull/60), to stop it blowing up for users on OSX :)

timblair:
> I assume this will just pull in version 2.4.2 of govuk_tech_docs? it didn't for me locally running bundle install but testing with that version looked good.

Are you sure you were on the correct branch? I just did a fresh clone and `bundle install` on the `bump-tech-docs` branch and it installed `2.4.2` as expected.

> The only change I'd like is for the Gemfile to lock ffi to 1.12, (like alphagov/datagovuk-tech-docs#60), to stop it blowing up for users on OSX :)

Is this a common thing? The GDS Way repo has been using a version greater than `1.12` for a fair while. I've also never seen the problem listed in that issue (using MacOS). `ffi` is being pulled in as a dependency from `govuk_tech_docs`. If it needs pinning, it should be done there?

tombye:
> Are you sure you were on the correct branch? I just did a fresh clone and `bundle install` on the `bump-tech-docs` branch and it installed `2.4.2` as expected.

I am sure about the branch. I just tried again and got `2.4.2` so maybe a timing thing? Either way, I'm happy this is not a problem.

>Is this a common thing? The GDS Way repo has been using a version greater than 1.12 for a fair while. I've also never seen the problem listed in that issue (using MacOS).

There are a few GDS'ers on that issue reporting experiencing it and I know of a few others doing the same on Slack a while ago. From reading the issue, it looks like it might effect people on Mohave only:

`An upgrade to MacOS10.15 should help`

>ffi is being pulled in as a dependency from govuk_tech_docs. If it needs pinning, it should be done there?

Makes sense. I'll open an issue on that repo instead.

tombye (approved):
Looks good to me 👍🏻

PR #633, opened 2021-08-06
Add guidance about use of password managers

benthorner:
This is based on some initial discussions in Slack. It's meant to
complement the existing doc on managing accounts, noting that the
secrets we store may include more than just account logins.

### DIFF
--- a/source/standards/storing-credentials.html.md.erb
+++ b/source/standards/storing-credentials.html.md.erb
@@ -25,7 +25,8 @@ Some of the third-party password managers people use at GDS include:
 - [**LastPass**](https://www.lastpass.com/) - Approved by Information Assurance and available to install in the GDS Self Service app. However, the free version can only be used on one type of device.
 - [**1Password**](https://1password.com/)
 - [**BitWarden**](https://bitwarden.com/) - An open source password manager.
-- [**KeePassXC**](https://keepassxc.org/) -  An offline password store, which you may want to backup somewhere.
+- [**KeePassXC**](https://keepassxc.org/) - An offline password store, which you may want to backup somewhere.
+- [**QtPass**](https://qtpass.org/) - Another offline store, integrated with Git and GPG / pass.
 
 ⚠️  Warning! Browser extensions which autofill credentials have [serious security concerns](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). If do use an extension, consider configuring so that you manually copy / paste credentials instead of allowing it to autofill them.
 
--- a/source/standards/storing-credentials.html.md.erb
+++ b/source/standards/storing-credentials.html.md.erb
@@ -36,7 +36,17 @@ Examples include shared login details for:
 - Software repositories e.g. npmjs.org, rubygems.org
 - Admin portals e.g. Fastly, DockerHub
 
-There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG. However, [this is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and we strongly encourage you to look around for possible alternatives. **Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
+[Follow the guidance for managing team credentials in the fist instance](/standards/accounts-with-third-parties.html).
+
+There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG. **Make sure you configure the repo as [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help prevent accidental disclosure.
+
+Using GPG in this way isn't ideal because:
+
+- We can't revoke access to credentials, unless we change them. Anyone who was given access can still decrypt credentials using their local copy of the repo and its commit history.
+
+- It creates a high barrier to entry, especially for people less familiar with Unix tooling. We've found GPG tools are [generally difficult to use](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and keyservers are unreliable.
+
+Investigate alternatives before adopting GPG-based credential stores for new teams.
 
 ## Service credentials
 
--- a/source/standards/storing-credentials.html.md.erb
+++ b/source/standards/storing-credentials.html.md.erb
@@ -27,7 +27,7 @@ Some of the third-party password managers people use at GDS include:
 - [**BitWarden**](https://bitwarden.com/) - An open source password manager.
 - [**KeePassXC**](https://keepassxc.org/) -  An offline password store, which you may want to backup somewhere.
 
-**Do not use browser extensions that autofill credentials**, as [these are fundamentally vulnerable to attack](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). If you do use an extension, make sure it's configured so that you manually copy / paste credentials.
+⚠️  Warning! Browser extensions which autofill credentials have [serious security concerns](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). If do use an extension, consider configuring so that you manually copy / paste credentials instead of allowing it to autofill them.
 
 ## Team credentials
 
--- a/source/standards/storing-credentials.html.md.erb
+++ b/source/standards/storing-credentials.html.md.erb
@@ -1,0 +1,48 @@
+---
+title: How to store credentials
+last_reviewed_on: 2021-08-06
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+Depending on how you [manage your accounts](/standards/accounts-with-third-parties.html), you, your team and the service you run may have credentials or other secrets that you need to store securely.
+
+## Personal credentials
+
+Examples include login details for:
+
+- GitHub
+- AWS
+- GOV.UK Signon
+
+[Use the password manager built into your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion).
+
+If you can't do this for some reason, then it's OK to use a third-party password manager. This could be necessary if your browser has an accessibility issue, or if you work with multiple browsers.
+
+Some of the third-party password managers people use at GDS include:
+
+- [**LastPass**](https://www.lastpass.com/) - Approved by Information Assurance and available to install in the GDS Self Service app. However, the free version can only be used on one type of device.
+- [**1Password**](https://1password.com/)
+- [**BitWarden**](https://bitwarden.com/) - An open source password manager.
+- [**KeePassXC**](https://keepassxc.org/) -  An offline password store, which you may want to backup somewhere.
+
+**Do not use browser extensions that autofill credentials**, as [these are fundamentally vulnerable to attack](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). If you do use an extension, make sure it's configured so that you manually copy / paste credentials.
+
+## Team credentials
+
+Examples include shared login details for:
+
+- Software repositories e.g. npmjs.org, rubygems.org
+- Admin portals e.g. Fastly, DockerHub
+
+There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG. However, [this is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and we strongly encourage you to look around for possible alternatives. **Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
+
+## Service credentials
+
+Examples include:
+
+- Functional account credentials e.g. API keys
+- Sensitive configuration (e.g. IP block lists)
+
+Some teams have "secrets" repositories and use GPG to encrypt files, which are then decrypted during the deployment process. As above, [storing secrets this way is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and we strongly encourage you to look around for possible alternatives. **Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
--- a/source/standards/storing-credentials.html.md.erb
+++ b/source/standards/storing-credentials.html.md.erb
@@ -16,7 +16,7 @@ Examples include login details for:
 - AWS
 - GOV.UK Signon
 
-[Use the password manager built into your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion).
+[Use the password manager built into your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion). This is simpler than setting up an extra account with a third party and avoids some of the potential issues below.
 
 If you can't do this for some reason, then it's OK to use a third-party password manager. This could be necessary if your browser has an accessibility issue, or if you work with multiple browsers.
 
--- a/source/standards/storing-credentials.html.md.erb
+++ b/source/standards/storing-credentials.html.md.erb
@@ -55,4 +55,6 @@ Examples include:
 - Functional account credentials e.g. API keys
 - Sensitive configuration (e.g. IP block lists)
 
-Some teams have "secrets" repositories and use GPG to encrypt files, which are then decrypted during the deployment process. As above, [storing secrets this way is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and we strongly encourage you to look around for possible alternatives. **Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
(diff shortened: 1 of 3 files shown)


source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,48 @@
...
+- AWS
+- GOV.UK Signon
+
+[Use the password manager built into to your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion).
+
+If you can't do this for some reason, then it's OK to use a third-party password manager. This could be necessary if your browser has an accessibility issue, or if you work with multiple browsers.
+
+Some of the third-party password managers people use at GDS include:
+
+- **LastPass** - Approved by Information Assurance and available to install in the GDS Self Service app. However, you can only use the free version on one type of device.
```
benjamineskola, line 25: Perhaps “the free version can only be used on one type of device". As worded, it might be interpreted as "only the free version is permitted to be used" or similar.

CrystalPea (dismissed):
I think it's a good summary of the discussion we've had, and sounds like a useful addition to the GDS way knowledge base 👏🏼 

I agree with @benjamineskola about the wording change on line 25, his version would make that sentence clearer.

benthorner:
Thanks for your reviews @benjamineskola and @CrystalPea!

I'm going to leave this up for a couple of days next week and then look at responding to all the comments.

source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,48 @@
...
+
+## Personal credentials
+
+Examples include login details for:
+
+- GitHub
+- AWS
+- GOV.UK Signon
+
+[Use the password manager built into to your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion).
```
ChrisBAshton, line 19: Spotted a typo:

```suggestion
[Use the password manager built into your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion).
```

Might be worth adding a bit of context to this paragraph too, e.g. "...because it provides autofill capabilities that are not exploitable in the same way as third party password managers".

source/standards/storing-credentials.html.md.erb
ChrisBAshton: Because all of the examples below are linked, we should be consistent here too:

```suggestion
- **[LastPass](https://www.lastpass.com/)** - Approved by Information Assurance and available to install in the GDS Self Service app. However, you can only use the free version on one type of device.
```

source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,48 @@
...
+**Do not use browser extensions that autofill credentials**, as [these are fundamentally vulnerable to attack](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). If you do use an extension, make sure it's configured so that you manually copy / paste credentials.
+
+## Team credentials
+
+Examples include shared login details for:
+
+- Software repositories e.g. npmjs.org, rubygems.org
+- Admin portals e.g. Fastly, DockerHub
+
+There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG. However, [this is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and we strongly encourage you to look around for possible alternatives. **Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
```
ChrisBAshton, line 39: Whilst this is all true, it reads as though we shouldn't use GPG for storing credentials, when the conclusion of the Slack thread was pretty much that it's the best/only approach we have right now (in terms of working with SSH keys uploaded to GitHub). Perhaps a rephrase like this would be better:

```suggestion
There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG, based on people's SSH keys uploaded to GitHub. At the time of writing, there are no clear alternatives, but it is worth considering other approaches as [GPG is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html).

**Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
```

source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,48 @@
...
+
+## Service credentials
+
+Examples include:
+
+- Functional account credentials e.g. API keys
+- Sensitive configuration (e.g. IP block lists)
+
+Some teams have "secrets" repositories and use GPG to encrypt files, which are then decrypted during the deployment process. As above, [storing secrets this way is fundamentally insecure](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and we strongly encourage you to look around for possible alternatives. **Any GitHub repositories that contain secrets should be [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-repository-visibility)** to help limit access by former staff.
```
ChrisBAshton, line 48: I don't think there's much value in duplicating the content above:

```suggestion
Some teams have "secrets" repositories and use GPG to encrypt files, which are then decrypted during the deployment process. As above, this is not very secure and we encourage you to look around for possible alternatives.
```

ChrisBAshton (commented):
Thanks for putting this together, Ben ⭐

source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,48 @@
...
+If you can't do this for some reason, then it's OK to use a third-party password manager. This could be necessary if your browser has an accessibility issue, or if you work with multiple browsers.
+
+Some of the third-party password managers people use at GDS include:
+
+- **LastPass** - Approved by Information Assurance and available to install in the GDS Self Service app. However, you can only use the free version on one type of device.
+- [**1Password**](https://1password.com/)
+- [**BitWarden**](https://bitwarden.com/) - An open source password manager.
+- [**KeePassXC**](https://keepassxc.org/) -  An offline password store, which you may want to backup somewhere.
+
+**Do not use browser extensions that autofill credentials**, as [these are fundamentally vulnerable to attack](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). If you do use an extension, make sure it's configured so that you manually copy / paste credentials.
```
richardTowers, line 30: This is quite a strong stance for us to take. I imagine there are lots of people at GDS (myself included, although I'm changing that now) who use browser extensions which autofill passwords.

I understand and mostly agree with the arguments in Tavis' article. But if we really want to take the "autofill extensions considered harmful" stance at GDS, I think we probably need to do more work than just publishing a piece of guidance. We'd need a whole comms campaign to get people to change their behaviour. Which we'd need to be very careful didn't do more harm than good (I wouldn't want misinformation like "GDS say password managers are bad" to spread).

As an intermediate step, perhaps we could water this down to a warning, like:

> ⚠️  Warning! Browser extensions which autofill credentials have [serious security concerns](https://lock.cmpxchg8b.com/passmgrs.html#interprocess-communication). Carefully consider the risks before using a browser extension. If do use an extension, consider configuring so that you manually copy / paste credentials instead of allowing it to autofill them.

source/standards/storing-credentials.html.md.erb
richardTowers: We should have something here saying "use individual accounts where possible". Something like:

> Wherever possible, give each person who needs access to these systems their own account. If possible, use single sign on from Google or GitHub, as this avoids the need for a separate process to add and remove accounts for joiners and leavers. Ensure that multi-factor authentication is enabled wherever possible.

> Sometimes shared accounts are necessary (for example for non-human users, like bots and build systems). In these cases try to minimise the privileges given to the shared account, and store the credentials in a secure system which the team can access. Consider putting a process in place to rotate these credentials periodically (e.g. every 6 months) to limit the potential for misuse by former staff.

> There's no established best practice for storing shared credentials at GDS... etc.

I think "fundamentally insecure" is too strong. GPG is bad, sure. But "fundamentally insecure" implies to me that it's broken at such a fundamental level that anyone can pwn me and get my passwords. Which I don't think is the case.

My biggest gripe with GPG is that it's extremely difficult to use. That has a security impact itself, but I'm more bothered by the time we waste working around obscure bugs, and the impact having to use such difficult tools has on our teams. Do frontend developers need access to your team secrets? Great, now they've got to learn how to debug an inevitably broken GPG setup. What about designers? User researchers? Can they even install GPG on their machines? The tooling ends up excluding and othering people, leading to an unnecessary two-tier situation where only developers can access some systems.

Maybe:

> There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG. However, using GPG in this way creates a very high barrier to entry, and can be costly to maintain. Investigate alternatives before adopting GPG based credential stores for new teams.

... maybe with a footnote where we can properly lay into GPG.

source/standards/storing-credentials.html.md.erb
richardTowers: I would suggest best practice here would be to use the secrets management infrastructure provided by your infrastructure or cloud provider. e.g. AWS Secrets Manager, or HashiCorp Vault.

Again, I think "fundamentally insecure" is too strong for GPG. It's more "fundamentally a pain in the arse".

source/standards/storing-credentials.html.md.erb
benjamineskola: > Some teams have a "credentials" GitHub repository and use pass to store credentials encrypted with GPG.

Regarding this specifically: is it worth mentioning the issues with this approach, regarding removing access when the historical versions of the password store may remain accessible?

source/standards/storing-credentials.html.md.erb
ChrisBAshton: Good points, Richard. Agreed - we should be careful about misinformation that could stem from this.

source/standards/storing-credentials.html.md.erb
benthorner: Sure, I've amended with that 👍.

source/standards/storing-credentials.html.md.erb
benthorner: Good points everyone! 

> We should have something here saying "use individual accounts where possible"

We do have [this guidance](https://gds-way.cloudapps.digital/standards/accounts-with-third-parties.html) already, which I'm trying not to duplicate but we could make the link to it more obvious.

How about this?

```
[Follow the guidance for managing team credentials in the fist instance](/standards/accounts-with-third-parties.html).

There's no established best practice for storing shared credentials at GDS. Some teams have a "credentials" GitHub 
repository and use [pass](https://www.passwordstore.org/) to store credentials encrypted with GPG. Make sure you
configure the repo as [internal or ideally private](https://docs.github.com/en/github/creating-cloning-and-archiving-
repositories/creating-a-repository-on-github/about-repository-visibility)** to help prevent accidental disclosure.

Using GPG in this way isn't ideal because:

- We can't revoke access to credentials, unless we change them. Anyone who was given access can still decrypt
 credentials using their local copy of the repo and its commit history.

- It creates a high barrier to entry, especially for people less familiar with Unix tooling. We've found GPG tools are
 [generally difficult to use](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) and keyservers are unreliable.

Investigate alternatives before adopting GPG based credential stores for new teams.
```

source/standards/storing-credentials.html.md.erb
benthorner: Oops, I've amended with typo corrected.

Regarding your suggestion, I agree it's helpful to explain "why" at this point, but without implying third party products are "exploitable". How about this:

> ...the first instance. This is simpler than setting up an extra account with a third party and avoids some of the potential issues below.

source/standards/storing-credentials.html.md.erb
benthorner: Thanks @richardTowers. I agree we need to make a standard that reflects what we actually do, and strike a balance between instruction and information to let people decide for themselves.

I've added a new commit (e838bb834a512643f692d1cb63a1ae7d5752052b) with your suggestion, except the second sentence: it's unclear exactly what we're asking people to do.

How does it look now?

source/standards/storing-credentials.html.md.erb
benthorner: Both reasonable comments, thanks. I duplicated the warnings because they may evolve separately, which is happening already. I didn't want to end up with "as above" (when there is no longer an "above").

How about this?

> Use the secrets management feature of your infrastructure or cloud provider e.g. AWS Secrets Manager, HashiCorp Vault. This should make it easy to control and audit access to the credentials.
>
> Older (pre-cloud) projects in GDS tend to have a "secrets" repository and use GPG to encrypt files, which are then decrypted during the deployment process. However, using GPG [isn't ideal](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html).

It would be nice if we had some links to further guidance or examples of how we use cloud secrets management. I'm aware of [this blog](https://technology.blog.gov.uk/2021/07/23/automatically-creating-and-rotating-gov-uk-application-secrets/), but maybe the provider docs are a better starting point. What do you think?

source/standards/storing-credentials.html.md.erb
richardTowers: That sounds good to me 👍🏻

source/standards/storing-credentials.html.md.erb
richardTowers: Yeah, I'm not sure if there's a great place to link. That blog is nice, but it's getting into automating rotations and maybe that's a bit more advanced than most people reading this guidance need. I guess if you can find some AWS / Hashi docs which are high enough level we could link to those.

source/standards/storing-credentials.html.md.erb
richardTowers: Yep, that looks good to me.

source/standards/storing-credentials.html.md.erb
benthorner: Great, I've applied this in https://github.com/alphagov/gds-way/pull/633/commits/32e21f6a7f0b2e54d3fe6c57ef77237b6999e42b.

I'll resolve this now, but anyone else feel free to comment if you think there's more to change here.

benthorner:
Thanks for all your comments everyone, I think I've addressed them.

Would you all like to take another look?

richardTowers (approved):
This looks good to me now (might be worth leaving it open for a few more hours though, just in case).

Thanks very much for taking the time to write this up Ben, great work!

source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,60 @@
...
+
+## Personal credentials
+
+Examples include login details for:
+
+- GitHub
+- AWS
+- GOV.UK Signon
+
+[Use the password manager built into your browser in the first instance](https://lock.cmpxchg8b.com/passmgrs.html#conclusion). This is simpler than setting up an extra account with a third party and avoids some of the potential issues below.
```
karlbaker02, line 19: Can you elaborate on some of the potential issues referenced here regarding personal passwords for work - I may have missed it but I'm unsure which issues you're referring to? 

I'm also interested in whether we should be advocating folks to use the password manager built into your browser of choice over a third party password manager - I guess this will depend on the potential issues from the previous question.

risicle: The mechanisms used by third party password managers to integrate with a browser are shaky at best and have a history of being easily tricked by malicious web pages.

I personally advocate using a totally separate password manager with no browser integration as that's akin to building a big obvious bridge from my least trusted environment to my most trusted environment.

I'm sure Ben saved a copy of the original slack thread where we discussed all these things.

benthorner: @karlbaker02 sounds like we might need to rephrase this to make it clearer. The link in the instruction is goes to an article that explains the issue with using third party password managers when combined with a browser extension. We also say this below. How about rephrasing the following sentence to:

> This is [more secure than using a browser extension](https://lock.cmpxchg8b.com/passmgrs.html#conclusion) and you don't need to setup an extra third party account.

> unless synced with some external service, such as their Google account

That's what I do: my passwords are synced with my GDS Google account. Arguably syncing in general is risky if e.g. you're using Safari and syncing your passwords to some random iCloud account that doesn't have 2FA. We don't advocate for or against it. Do you think we need to say something explicitly?

source/standards/storing-credentials.html.md.erb
```diff
@@ -0,0 +1,60 @@
...
+
+If you can't do this for some reason, then it's OK to use a third-party password manager. This could be necessary if your browser has an accessibility issue, or if you work with multiple browsers.
+
+Some of the third-party password managers people use at GDS include:
```
benthorner, line 23: @risicle just following up on [your comment about QtPass](https://github.com/alphagov/gds-way/pull/633#issuecomment-897481484).

I had a quick look and it seems to be an offline, GPG-based approach. Do you think we should recommend other people use it here?

risicle: Well, I'd recommend it as much as any of the others.

risicle: (It also has git support to let you push your stores around)

source/standards/storing-credentials.html.md.erb
karlbaker02: I don't think I've read enough on the subject to be honest to say whether I agree that we should be defaulting to use the password managers in-built into browsers over third-party password managers wherever possible for personal work passwords, as well as definitively saying what is more secure (there's more attack surface potentially with an integration, but that doesn't mean the browsers won't do something stupid in the future / introduce a bug which exposes stored credentials etc). @risicle raises a good point about password managers which are completely independent of browser which one might argue are more secure given that there is no integration at all with your browser of choice.

benthorner: That's OK, I'm not an expert on this subject either. A fair few people have contributed to this though, which is reassuring. I agree browsers aren't invulnerable, but that's true of all software.

Specifically about your original comment and my reply: is there anything you'd like me to change?

source/standards/storing-credentials.html.md.erb
benthorner: Haha, OK. It sounds like an interesting choice and good to have one other offline option for balance (without turning this into a list of all password managers!).

I've added it in a new commit: https://github.com/alphagov/gds-way/pull/633/commits/4b53bb13befd0f7059791dc41157ccda5977b068

richardTowers (approved):

PR #638, opened 2021-08-27
Migrate GDS Way to Github Actions

tomrosier:
Migrate GDS-way to use github actions as it enables us to decommission concourse


0atman (approved):
Same here, go on and merge to test this and see what breaks :ok_hand:

.github/workflows/bundle_and_release.yml
```diff
@@ -0,0 +1,31 @@
...
+          sudo apt-get install -y ruby=1:2.7+1
+          sudo apt-get install -y nodejs=10.19.0~dfsg-3ubuntu1
+          sudo apt-get install -y cf7-cli
+          
+          sudo gem install bundler:1.17.2
+          
+          bundle install --without development
+          bundle exec middleman build
+          
+          cf login -a https://api.cloud.service.gov.uk -u ${{ secrets.CF_EMAIL }} -p ${{ secrets.CF_PASSWORD }} -o gds-tech-ops -s docs
```
AP-Hunt, line 29: The best way to do non-interactive authentication with CF is to set the `CF_USERNAME` and `CF_PASSWORD` env vars, then do `cf auth` in the script

paroxp (commented):
I'd probably break the job down into smaller chunks. Something like:

```
- update-system
- install dependencies
- run tests
- deploy artifact
```

Not only the division, but perhaps using different custom designed actions, so we don't rely on `apt-get` and versioning?

- https://github.com/actions/setup-node
- https://github.com/ruby/setup-ruby
- https://github.com/marketplace/actions/cf-cli-action

What do you think?

PR #646, opened 2021-09-28
Update browser support page

lfdebrux:
Updates person to contact for access to BrowserStack and bumps review date.

Currently the Head of Frontend is away, so in that case people should contact the Lead Developer for their programme. Also remove names to avoid having to update the docs whenever the person in the role changes.

PR #648, opened 2021-09-30
Update point of contact for Assistiv Labs access

36degrees:
Currently the Head of Frontend is away, so in that case people should contact a Lead Frontend Developer.

Mimics the changes made in #646 for the BrowserStack access.


lfdebrux (approved):

36degrees:
Thanks for the speedy review, @lfdebrux 🙌🏻

PR #659, opened 2021-11-17
Remove quarterly planning guidance

timblair:
GDS no longer consistently follows a quarterly schedule for its planning, and the current structure of the organisation is no longer set up to facilitate working in this ways (e.g. the Investment Board no longer exists, people moves don't happen between teams / directorates at quarterly boundaries, etc).

Given this reality, and some ongoing long-term operational model changes, it makes sense to remove this guidance entirely until there is something more standardised to document.


willp-bl (approved):
:shipit:

PR #663, opened 2021-12-23
Add to GitHub Actions best practise docs

36degrees:
Add guidance on:

- setting granular permissions for the GITHUB_TOKEN
- using environments to restrict access to secrets
- protecting the workflows folder using a CODEOWNERS file
- enforcing which actions are allowed to run in a repo

Most of this is based on what we're doing on repos owned by the Design System team – some of which is based on GitHub's blog post ['Implementing least privilege for secrets in GitHub Actions'][1].

Restructure a couple of bits to try and group thematically related points together.

[1]: https://github.blog/2021-04-13-implementing-least-privilege-for-secrets-in-github-actions/


katstevens (approved):
LGTM 👍

PR #671, opened 2022-01-24
JavaScript content check and minor example tweak

injms:
Checked the content for the JavaScript coding style page and bumped the review date.

Small change - split the naming convention example into two to make it clearer which example refers to which rule.


tombye (approved):
Looks good to me, though got to this late so date will need changing now :(

philandstuff:
Don't think the date matters too much - merging

PR #677, opened 2022-02-08
BAU: Use Docker for local site preview

mrwilson:
Sorting out ruby versions and installations can be a pain, so this re-uses a pattern common in Digital Identity/Verify to run the preview site in a Docker container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM ruby:2.6.6-buster
+
+EXPOSE 4567:4567
+EXPOSE 35729:35729
+
+WORKDIR /usr/src/gems
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ruby:2.6.6-buster@sha256:1512719e1172fbcaa913997ffcb4b2be0948038fe255040a29f819f6c54c328d
```
philandstuff, line 1: ISTR if you pin to a hash, that pins you to a specific architecture? We should check this works on both intel and ARM.

philandstuff (approved):

PR #679, opened 2022-02-16
WIP: Add manual on tagging AWS resources

philandstuff:
This is very much work-in-progress but I thought it'd be good to start
writing this.

### DIFF
--- a/source/manuals/aws-tagging.html.md.erb
+++ b/source/manuals/aws-tagging.html.md.erb
@@ -57,6 +57,7 @@ GOV.UK Sign In is using the following tags:
  
 This is based on:
 
+ - [AWS documentation and best practices on tagging](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html)
  - [GDS's former tagging strategy](https://docs.google.com/presentation/d/1LHLKPclfrn5KVFrFd2WqyPOYpS6wXklE4Lexb2rJNW0/edit#slide=id.g10d43e3636_2_51)
  - [MoJ digital tagging strategy](https://technical-guidance.service.justice.gov.uk/documentation/standards/documenting-infrastructure-owners.html)
  - [the duckbill group's guide to tagging best practices](https://www.duckbillgroup.com/blog/aws-cost-allocation-guide-tagging-best-practices/)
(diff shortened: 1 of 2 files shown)


mrwilson:
I'd probably add a link to https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html, as it has things like the tag name / tag value validation rules.

source/manuals/aws-tagging.html.md.erb
```diff
@@ -0,0 +1,62 @@
...
+ - `Product`: should be `GOV.UK Sign In`
+ - `System`: the name of the software system, for example `Authentication` or `Identity proofing and verification core`. Avoid abbreviations.
+ - `Environment`: should be one of `production`, `staging`, `integration`, or `development`.
+ - `Owner`: an email address for an owner for the resource. For dev environments, this will be an individual email address; elsewhere it will be a group address.
+
+### Optional
+
+ - `Service`: used to describe the function of a particular resource (for example: account management, session storage, front end)
+ - `Name`: a name for this particular resource. This should be unique within a deployment (terraform deployment, cloudformation stack, etc)
+ - `Source`: the URL(s) for any source code repositories related to this resource, comma separated
```
galund, line 54: Commas are allowed in URIs so can this be space-separated instead?

Stoods-hd:
2 potential things:
1) Do we add anything around whether code for a resource is "signed" - or do we assume everything must be signed and we can trace what is not?
2) Could we provide specific options for what is entered for certain values?
e.g. System must be IPV, TxMA, Auth, etc.  (Longhand obviously :) )
This allows us to also keep track of services that form part of DI

galund:
For (2) I think we should have that list as an RFC we keep up-to-date in https://github.com/alphagov/digital-identity-architecture.. (1) maybe also best if we do something DI-specific and then take what we learn to the rest of GDS?

source/manuals/aws-tagging.html.md.erb
padajo: I think "system" is ambiguous in this context. Tagging is useful for cost control and audit, so it's useful to identify both those elements. While "team" may change over time, the importance is who is responsible for cost control of an element of the overall solution, so... what does that mean in terms of "system"?

philandstuff:
@Stoods-hd for (1), I'm not sure what problem you're trying to solve?

for (2), we should probably have a list, but it maybe shouldn't live in the GDS way, since it's DI-specific. Maybe confluence?

willp-bl (approved):

PR #687, opened 2022-03-02
Update flake8 section

leohemsted:
it was a bit too verbose, massively trimmed down some example stuff. Realigned a bunch of headers to be under flake8 instead of as children of "linting". Also, added isort to the linting section.

Also did some of my own linting of the URL references section - removed a bunch of old unused links and grouped up the URLs by theme


benvand (approved):
LGTM 👍  
Addition of isort is good

PR #689, opened 2022-03-08
Update the date of review on this document

Nooshu:
We have a separate stream of work happening in the SM related to browser support, so this can be bumped.


alex-ju (approved):

philandstuff:
@Nooshu @alex-ju if this required no changes, is 4 months too frequent for review?

alex-ju:
We review browser usage every 6 months (and update the service manual accordingly, if required) – so I'd say 6 months would suit this better.

philandstuff:
@alex-ju can you raise a PR to update the review cadence?

PR #690, opened 2022-03-08
Update document review date

Nooshu:
I don't believe our stance on this has changed, so creating a patch to bump the date. If this isn't the case, raise it in the PR and we can make the larger changes required.


philandstuff:
@Nooshu @alex-ju  if this required no changes, is 3 months too frequent for review?

alex-ju:
The HTML spec doesn't change as often as we might think, arguably 1 year might be more suitable for this.

philandstuff:
@alex-ju ok can you create a PR to change the review cadence? (line 4 in the markdown file)

PR #694, opened 2022-03-15
Clean up NodeJS guidance

injms:
Minor formatting tweaks to the NodeJS at GDS guide - the main one being to make the code examples conform to StandardJS, as that's the convention that should be used when writing code.

Note - I haven't touched the section on whether NPM is the tool of choice. I believe that Yarn is used in certain parts of the organisation, but am unsure whether that's tied to Ruby on Rails or not. I think that's a can of worms that is best resolved in another pull request as it's a fairly major change.

### DIFF
--- a/source/manuals/programming-languages/nodejs/index.html.md.erb
+++ b/source/manuals/programming-languages/nodejs/index.html.md.erb
@@ -75,11 +75,11 @@ Use these files to `require` all the modules functions.
 
 ```js
 // product/index.js
-const product = require('./product');
+const product = require('./product')
 
 module.exports = {
   create: product.create
-};
+}
 ```
 
 > Store test files within the implementation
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -445,21 +438,20 @@ for (let i = 0; i < 10; i += 1) {
 }
 
 // Over:
-for (let i=0,j=0;i<10 && j<10;j++,i=(j==10)?i+1:i,j=(j==10)?j=0:j,console.log(i,j)){}
+for (let i = 0, j = 0; i < 10 && j < 10; j++, i = (j == 10) ? i + 1 : i, j = (j == 10) ? j = 0 : j, console.log(i, j)) {}
```
tombye, line 441: At a glance it seems a bit odd to add spacing to an example of code written to take up less space. Do you think it's needed here?

tombye (commented):
The majority of these changes look fine to me, I only made one comment and it's not a blocker if you feel the change it references should happen.

source/manuals/programming-languages/nodejs/index.html.md.erb
injms: Good point - I've removed this change.

source/manuals/programming-languages/nodejs/index.html.md.erb
```diff
@@ -1,6 +1,6 @@
 ---
 title: Using Node.js at GDS
-last_reviewed_on: 2021-07-02
+last_reviewed_on: 2021-03-15
```
querkmachine, line 3:
```suggestion
last_reviewed_on: 2022-03-15
```

injms: Good spot - thank you!

tombye (approved):
Thanks for doing the changes. Looks good to me now 👍🏻

PR #696, opened 2022-03-21
Add guidance on making commits atomic

chao-xian:
Most developers are very good at making their commits granular/atomic. But it takes practise and isn't currently our recommended way. It should be something that we strive for.


source/standards/source-code/working-with-git.html.md.erb
```diff
@@ -0,0 +1,196 @@
...
+
+ - [GitHub Actions](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/) supports a `$default-branch` macro
+ - [concourse-git-resource](https://github.com/concourse/git-resource): if you do not set `branch`, it will check out the default branch
+
+## Commits
+
+### Atomic commits
+
+Each commit should be atomic, meaning that each commit should only include changes that achieve a given, granular, step.
+They should also be logical steps that follow each other. As Anna Shipman describes it,
```
injms, line 37: Small suggestion - slight rewording to hopefully make it a bit clearer, and split it up over three bullet points to make it easier to scan:

```suggestion
Each commit should be atomic - this means that each commit:

* is self-contained
* only includes changes that achieve a specific step
* should be in a logical order

As Anna Shipman describes it,
```

injms:
I think atomic commits are very useful, especially in how it helps makes reviewing clearer when stepping forwards and backwards through the commits whilst still having a (hopefully) working branch. I added a small suggestion to make the description of what an atomic commit is a bit clearer.

chao-xian:
Thanks @injms!

Issue #700, opened 2022-03-31
Re: 'The GDS Way'


Aidan23Smith:
Problem with 'The GDS Way' (https://gds-way.cloudapps.digital/)

Hi,

Just had a look through the GDS Way and and realised that the [How to store source code](https://gds-way.cloudapps.digital/standards/source-code.html) link doesn't appear to work.

I hope this helps,
Aidan

philandstuff:
Ah I think this was unintentionally caused in #696 by @chao-xian

mrwilson:
Thanks for letting us know @Aidan23Smith 

There's prior art for link checking and other linting (using `html-proofer`) in other projects using the tech docs template (such as [the GOV.UK Sign-In technical documentation](https://github.com/alphagov/di-authentication-tech-docs/blob/main/config.rb#L12))

Maybe that's worth folding into the GDS Way repo too.

PR #701, opened 2022-03-31
BAU: Check internal and external links

mrwilson:
This PR introduces checking for internal links (within the GDS Way's own content) and external links (content on other sites).

We don't want to have broken links in the GDS Way as this is a bad and confusing user experience. This was highlighted in issue #700.

This PR builds on existing work in https://github.com/alphagov/tech-docs-template/pull/187 and would aim to be folded into the tech docs template eventually.

This PR is also a place for discussion around:
- updated content and new external links prompted by the link checker
- how to handle rate-limiting by GitHub when checking links to repos, of which there are many in the project


Gemfile.lock
```diff
@@ -1,23 +1,23 @@
 GEM
```
36degrees, line 1: There's quite a lot of dependencies being bumped here which seem unrelated to the commit they're in ('ignore GitHub links') – can we split them into a separate commit or PR?

source/partials/_nav-version-control-deployments.html.erb
```diff
@@ -1,5 +1,6 @@
 <ul>
   <li><a href="/standards/source-code/">How to store source code</a></li>
+  <li><a href="/standards/source-code/index.html">How to store source code</a></li>
```
36degrees, line 3: Isn't this a duplicate of the link above?

source/standards/optimise-frontend-perf.html.md.erb
```diff
@@ -96,7 +96,7 @@ You can find out more about improving your website’s frontend performance by r
 The Service Manual has more suggestions about [how you can test frontend performance][].
 
 [frontend performance]: https://www.gov.uk/service-manual/technology/how-to-test-frontend-performance
-[minification]: https://minifycode.com/
+[minification]: https://web.dev/reduce-network-payloads-using-text-compression/
```
36degrees, line 99: We've updated this URL but removed the link from minification on line 21. Is that intentional? Is this link referred to from anywhere else?

source/standards/principle-least-access.html.md.erb
```diff
@@ -43,6 +43,6 @@ If you’re using the gds-users account to log into your AWS accounts, you shoul
 
 - [NCSC - 10 steps to cyber security](https://www.ncsc.gov.uk/collection/10-steps)
 - [NCSC - privileged user management](https://www.ncsc.gov.uk/guidance/introduction-identity-and-access-management#section_6)
-- [NIST Special Publication 800-53 - AC-6 least privilege][polp]
+- [NCSC - preventing lateral movement][polp]
 
-[polp]: https://csrc.nist.gov/Projects/risk-management/sp800-53-controls/release-search#!/control?version=5.1&number=AC-6
```
36degrees, line 48: I don't have the context here but this link appears to resolve fine. If we're changing the link just because we think the NCSC guidance is better or more appropriate, should we split this change out as it's not broken?

36degrees (commented):
Thanks for picking this up, @mrwilson! A few things I'm not 100% sure about – might just be because I'm missing context.

My Ruby is rusty at best, so you might want a second set of eyes on some of this, depending on your confidence in the changes.

Gemfile.lock
mrwilson: Yep, we can be more specific with this, I "just" copied the working Gemfile.lock in order to get to green,.

source/partials/_nav-version-control-deployments.html.erb
mrwilson: Yep, will remove

source/standards/optimise-frontend-perf.html.md.erb
mrwilson: This was a mistake, I've added the link back

source/standards/principle-least-access.html.md.erb
mrwilson: Yep, I changed this in an early draft to get it working and didn't change it back, reverted in next change.

domoscargin:
How much do we care about hashes if the target page at least loads?

For instance, we're not catching the fact that this link:
https://github.com/alphagov/gds-way/blob/db16b146f1469d7538497f30845e0a4f5704edf1/source/manuals/programming-languages/html.html.md.erb#L15 has updated to https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support

Looks like we could enable `check_external_hash` (and `check_internal_hash`?) to catch these.

mrwilson:
@domoscargin That's a very good shout.

According to the HTMLProofer docs, the default value for both of those flags is true, however I will add them to the configuration for the sake of explicitness and not relying on default behaviour that may change.

richardTowers (approved):
Looks good to me!

Only thing that I think will be slightly annoying is the inevitable "I just want to fix this typo, but to get the build working I need to fix these unrelated broken links". But that's probably better than having broken links, and I don't think an alternative cadence (e.g. daily check with a slack message, doesn't fail the build) would work given the limited ownership of GDS way.

PR #695, opened 2022-03-16
CDIO Cyber Security Vulnerability Management Service

jonmichaelpitt:
This PR is for the creation of the Vulnerability Management Service information page, providing details of the service and how to onboard to it. This will act as a central point for providing CO departments with further information.

Author: jonmichaelpitt <jon-michael.pitt@digital.cabinet-office.gov.uk>

PR #711, opened 2022-05-26
Remove the CDIO Pillars page

timblair:
This page is now just information about the structure of CDIO, rather than specific guidance relevant to the building or operating of services.

The information on the Cyber team is likely useful, but is better placed somewhere else. For now, any references to contacting the Cyber team have been replaced with a link to their Slack channel. This is also somewhat related to #695 which is more of a product page than guidance.


timblair:
@gabrielcurrie I'd be interested in your thoughts about this. There's a reasonable amount of Cyber-related content in the GDS Way now, but a fair amount of it is not guidance about how to do a thing, but more information about the services that Cyber offers, or the structure of the team. It feels like there should be a better place for that information than here.

gabrielcurrie:
Hi @timblair I agree with this, and we are in the process of developing an internally-facing Google Site to host this kind of information about the team. Happy to remove from the GDS Way.

PR #712, opened 2022-07-11
Bump review date and increase review cadence for style guide page

injms:
Reviewed style guide content - nothing needs updating, so bumping the last reviewed date.

### DIFF
--- a/source/manuals/programming-languages.html.md.erb
+++ b/source/manuals/programming-languages.html.md.erb
@@ -1,8 +1,8 @@
 ---
 layout: core
 title: Style guides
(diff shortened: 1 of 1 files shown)


philandstuff (dismissed):
is it worth bumping the `review_in`? I don't think this needs reviewing every 6 months, maybe 12 instead?

I had a look at this and the only thing I raised an eyebrow at was the `editorconfig` - just because I'm not sure many people are using it in practice. @Nooshu added it, I don't think it does any harm, so I'm happy to leave it as-is.

philandstuff (approved):

PR #717, opened 2022-07-20
Remove GOV.UK PaaS from hosting recommendations

kr8n3r:
As PaaS is going away, we should stop recommending it.


lfdebrux (approved):

pauldougan:
I added @louzoid-gds who has an interest in this too

pauldougan:
i think this is very much a first step and that in time will expand with some more specific guidance on which of the many aws hosting options you should consider. Apprunner, Bean Stalk, ECS, EKS, Fargate ...

PR #722, opened 2022-07-26
Update data stores standard

philandstuff:
I removed the section which contained general-purpose advice for
choosing a data store.  I didn't find the advice to be terribly
actionable, nor terribly specific to GDS's particular technology
environment.

Reducing the content in the GDS way makes it easier to review in future.


louzoid-gds (commented):
Thanks Phil. I agree with your reasoning and would go a step further and suggest this whole page isn't very actionable. The point about using managed services is a good one but really applies more broadly than to just DBs.
What'd you think about deleting this page?

philandstuff:
@louzoid-gds I agree - I've updated the PR to delete the whole page

louzoid-gds (commented):
Thanks Phil! Ok to also remove from the nav?

philandstuff:
@louzoid-gds thanks, done

louzoid-gds (approved):
Thanks Phil!

PR #730, opened 2022-08-19
Update Understanding the risks Page

MargauxGir:
Added link to the Cyber Security Team’s Threat Modelling service page, as well as information on the benefits of threat modelling. The rest of the content is accurate.


source/standards/understanding-risks.html.md.erb
```diff
@@ -19,7 +21,16 @@ The Service Manual has some recommendations which can reduce risk to your servic
 
 [Modelling threats][] can help you gain a clearer understanding of threats against your service. GDS uses [Attack Tree][] development workshops to model threats. Any workshops you run should cover all potential [attack vectors][].
```
louzoid-gds, line 22: I think we should perhaps cross-link here to the GDS Way page on threat modelling instead of the OWASP page

louzoid-gds (commented):
Thanks for reviewing! I've added a minor comment on cross-linking to the threat modelling page

louzoid-gds (approved):
Looks good to me

PR #732, opened 2022-08-22
Update accessibility.html.md.erb

Nooshu:
Bumped the last update date and next review in January, hopefully After more WCAG 2.2 details are released


injms (approved):
Makes sense to review in time for the proposed release date for WCAG 2.2 👍

PR #734, opened 2022-08-25
Fix broken links and make capitalisation consistent

injms:
Reviewed contents, fixed a couple of links that had changed, and updated some capitalisation to be more consistent.


deanwilson (approved):
Looks good. Merge away

PR #738, opened 2022-09-06
Update principle-least-access.html.md.erb

MargauxGir:
Content is still accurate, deletion of a sentence as it is no longer accurate.


louzoid-gds (changes requested):
Changes look good. If you're happy with the content, can you update the last_reviewed_on date at the top of the page? Thank you!

PR #739, opened 2022-09-08
Add a page on data stores / data sources

mammykins:
* Revive old data store page but change emphasis
* Suggest that analytics should be baked into a service
* Provide contact slack channel for data services to advise
* Point to the data audit and how data stores need named owners


mammykins:
I couldn't build locally, so was unable to view a live preview!

louzoid-gds:
Closing the PR with agreement from Matthew. We're exploring other options with the information services area teams

PR #748, opened 2022-09-26
Fix link and bump review date

deanwilson:
I think that's the best match for what the link intends to cover but it is a guess on my part


sihugh (approved):
Looks like an appropriate link 👍

PR #757, opened 2022-09-26
Add an extra step and bump the review date

deanwilson:
The guidance here is generic but an OK starting point


louzoid-gds (approved):
LGTM

PR #759, opened 2022-09-26
Reducing review date so we can decide on prometheus beta

deanwilson:
Is it still in Beta? Who's using it now?


mammykins (approved):
Review cadence adjusted, now every 3 months. I assume this will change back to 6 when decision on prometheus beta occurs?

Clarification to a sentence.

PR #775, opened 2022-12-08
Update dependancies as ubuntu-latest is now jammy

erino:

PR #777, opened 2023-01-03
Use GitHub Actions to install Node and Ruby for deploys

lfdebrux:
Changes deploy workflow to install Ruby and Node.js using the officially recommended actions [[1], [2]], instead of from the Ubuntu repos. This PR does this by copying the approach used by alphagov/govuk-design-system [[3]], where the test workflow is used by the deploy workflow to create a build artifact which is then deployed. This means the versions of Ruby and Node.js used are defined in one fewer place, and the only code in the deploy pipeline that is unique to doing a deploy is the code which does the deploy.

I suggest doing this as an alternative to #775, if we want to continue using Ruby 2.7.

[1]: https://github.com/ruby/setup-ruby
[2]: https://github.com/actions/setup-node
[3]: https://github.com/alphagov/govuk-design-system/tree/7c9e6f6353ab755ba1276899a0d86799cfc092b1/.github/workflows


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,65 @@
...
+            });
+            const mainBranchCommitSha = response.data.sha
+            if (context.sha !== mainBranchCommitSha) {
+              core.info(`Workflow run GITHUB_SHA ${context.sha} does not match main branch commit sha ${mainBranchCommitSha}`)
+              core.error('Code to deploy does not match code in main branch')
+              core.setFailed('Refusing to deploy, please do not re-run this workflow run')
+            }
+
+      - name: Download build artifact
+        uses: actions/download-artifact@v3
```
LeePorte, line 42: I'd be tempted to pin this to the SHA rather than the version `actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741` in this case as the version tag can be replaced, but the SHA can't

.github/workflows/test.yaml
```diff
@@ -0,0 +1,41 @@
...
+      upload-artifact:
+        default: false
+        required: false
+        type: boolean
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
LeePorte, line 16: I'd also go for the SHA here

.github/workflows/test.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
```
LeePorte, line 19: I'd also go for the SHA here

.github/workflows/test.yaml
```diff
@@ -0,0 +1,41 @@
...
+      with:
+        ruby-version: '2.7'
+        bundler-cache: true
+
+    - name: Set up Node
+      uses: actions/setup-node@v1
```
LeePorte, line 25: I'd also go for the SHA here

LeePorte (commented):
Pipeline looks fine, though I would advise pinning to git SHAs As a best practice approach.

LeePorte (dismissed):

LeePorte (approved):

PR #792, opened 2023-03-01
Add instructions for making EventMachine gem work on Apple silicon

alexbishop1:
To run locally on an Apple silicon Mac (such as one with an M1 chip), the EventMachine gem needs to built with OpenSSL 1.1 from Homebrew.

Also change how we set the Bundler path to stop using a deprecated method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,12 @@ to update the GDS Way.
 
 To preview or build the website, we need to use the terminal.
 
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):

owenblacker (approved):

dj-maisy (approved):
Looks legit. Do I have the power to approve? 👍

PR #800, opened 2023-03-22
BAU: Update program name

huwd:
Couple of program clarifications,

- Use the name our program wants to use.
- Slightly nuance the line about our use of serverless tech for clarity.

### DIFF
--- a/source/manuals/aws-tagging.html.md.erb
+++ b/source/manuals/aws-tagging.html.md.erb
@@ -36,10 +36,11 @@ Currently, we do not enforce tags.
 
 In future, we may wish to consider mechanisms such as alerting on untagged resources, or automatically deleting untagged resources.
 
+
 ### Mandatory
 
- - `Product`: for example `GOV.UK` or `DSP`
- - `System`: the name of the software system (for example `Authentication` or `Identity proofing and verification core`. Avoid abbreviations)
+ - `Product`: `GOV.UK One Login` / `GOV.UK` or `DSP`
+ - `System`: the name of the software system, for example `Authentication` or `Identity proofing and verification core`. Avoid abbreviations.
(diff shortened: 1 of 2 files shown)


source/manuals/aws-tagging.html.md.erb
```diff
@@ -36,13 +36,13 @@ Currently, we do not enforce tags.
...
-## Tags used in GOV.UK Sign In
+## Tags used in GOV.UK One Login
 
-GOV.UK Sign In is using the following tags:
+GOV.UK One Login is using the following tags:
 
 ### Mandatory
 
- - `Product`: should be `GOV.UK Sign In`
+ - `Product`: should be `GOV.UK One Login`
```
huwd, line 45: @paroxp, could I confirm if we merged that, that I'm not going to break any reports we have.

I think now because the tags I've already seen are a bit of a mess of different names already?

And I know we're about to transition to our own org too?

paroxp: If we were to break any reports, which there may be little of, we shouldn't be blocked on iterating on our changes :) 

I think we can take the heat, and fix/adjust the reports later if that turns to be an issue.

huwd: @deanwilson gonna treat that as a "happy to go" from my program

huwd: @paroxp would you mind giving this a ✅ ?

paroxp (dismissed):

huwd:
sorry @paroxp forgot about this one. Just rebased against main, you happy to ✅ and I'll send it out this time?

timblair (approved):
Approving based on previous approvals.

PR #818, opened 2023-06-09
Update ruby to 3.1.3 and refresh Gemfile.lock

stephengrier:
A `bundle install` currently fails on arm64 devices with `ld: symbol(s) not found for architecture arm64` errors when installing version 1.2.7 of the eventmachine gem. This could possibly be fixed by upgrading the evenmachine gem, but there are a bunch of other gems that are out of date so let's take the opportunity to bring them all up-to-date.

Also, ruby 2.x has been end of life since March 2023 so update to the most recent version of ruby available for the `ubuntu-22.04` GitHub Actions runner, which is 3.1.3.

Delete `ruby-version` from the GitHub Build and test workflow so that it takes the ruby version from the `.ruby-version` file.


lfdebrux (approved):
Looks good to me!

louzoid-gds (approved):

PR #821, opened 2023-06-28
Update base image

LeePorte:


LeePorte:
Should be opened against fork

PR #827, opened 2023-08-24
Updating Git branching/merging conventions

mrwilson:
This change refreshes the guidance on git branching and merging to be less prescriptive and more descriptive of options.

Commits:
- Remove hard line wraps.
- Update branching and merging conventions
- Rebase vs Merge guidance

### DIFF
--- a/source/standards/source-code/working-with-git.html.md.erb
+++ b/source/standards/source-code/working-with-git.html.md.erb
@@ -142,37 +142,20 @@ commands like git merge and git revert.
 
 ## Branching/merging conventions
 
-You may often choose to work on a particular feature on a "feature branch"
-rather than directly on `main`. Indeed, given how cheap branches are in Git,
-[using branches](http://git-scm.com/book/en/Git-Branching-Basic-Branching-and-Merging) is positively encouraged.
+You may often choose to work on a particular feature on a "feature branch" rather than directly on `main`. Indeed, given how cheap branches are in Git, [using branches](http://git-scm.com/book/en/Git-Branching-Basic-Branching-and-Merging) is positively encouraged.
 
-You are encouraged to make liberal use of Git's [history rewriting
-features](http://git-scm.com/book/en/Git-Tools-Rewriting-History) while working
-locally, in order to arrange your commits into appropriate logical chunks that
-will make sense to your fellow developers. In particular, you may find
-`git rebase --interactive` very useful. You are also encouraged to avoid merge
-commits and use `git rebase main` instead. However, you should not rewrite commits
-that have been pushed unless you:
+You are encouraged to make liberal use of Git's [history rewriting features](http://git-scm.com/book/en/Git-Tools-Rewriting-History) while working locally, in order to arrange your commits into appropriate logical chunks that will make sense to your fellow developers. In particular, you may find `git rebase --interactive` very useful. You are also encouraged to avoid merge commits and use `git rebase main` instead.
 
-  * are very sure that no-one else will be affected by you rewriting the
-    branch history
-  * have an extremely good reason. For example: someone has committed
-    sensitive information (personally identifiable data, passwords and suchlike)
-    and it needs purging from history
+However, you should not rewrite commits that have been pushed unless you:
 
-When in doubt you should err towards smaller commits, which can be rebased
-together later. It's harder to break large commits out into smaller chunks.
+  * are very sure that no-one else will be affected by you rewriting the branch history
+  * have an extremely good reason. For example: someone has committed sensitive information (personally identifiable data, passwords and suchlike) and it needs purging from history
 
-The smaller commits should still be logical chunks, but this will give context
-for a more specific change and make git tools like `annotate` and `log` more
-useful.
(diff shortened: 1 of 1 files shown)


source/standards/source-code/working-with-git.html.md.erb
```diff
@@ -142,37 +142,24 @@ commands like git merge and git revert.
...
 
-You may often choose to work on a particular feature on a "feature branch"
-rather than directly on `main`. Indeed, given how cheap branches are in Git,
-[using branches](http://git-scm.com/book/en/Git-Branching-Basic-Branching-and-Merging) is positively encouraged.
-
-You are encouraged to make liberal use of Git's [history rewriting
-features](http://git-scm.com/book/en/Git-Tools-Rewriting-History) while working
-locally, in order to arrange your commits into appropriate logical chunks that
-will make sense to your fellow developers. In particular, you may find
-`git rebase --interactive` very useful. You are also encouraged to avoid merge
```
BeckaL, line 153: I think here the advice to avoid merge commits when updating your feature branch with the lastest main is still good advice and worth keeping, and kind of separate to the 'do you merge or rebase when merging into main' debate.

BeckaL: Maybe it can be clarified to something like

"If you are working on a feature branch and main gets updated, you may need to update your feature branch with main. In this case, it's useful to `git rebase main` rather than creating a merge commit on your feature branch, since this keeps the history cleaner."

BeckaL (commented):
Think this is a nice clean rewrite, thank you! Just a comment about preserving one bit of the existing advice which I think is still good practice.

mrwilson:
> Think this is a nice clean rewrite, thank you! Just a comment about preserving one bit of the existing advice which I think is still good practice.

Yep, in hindsight I agree, I'll pull it back in and tweak it a bit.

BeckaL (dismissed):
LGTM!

source/standards/source-code/working-with-git.html.md.erb
```diff
@@ -142,37 +142,28 @@ commands like git merge and git revert.
...
-branch), in particular one that has previously been shared with colleagues, you
-should use `git merge`'s `--no-ff` option to preserve evidence of your feature
-branch in the repository history. This advice may be freely ignored for smaller
-local feature branches for which a fast-forward merge will look like any other
-routine development work on `main`.
+You should develop on a short-lived "feature" [branch](https://git-scm.com/book/en/Git-Branching-Basic-Branching-and-Merging) rather than directly on the trunk branch, which is usually `main` or `master`, unless there is a compelling reason to do otherwise.
+
+For example if your repository has many binary files, the `git diff` will not provide useful change information, and you may choose to collaborate directly on the trunk branch.
+
+You should prefer staging your commits as small chunks rather than having a single commit per pull-request.
```
sengi, line 149: I think there's a lot of scope for people (particularly those who are new to working with VCS/DVCS) to misinterpret "small chunks". I think we can be a bit more specific and accurate while conveying more about the underlying intent:

```suggestion
You should prefer organising your commits as small, self-contained changes rather than having a single commit per pull-request. Ideally, each commit should represent a coherent set of changes.
```

sengi (approved):
Thanks for improving this!

sengi (approved):
Awesome, thanks!

mrwilson:
@sengi I've absorbed your change and added some extra detail about each commit still passing tests etc

source/standards/source-code/working-with-git.html.md.erb
```diff
@@ -142,37 +142,28 @@ commands like git merge and git revert.
...
+This will make your changes easier for other people to understand them at review time.
+
+You may need to update your feature branch with the latest changes from the trunk branch when you have multiple people contributing to the same repository.
+In this case, you should use [`git rebase`](https://docs.github.com/en/get-started/using-git/about-git-rebase) with the trunk branch name.
+This applies your local changes on top of the most recent changes to the trunk branch, rather than creating a merge commit on your feature branch, which keeps the history linear and makes it easier to review.
+
+Git allows you to [rewrite local changes](https://git-scm.com/book/en/Git-Tools-Rewriting-History) by combining commits, splitting commits apart, changing commit messages and many more operations.
+
+You should not rewrite commits that have been merged or pushed to the trunk branch unless there is a compelling reason such as committing personally identifiable information or application secrets.
```
chao-xian, line 158: I thought that such commits still exist unless we ask GitHub to scrub it entirely from the reflog for us anyway?

sengi: I believe there's a garbage collection process for orphaned commits, so they'll be deleted eventually. So if it's, say, a key that's already been revoked then we wouldn't need to ask them to expunge anything, but if it's PII then we most likely would.

sengi: (I'm not suggesting adding those details here btw — just giving context for why I think the sentence is accurate as-is.)

sengi (approved):

mrwilson:
I'm going to merge this as-is, but I think we should raise an additional PR with new content about "What do I do if I've published secrets/PII to GitHub?"

PR #831, opened 2023-09-19
Fix link redirect in java.html.md.erb

dav-idc:
Swapped out a link with its new version.


DavidBiddle (approved):
@dav-idc Change looks good to me. I think the test failure was just one of the other links on the page being flaky - if you rerun the action I think it should pass.

timblair:
Fixed in #842.

PR #837, opened 2023-10-25
Add a new test manual

AgaDufrat:
Add a new manual to check if the newish External links check is preventing from adding new docs


huwd:
@AgaDufrat do you think this is still a live change?
or could we close this down and come back to the branch when there's more updates?

PR #840, opened 2023-11-01
Bump browser-support page review date

DavidBiddle:
Bumps the review date for this page. Also extends the `review_in` date to 12 months - there isn't much content on this page and the content that is here just links elsewhere, so it's unlikely to go out of date.


domoscargin (approved):
Sounds about right!

PR #844, opened 2023-11-02
Make HTML Proofer ignore links to new files

robinjam:
Each page produced during the build contains a `link rel="canonical"` tag and `twitter:url` and `og:url` meta tags all pointing to the page itself. This means that HTML Proofer fails on all newly created files within feature branches, because the newly created files do not yet exist in production. See #835 for an example of this issue.

This PR causes HTML Proofer to ignore links to newly-created pages. The newly created pages will still have their own external links tested, as before.

(n.b. `delete_suffix` is a no-op if the string does not end with that suffix)

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -9,6 +9,15 @@ set :layout, 'custom'
 
 after_build do |builder|
   begin
+    # Newly created files on feature branches do not yet exist in production, so tell HTML Proofer to ignore links to them
+    # Based on https://github.com/gjtorikian/html-proofer#ignoring-new-files
+    merge_base = `git merge-base origin/main HEAD`
+    new_urls = %x(git diff -z --name-only --diff-filter=AC #{merge_base})
+      .split("\0")
+      .filter { |path| path.start_with? "source/" }
+      .map { |path| path.delete_prefix("source/").delete_suffix(".erb").delete_suffix(".md") }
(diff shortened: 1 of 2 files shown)


config.rb
```diff
@@ -9,6 +9,15 @@
 
 after_build do |builder|
   begin
+    # Newly created files on feature branches do not yet exist in production, so tell HTML Proofer to ignore links to them
+    # Based on https://github.com/gjtorikian/html-proofer#ignoring-new-files
+    merge_base = `git merge-base main HEAD`
+    new_urls = %x(git diff -z --name-only --diff-filter=AC #{merge_base})
+      .split("\0")
+      .filter { |path| path.start_with? "source/" }
+      .map { |path| path.delete_prefix("source/").delete_suffix(".erb").delete_suffix(".md") }
```
AgaDufrat, line 18: Could we do
```suggestion
      .map { |path| path.delete_prefix("source/").delete_suffix(".erb.md") }
```

robinjam: `delete_suffix(".erb").delete_suffix(".md")` correctly handles both `"foo.md.erb"` and `"foo.md"` (as well as e.g. `"foo.txt.erb"`)

ChrisBAshton (approved):
LGTM. Thanks for fixing this ⭐

PR #848, opened 2023-11-02
Skip external link check during CI on `main` branch

robinjam:
This commit splits building and testing external links into two separate actions, and skips the link checker for builds on the main branch.

This ensures that the main branch is always in a deployable state, even if it contains broken links (for example, if it contains new files that have self-referential links within them, as is the state of this repo right now).

Note that CI is expected to fail for this PR. The issue should go away after the next successful deploy, in the meantime we will need to merge this PR bypassing the status check.


ChrisBAshton (approved):
Nice 👍

PR #849, opened 2023-11-15
Update operating-systems.html.md.erb

Jonathan-Scott14:
I have reviewed the Hardening section and added references to the CAF and to the CIS Benchmarks.

A tech lead should review the other content.

### DIFF
--- a/source/standards/operating-systems.html.md.erb
+++ b/source/standards/operating-systems.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Operating systems for virtual machines
-last_reviewed_on: 2022-07-14
+last_reviewed_on: 2023-11-15
 review_in: 12 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/standards/operating-systems.html.md.erb
```diff
@@ -26,9 +26,9 @@ Standard releases only receive security updates for 9 months after their initial
 
 ## Hardening
 
-You must apply hardening to your operating system. For example, you should disable root SSH access and deny SSH access using a password.
+You must apply hardening to your operating system. The NCSC [Cyber Assessment Framework] has a specific objective for this (B4.b).
```
lfdebrux, line 29: Lower down the page we also refer to the NCSC, but we expand the acronym and then define it:

https://github.com/alphagov/gds-way/blob/88be4a67e8609b0d1056c7a6a75da2515d1adae0/source/standards/operating-systems.html.md.erb#L37

Maybe we should do that in this new line, and change the existing line to just use the acronym.

huwd (approved):

PR #851, opened 2023-11-15
Update logging.html.md.erb

Jonathan-Scott14:
Removed reference to Logit. Updated reference for IBM report. Added reference to the CAF. Added reminder on ensuring secure storage.

Removed associated Logit references.

I cannot comment on the use of DropWizard or Cloud Foundry.

### DIFF
--- a/source/standards/logging.html.md.erb
+++ b/source/standards/logging.html.md.erb
@@ -14,6 +14,10 @@ Use [Splunk] to store and query infrastructure, application and audit logs.
 Splunk is a cloud-based SaaS tool for short and long-term storage,
 visualisation, alerting, and reporting.
 
+Your product should have a proportionate design for short and long term storage of logs and ensuring the Confidentiality, Integrity, and Availability of logs.
+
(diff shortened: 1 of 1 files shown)


source/standards/logging.html.md.erb
```diff
@@ -14,16 +14,9 @@ Use [Splunk] to store and query infrastructure, application and audit logs.
 Splunk is a cloud-based SaaS tool for short and long-term storage,
 visualisation, alerting, and reporting.
 
-### Logit deprecation notice
+Your product should have a proportionate design for short and long term storage of logs and ensuring the Confidentiality, Integrity, and Availability of logs.
 
-The shared GDS [Logit] account can still be used for existing environments;
-however, you should use Splunk for new logging requirements.
```
stephengrier, line 20: Hi @Jonathan-Scott14 I don't think we can remove references to Logit yet because it is still used by GOV.UK, Notify and PaaS (though obviously the PaaS itself is going away) and they are not likely to move to Splunk.

Jonathan-Scott14: I've resolved your comments :D

huwd (approved):

PR #852, opened 2023-11-15
Update secrets-auditing.html.md.erb

Jonathan-Scott14:
tried to correct ACL link


huwd:
Ah looks like there's now a merge conflict 🤔 
You happy to rebase against main and amend?

PR #857, opened 2023-11-21
Update security-overview-for-websites.html.md.erb

Jonathan-Scott14:
Reviewed by T Bradley.
1. Encryption could be further elucidated.
2. MFA could be further elucidated.
3. DR and BC could be included.

JS added info on MFA, context-aware access etc. Other changes would require discussion with tech leads (e.g. does 15 step become... 17 step?)

### DIFF
--- a/source/manuals/security-overview-for-websites.html.md.erb
+++ b/source/manuals/security-overview-for-websites.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Security overview for websites
-last_reviewed_on: 2022-02-23
+last_reviewed_on: 2023-11-21
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manuals/security-overview-for-websites.html.md.erb
```diff
@@ -16,6 +16,12 @@ Use named accounts to log in for all systems (from registrars to content managem
 Use Single Sign-On (SSO) wherever possible to link accounts to Google using OAuth or OIDC.
 If the service does not support SSO, make sure you use a gov.uk email and enable multi-factor authentication (MFA), ideally using a phishing-resistant mechanism like WebAuthn (hardware keys).
 
+If applicable, use more secure forms of MFA for higher risk accounts. For example, hardware tokens such as Yubikeys for privileged accounts, app-based authentication for standard accounts, and SMS MFA for low risk accounts.
```
huwd, line 19: Do we have an opinion on passkeys as a 2nd factor?
I note github recently enabled them?

huwd (approved):
One thought, but could always be an addition later. Otherwise LGTM

source/manuals/security-overview-for-websites.html.md.erb
Jonathan-Scott14: I like passkeys as 2nd factor, though they generally seem to be tied to an authenticator, hardware token or password manager according to [this](https://docs.github.com/en/authentication/authenticating-with-a-passkey/about-passkeys). I'd apply the same logic as per above on how passkeys are actually implemented - so a passkey on a hardware token being preferable over an authenticator app for example.

PR #858, opened 2023-11-21
Update threat-modelling.html.md.erb

Jonathan-Scott14:
T Bradley reviewed and suggested examples would be helpful.

JS added example based on Pay and contrasted with GOV.UK.


huwd (approved):
Addition of an example seems helpful

PR #860, opened 2023-11-21
Update secrets-acl.html.md.erb

Jonathan-Scott14:
Reviewed by A Dion.


huwd (approved):
Syntactically correct, content seems good from the perspective of a dev and was properly reviewed according to the PR by security folk.

PR #868, opened 2024-02-08
Expand github action pinning guidance to include update approach

huwd:
SHA pinning is a sensible approach to mitigate potential supply chain attacks. See some great blog posts here on the approach:

https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash also
https://michaelheap.com/improve-your-github-actions-security/#using-pin-github-action

However one down side is that SHA's are not very human readable. It can be difficult to tell from the SHA if the version we've pinned has an update, or if that update is a security or important fix.

Best practice therefore is to place a comment after the pinned version listing the semantic version for a third party github action.

This gets you best of both worlds, maintainability plus certainty.

It might look at bit like this:

```
jobs:
  check-pull-request:
    runs-on: ubuntu-latest
    steps:
      - name: Check out repository code
	uses: actions/checkout@ee066bloop # pin @v2
      - name: Install Ruby uses:
	ruby/setup-ruby@22acsewblah # pin@v1
```

Consistency here also helps us manage this code in line with the GDS Way requirement to Update dependencies frequently when managing third party dependencies:
https://gds-way.digital.cabinet-office.gov.uk/standards/tracking-dependencies.html#update-dependencies-frequently

Since October 2022 DependaBot will now look for comments on SHA pinning and automatically suggest updates. Similar approaches may be possible for other dependency management tools.

Dependabot currently supports a range of different comment syntaxses which can be viewed here:
https://github.com/dependabot/dependabot-core/pull/5951#issuecomment-1298147084

I've tried to keep the guidance general and open, leaving detail to this commit history, given the range of different tools on use across GDS.

The principles are:

- Pin your actions using SHAs
- Ensure human readability by commenting the semver on the line with the action
- Explore if your usual dependency management process, especially automated ones like DependaBot can help flag and raise visibility on new versions.

## Looks like

image
image

### DIFF
--- a/source/standards/source-code/use-github.html.md.erb
+++ b/source/standards/source-code/use-github.html.md.erb
@@ -55,13 +55,24 @@ Consider protecting the `.github/workflows` folder by using [a CODEOWNERS file](
 
 Consider creating a Workflow Template in the [alphagov workflow folder](https://github.com/alphagov/.github/tree/main/workflow-templates) if you need to share a similar workflow between many repositories.
 
-[Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible. If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions).
+[Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible.
+
+If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions) and [configure Dependabot to keep your actions up to date](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions) by adding a comment on the same line with the tag the commitsha represents. For example:
+
+```
+- uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 #v2.1.0
+```
+
+Pinned versions should include the semver version in a comment next to the SHA, helping humans understand which versions we are pinned to.
+Where possible, allow automated dependency management tools to scan these version comments and suggest updates.
+
 Third-party actions should only be used if:
 
 - The provider is verified by GitHub (for example, [aws-actions](https://github.com/aws-actions))
 - The action is complex enough that you cannot write your own local action
 - You have fully reviewed the code in the version of the third-party action you will be using
 - You have pinned the specific version in your workflow and in the repository settings, using a Git commit SHA
+- You have included the semver version in a comment next to the SHA, helping humans understand the version and automated tools report on what is out of date (for example dependabot)
 - The third-party action is actively maintained, well-documented and tested ([follow the guidance on third party dependencies](/standards/tracking-dependencies.html)).
 
 You can enforce this in the settings for Actions by choosing ['Allow select actions'](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-specific-actions-to-run) and then 'Allow actions created by GitHub' and 'Allow Marketplace actions by verified creators' as required.
@@ -90,7 +101,9 @@ You should use your `@digital.cabinet-office.gov.uk` email during the sign up pr
 
 * [How to store source code](index.html)
 * [Working with Git](working-with-git.html)
+* [Updating actions with DependaBot][github-dependabot-actions]
 
 [GitHub]: https://technology.blog.gov.uk/2016/05/31/how-we-use-git-at-the-government-digital-service/
 [alphagov]: https://github.com/alphagov/
 [govuk-one-login]: https://github.com/govuk-one-login
+[github-dependabot-actions]: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


source/standards/source-code/use-github.html.md.erb
```diff
@@ -52,12 +52,16 @@ Consider protecting the `.github/workflows` folder by using [a CODEOWNERS file](
 Consider creating a Workflow Template in the [alphagov workflow folder](https://github.com/alphagov/.github/tree/main/workflow-templates) if you need to share a similar workflow between many repositories.
 
 [Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible. If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions).
+Pinned versions should include the semver version in a comment next to the SHA, helping humans understand which versions we are pinned to.
+Where possible, allow automated dependency management tools to scan these versions and suggest updates.
```
willp-bl, line 56: `versions` -> `version comments`

Worth adding a link to the comment formats that are supported too

source/standards/source-code/use-github.html.md.erb
```diff
@@ -52,12 +52,16 @@ Consider protecting the `.github/workflows` folder by using [a CODEOWNERS file](
...
+
 Third-party actions should only be used if:
 
 - The provider is verified by GitHub (for example, [aws-actions](https://github.com/aws-actions))
 - The action is complex enough that you cannot write your own local action
 - You have fully reviewed the code in the version of the third-party action you will be using
-- You have pinned the specific version in your workflow and in the repository settings, using a Git commit SHA
+- You have pinned the specific version in your workflow and in the repository settings
```
willp-bl, line 63: best keep this line the same so it is clear that a SHA should be used

willp-bl (changes requested):
Looks good just needs a couple of minor changes.

source/standards/source-code/use-github.html.md.erb
```diff
@@ -52,12 +52,16 @@ Consider protecting the `.github/workflows` folder by using [a CODEOWNERS file](
 Consider creating a Workflow Template in the [alphagov workflow folder](https://github.com/alphagov/.github/tree/main/workflow-templates) if you need to share a similar workflow between many repositories.
 
 [Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible. If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions).
+Pinned versions should include the semver version in a comment next to the SHA, helping humans understand which versions we are pinned to.
+Where possible, allow automated dependency management tools to scan these version comments and suggest updates.
```
huwd, line 56:
```suggestion
- You have pinned the specific version in your workflow and in the repository settings, using a Git commit SHA, and include the version tag as a comment.
- You have dependabot configured to ensure that any security or functional updates to those actions are kept up to date (e.g. deprecation of language version) 
```

source/standards/source-code/use-github.html.md.erb
```diff
@@ -52,12 +52,16 @@ Consider protecting the `.github/workflows` folder by using [a CODEOWNERS file](
 Consider creating a Workflow Template in the [alphagov workflow folder](https://github.com/alphagov/.github/tree/main/workflow-templates) if you need to share a similar workflow between many repositories.
 
 [Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible. If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions).
```
huwd, line 54:
```suggestion
[Create your own local actions](https://docs.github.com/en/actions/creating-actions/about-actions) wherever possible.

If using GitHub-owned actions, [pin to a specific version](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions) and [configure Dependabot to keep your actions up to date](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions) by adding a comment on the same line with the tag the commitsha represents. For example:

`- uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 #v2.1.0`
```

huwd: These are @andyloughran's changes from https://github.com/alphagov/gds-way/pull/874/files

Posting here so we can discuss what's worth pulling in?

source/standards/source-code/use-github.html.md.erb
huwd: These are @andyloughran's changes from https://github.com/alphagov/gds-way/pull/874/files

Posting here so we can discuss what's worth pulling in?

source/standards/source-code/use-github.html.md.erb
huwd: I quite like the code snippet, I was on the fence about putting one in, but decided to try and keep it all trimmer in a chunky section. What'd you think? Worth doing anyway?

Other thoughts:
- I think we should be clear we want version pinning on _any_ action? Potentially even ones we make ourselves no? So the distinction isn't between github actions / third party or internal?
- I _really_ want to talk about DependaBot here, it's soo useful in this context, but then erred, the dependency requirements in the [GDS way are more open](https://gds-way.digital.cabinet-office.gov.uk/standards/tracking-dependencies.html#update-dependencies-frequently). We reference DependaBot, but also allow for alternatives Snyk and some kinda python milarky. So wanted to outline it as a principle without calling out the impliemntation? I left more specific advice in the commit message though... perhaps that's too buried 🤔. I'd be interested in thoughts from likes of @timblair / @willp-bl  on that one, they've got a better birds-eye-view across GDS than me.

source/standards/source-code/use-github.html.md.erb
huwd: Aside from the specific mention of dependabot, I think we're saying the same thing here.

Open to others to weigh in on which reads better. I'm not notorious for my clear writing style.

source/standards/source-code/use-github.html.md.erb
willp-bl: Giving a specific example of how it works and noting that you don't have to use that specific tool/service should be ok. It will help explain what it is and you could link to the relevant guidance about dependencies

I can't think of a reason why we wouldn't have it on every action, even those we may make ourselves.

huwd: Cool, I'll come back with an amendment here that nicks a bit more of Andy's code.

source/standards/source-code/use-github.html.md.erb
andyloughran: I think I prefer yours Huw; in the cold light of day 'repository settings' doesn't mean much..

source/standards/source-code/use-github.html.md.erb
andyloughran: Yeah, the only bit that is missing is that the Dependabot test fixture allows for many different types of version comment; whereas we're providing a single way.  There's no documentation outside of the fixture that I could find that explains what's supported or not.

huwd: Cool, I'll allude to that

stephengrier:
This seems like a good change to me and is inline with what we're already doing in some places. I'm happy to approve when the comments are resolved.

huwd:
Rightio folks,
Finally got this done, ready for a re-review if one of @andyloughran, @stephengrier, @willp-bl wouldn't mind another look?

heathd (approved):
I think this has been reviewed by several people, the review comments have been acted on, so I'm happy for this to be merged in. Thanks Huw this looks like some valuable advice

PR #870, opened 2024-02-08
Attempt to fix deployments by tweaking github actions

huwd:
I'm suspicious that some of our actions aren't acting like they should.

I've updated the deploy artifact action,
Also added a suggested permission fix in from their [README](https://github.com/actions/upload-pages-artifact?tab=readme-ov-file#example-permissions-fix-for-linux) (my current main theory here, see commit for why).

Whilst I was there updated others.

It's bundling a few things, but can drop commits if folks disagree.
Would be nice to work with a clean slate of steps that mirror their readmes.


alexbishop1 (approved):
🤞

DavidBiddle (approved):

PR #871, opened 2024-02-08
Remove deploy

huwd:
1. Removes deploy script which targets PaaS
2. Ensures we're still using build and test ahead of deploying to pages.


.github/workflows/deploy-to-pages.yaml
```diff
@@ -14,6 +14,10 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  build:
+    uses: ./.github/workflows/test.yaml
+    with:
+      upload-artifact: true
   deploy:
```
willp-bl, line 21: There's a `needs: build` in the deleted yaml that might be needed?

alexbishop1 (approved):
Looks reasonable

alexbishop1 (approved):

DavidBiddle (approved):

PR #872, opened 2024-02-09
Update github actions within test.yml

huwd:
As I've upgraded the other ones may as well do these too.

The specific thing we want to get in here however is that we have _one_ external action here `ruby/setup-ruby`.

When we'd upgraded the `deploy-to-pages.yml` I'd made a point of simplifying the action permissions.
All `action/*` actions are created by github and are allowlisted.

The _only_ third party action that is now allow listed is:
`ruby/setup-ruby@22fdc77bf4148f810455b226c90fb81b5cbc00a7 # v1.171.0`

Having done a scan of the repo I can't find anywhere else we're using `ruby/setup-ruby@*`

Ideally I still don't like these limit rules being pinned to SHAs within the action permissions.
I'd much prefer `ruby/setup-ruby@v1` then have to update that for a major bump...

But I note we've always been dealing with a setup-ruby@v1 here, that rule has been in the allowlist and we've still been seeing failures.

So I'll learn how the action pinning works in permissions later... let's try and get this through.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,7 +37,7 @@ jobs:
     # Share data between the build and deploy jobs so we don't need to run `npm run build` again on deploy
     # Upload the deploy folder as an artifact so it can be downloaded and used in the deploy job
     - name: Upload artifact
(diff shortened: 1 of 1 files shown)


DavidBiddle (approved):
Looks sensible to me

PR #874, opened 2024-02-09
Updated the documentation on Github Actions to ensure pinned dependencies are kept updated.

andyloughran:
Dependabot has the capability to keep Github Actions updated, even when pinned to a specific commitsha rather than tag.

There are many formats that are allowed, but I've only provided a single example in the docs: https://github.com/dependabot/dependabot-core/blob/b4112ce4639d7eed1e3b2e0792eb7533f7cb125f/github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml#L7-L30 

There doesn't appear to be good end-user facing documentation produced for dependabot on this feature, just the discussions on two PRs:

- https://github.com/dependabot/dependabot-core/pull/5951 (where they added the capability to query comments in dependabot for github actions)
https://github.com/dependabot/dependabot-core/issues/4691 (to auto-update the comment if the commitsha is updated)

Thanks to @huwd for the headsup on this one.


andyloughran:
Closed in favour of https://github.com/alphagov/gds-way/pull/868

PR #875, opened 2024-02-13
update linting rules

leohemsted:
main points are:

* we use black to dictate formatting
* we use ruff as it's much faster than flake8
* ruff covers isort too so we can squash that

### DIFF
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -18,10 +18,12 @@ and consistent, within, and across, projects at GDS.
 
 ## Code formatting
 
-We follow [PEP 8][]; where PEP 8 does not express a view (for example, on the usage of
+We use [Black][] to format our code. Black is an opinionated formatter that follows
+[PEP 8][]; where Black and PEP 8 do not express a view (for example, on the usage of
 language features such as metaclasses) we defer to the [Google Python style guide][GPSG].
-Use these as references unless something is explicitly mentioned here. These rules should be followed in conjunction with the advice on consistency
-on the main [programming languages manual page][gds-way-code-style-guides].
+Use these as references unless something is explicitly mentioned here. These rules
+should be followed in conjunction with the advice on consistency on the main
+[programming languages manual page][gds-way-code-style-guides].
 
 If you want to add a new rule or exception please create a pull request against this repo.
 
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -35,119 +37,81 @@ Couple this with the fact that much of the time GDS developers are coding web ap
...
-    models.Address.street_address_line_1
-    == user['address']['street_address_line_1']
-):
-    pass
-
-
-Do
-
-if not models.Address.query(models.Address.street_address_line_1 == user['address']['street_address_line_1']):
-    pass
```
leohemsted, line 51: i removed this block because if you have `black` enabled, then that takes all the opinion out of which of these is better (in this case it uses the second one). i think it's still nice to include the paragraphs above as justification for _why_ we set line length to 120

louzoid-gds:
Thanks Leo! Just a reminder to bump the review date as well.

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -35,119 +37,81 @@ Couple this with the fact that much of the time GDS developers are coding web ap
...
-any would be particularly beneficial to your code base.
-
-Examples include checks for requiring copyright/licensing strings, requiring docstrings
-or warnings for upcoming deprecations.
 
-A list can be found [by performing a PyPI search.][flake8-plugins]
+#### Ruff ignores
 
-#### Flake8 per file ignores
+Ruff can ignore particular lines or files. You can
```
CrystalPea, line 59:
```suggestion
Ruff can ignore particular lines or files.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -35,119 +37,81 @@ Couple this with the fact that much of the time GDS developers are coding web ap
...
 
-A particularly useful feature of Flake8 is the ability to specify rule
-exemptions on a per directory, per file, or per regex match basis.
+A particularly useful feature of ruff is the ability to specify rule exemptions on a
```
CrystalPea, line 61:
```suggestion
A particularly useful feature of ruff is the ability to specify rule exemptions
```

CrystalPea (commented):
Good work - I spotted a few artefacts left after editing content, left suggestions:

CrystalPea (approved):

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -35,119 +37,81 @@ Couple this with the fact that much of the time GDS developers are coding web ap
...
 ## Linting
 
 
-### Flake8
-
-This manual advises the use of the [Flake8][] command line checker as an all in one lint, codestyle and complexity checker.
+### Ruff
 
-#### How to use Flake8
+This manual advises the use of the [Ruff][] command line checker as an all in one lint,
```
louzoid-gds, line 46: Just spotted you're missing a reference link here. Ruff also linked to on line 73

leohemsted: Good spot! fixed and pushed. wonder if theres any md linting 🤔

louzoid-gds (approved):
Giving an approving tick as it's already been reviewed by Pea and internally advertised on GDS Slack. Looking good - thanks Leo

PR #877, opened 2024-02-21
Update DNS records guidance

louzoid-gds:
Removed references to tools that are no longer used in GDS. 
The line also mentions using version control to publish records. Using version control is already mentioned on the page so not sure this adds much to already common practice... Reviewers - shout if you disagree :)


DavidBiddle (approved):
Agree that the version control point is already covered, and happy to take your word for it on GOV.UK DNS and octoDNS no longer being used in GDS!

PR #884, opened 2024-04-08
ENG-184 Make it clear that users in alphagov must have a GDS email address

heathd:
the statement here was a bit vague. We are now tightening up the enforcement of this policy, and so I'm amending the wording here.

I've slightly restructured the text as there was a mixture of 'advice about configuring a repo' and 'how to set up your user account and gain access'. I've created separate headings for those two topics

**I intend to merge this on 15 April if there are no objections**

[Current page](https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/use-github.html#content)

new wording shown in screenshot

Screenshot 2024-04-08 at 13 31 55

### DIFF
--- a/source/standards/source-code/use-github.html.md.erb
+++ b/source/standards/source-code/use-github.html.md.erb
@@ -6,17 +6,21 @@ review_in: 6 months
 
 # <%= current_page.data.title %>
 
-[GDS uses GitHub as its code repository](https://technology.blog.gov.uk/2016/05/31/how-we-use-git-at-the-government-digital-service/).
+GDS uses the [alphagov] organisation on [GitHub] to collaborate on code. The GOV.UK One Login programme uses [govuk-one-login].
 
-Put new repositories for GDS services in the [alphagov](https://github.com/alphagov/) organisation on GitHub.
+### Getting access to `alphagov`
 
-You can use your personal GitHub account to access alphagov but please ensure that you also link your GDS email address to your account. Ask your tech lead or technical architect to invite you. GDS will revoke your access to alphagov when you leave GDS.
+You can use your personal GitHub account to access alphagov if you wish.
 
-To secure your Github repository, make sure you:
+All github user accounts added to `alphagov` *must* be [connected](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/adding-an-email-address-to-your-github-account) with a valid `digital.cabinet-office.gov.uk` email address. Accounts not connected to a valid email address will be removed from `alphagov`.
 
-- configure two-factor authentication for your account
-- have considered [encrypting some of the repository's contents](/standards/storing-credentials.html#team-credentials)
-- consider [protecting your main branch](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) to prevent changes being committed without a suitable review
+You must also set up two-factor authentication on your account.
+
+To join `alphagov` ask your tech lead or technical architect to invite you. Make sure you've connected your GDS email address to your account first, otherwise your account will be removed.
+
+### Configuring github repositories
+
+Consider [protecting your main branch](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) to prevent changes being committed without a suitable review.
 
 You can also consider backing up your Git repositories to another location (this should be a team responsibility). If you are using AWS to host your service
 [AWS CodeCommit](https://aws.amazon.com/codecommit/) is one option.
@@ -86,3 +90,7 @@ You should use your `@digital.cabinet-office.gov.uk` email during the sign up pr
 
 * [How to store source code](index.html)
 * [Working with Git](working-with-git.html)
+
+[GitHub]: https://technology.blog.gov.uk/2016/05/31/how-we-use-git-at-the-government-digital-service/
+[alphagov]: https://github.com/alphagov/
+[govuk-one-login]: https://github.com/govuk-one-login
--- a/source/standards/source-code/use-github.html.md.erb
+++ b/source/standards/source-code/use-github.html.md.erb
@@ -18,7 +18,7 @@ You must also set up two-factor authentication on your account.
 
 To join `alphagov` ask your tech lead or technical architect to invite you. Make sure you've connected your GDS email address to your account first, otherwise your account will be removed.
 
-### Configuring github repositories
+### Configuring GitHub repositories
 
 Consider [protecting your main branch](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) to prevent changes being committed without a suitable review.
 


source/standards/source-code/use-github.html.md.erb
```diff
@@ -6,17 +6,21 @@ review_in: 6 months
...
+You can use your personal GitHub account to access alphagov if you wish.
 
-To secure your Github repository, make sure you:
+All github user accounts added to `alphagov` *must* be [connected](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/adding-an-email-address-to-your-github-account) with a valid `digital.cabinet-office.gov.uk` email address. Accounts not connected to a valid email address will be removed from `alphagov`.
 
-- configure two-factor authentication for your account
-- have considered [encrypting some of the repository's contents](/standards/storing-credentials.html#team-credentials)
-- consider [protecting your main branch](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) to prevent changes being committed without a suitable review
+You must also set up two-factor authentication on your account.
```
paroxp, line 17: This is enforced by GitHub themselves, right? As in, a toggle on the organisation will redirect people to setting up 2FA.

Is it worth to test and mention that behaviour?

timblair: Yep, this is enforced at the Enterprise level.

source/standards/source-code/use-github.html.md.erb
```diff
@@ -6,17 +6,21 @@ review_in: 6 months
...
+
+To join `alphagov` ask your tech lead or technical architect to invite you. Make sure you've connected your GDS email address to your account first, otherwise your account will be removed.
+
+### Configuring github repositories
```
timblair, line 21:
```suggestion
### Configuring GitHub repositories
```

source/standards/source-code/use-github.html.md.erb
```diff
@@ -6,17 +6,21 @@ review_in: 6 months
 
 # <%= current_page.data.title %>
 
-[GDS uses GitHub as its code repository](https://technology.blog.gov.uk/2016/05/31/how-we-use-git-at-the-government-digital-service/).
+GDS uses the [alphagov] organisation on [GitHub] to collaborate on code. The GOV.UK One Login programme uses [govuk-one-login].
 
-Put new repositories for GDS services in the [alphagov](https://github.com/alphagov/) organisation on GitHub.
+### Getting access to `alphagov`
```
timblair, line 11: This leads me to questions like "what if I need access to `govuk-one-login`? Are there differences in terms of this guidance between `alphagov` and `govuk-one-login` and, if there are, why, and what are those difference? IMO we should have general guidance for the use of GitHub across GDS (regardless of org), and then separate sections for when something might differ (e.g. the process for requesting access).

source/standards/source-code/use-github.html.md.erb
```diff
@@ -6,17 +6,21 @@ review_in: 6 months
...
 
-You can use your personal GitHub account to access alphagov but please ensure that you also link your GDS email address to your account. Ask your tech lead or technical architect to invite you. GDS will revoke your access to alphagov when you leave GDS.
+You can use your personal GitHub account to access alphagov if you wish.
 
-To secure your Github repository, make sure you:
+All github user accounts added to `alphagov` *must* be [connected](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/adding-an-email-address-to-your-github-account) with a valid `digital.cabinet-office.gov.uk` email address. Accounts not connected to a valid email address will be removed from `alphagov`.
```
timblair, line 15: As above, is this requirement only for `alphagov`, or also for `govuk-one-login`?

(Separate point: it's GitHub, not github.)

source/standards/source-code/use-github.html.md.erb
heathd: I actually don't know what the policy is about access to govuk-one-login org .. is it documented anywhere Rafal?

source/standards/source-code/use-github.html.md.erb
heathd: What Tim said, do you think I should make that explicit in the text?

source/standards/source-code/use-github.html.md.erb
galund: I'd ask @paroxp or someone for confirmation, but I believe we have the same rule, tho whether or not it's enforced I don't know.

It's unfortunate that the GitHub terraform thing we use attaches users via username rather than via email address, cos that would be a much more obvious initial control point I'd have thought.

(Or does it?? That's how we use it, but it looks like the [GitHub API does at least accept email addresses instead](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#create-an-organization-invitation--parameters)?)

source/standards/source-code/use-github.html.md.erb
paroxp: I don't know - don't mind :) 

I guess this is for the users, so - would someone be confused by not having access to repository they've allegedly been onboarded to for not having 2FA? I honestly don't know :D 

I think what you said, is good enough of a foundation.

source/standards/source-code/use-github.html.md.erb
paroxp: Got something [documented in our Team Manual](https://github.com/govuk-one-login/team-manual/blob/main/source/new-starter-guide/github-access-management.html.md.erb) (happy to share the hosted page URL over Slack).

The things to point out that are different:
* We require SC for normal access
* If SC is not held, risk is recorded and accepted, resulting in individuals being assigned as outside collaborators
* We use IaC Terraform to manage the state of

source/standards/source-code/use-github.html.md.erb
paroxp: I don't see how this requirement needs to differ. Having that requirement helps us track who does indeed need the access, and who should be rolled off.

To @galund's point - I believe we can setup invitations by email, and the users will be able to then decide how they want to log in. But I am pessimistic, if the email is then associated/recorded anywhere. Effectively meaning once people are invited, we lose sight of who they are.

> (Separate point: it's GitHub, not github.)

++

galund: Yeah I think doing everything (but especially initial invite) by email helps, because then you know at least at the start they'll have been set up right. And then _if_ we do offboarding correctly, well then we know that we've not got anyone in the org that shouldn't be.

(I still think the automated check is a good idea because people might get invited in various ways, but if we start the right way it seems more likely we won't have to fix people up later.)

source/standards/source-code/use-github.html.md.erb
heathd: Rafal if you think we should link to guidance for govuk-one-login, please would you propose some suitable wording? 

I would like to get this change merged, and since the previous guidance didn't say anything about govuk-one-login, I have a slight preference for picking that up in a separate PR if that would be acceptable to others here?

heathd:
I'm keen to get this merged, there's some missing guidance about using the govuk-one-login org but since that was absent in the previous incarnation of this page, I'd like to propose that we merge this change as-is and raise another PR to add guidance about govuk-one-login.

Does anyone have any objections to that?

paroxp (approved):

PR #886, opened 2024-04-22
Bump to Ruby 3.2.3

huwd:
Note ruby v3.0.0 is currently out but there's been some reported problems running it within Apple Silicone due to some changes at the C library level.

For now doing a cautious upgrade until we've confirmed we're unblocked on the mac / docker / ruby combo.

See: https://github.com/govuk-one-login/authentication-tech-docs/pull/165


huwd:
redcarpet is a dependency within the govuk_tech_docs 😭 really [have to do an update over there at some point soon](https://github.com/alphagov/tech-docs-gem/blob/main/govuk_tech_docs.gemspec#L50)...

huwd:
I wonder if the time has come to bite the bullet over here...
https://github.com/alphagov/tech-docs-gem/pull/348

Might do a round of dep updates?

DavidBiddle:
@huwd looks like tech-docs-gem has updated redcarpet, doesn that change anything for this PR?

DavidBiddle (approved):
Seems to build locally without any problems now so I'm happy to approve

PR #892, opened 2024-04-24
Make commit messages "imperative"

### LINKED ISSUE #882: Absurdly strict grammatical point
We say:

>  The Git convention is to write these in the present tense

https://github.com/alphagov/gds-way/blob/d19d71e7f97703290f8a1275d8ca0d0a02307b7d/source/standards/source-code/working-with-git.html.md.erb#L110

But tense is irrelevant: commit messages should be written in the _imperative mood_.  That is: they should be commands which bring the repository to a desired state.

The example we give is:

> Leverage best-of-breed synergies going forward

which is obviously in the present tense. But it could be re-written as:

> Leverages best-of-breed synergies going forward

and _still_ be in the present tense.

### What is to be done?

Just as we append "by zombies" to a clause to reveal the passive voice, we can vouch safe the imperative mood by prepending a silent "please" thus:

> [please] Leverage best-of-breed synergies going forward

cjfryer:
It's ambiguous to say that commit messages should be in the present tense.  Both of the following conform to that rule, but the former is preferred:

- "Leverage best-of-breed synergies going forward"
- "Leverages best-of-breed synergies going forward"

We want the "imperative mood" here.  I've added an aide memoire about prepending a silent "please".

Fixes #882

### DIFF
--- a/source/standards/source-code/working-with-git.html.md.erb
+++ b/source/standards/source-code/working-with-git.html.md.erb
@@ -107,10 +107,16 @@ ticket is publicly viewable (and likely to remain so).
 Commit messages should start with a one-line summary no longer than 50
 characters. Various Git tools (including GitHub) use this as the commit
 summary, so you should format it like an email subject, with a leading capital
-and no full stop. The Git convention is to write these in the present tense.
+and no full stop. The Git convention is to write these in the imperative mood,
+as if you are issuing a command to the repository.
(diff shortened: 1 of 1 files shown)


louzoid-gds (dismissed):
A helpful revision imo

louzoid-gds (approved):

PR #898, opened 2024-05-09
Update accessibility page

### LINKED ISSUE #834: Improve alt text guidance in accessibility.html.md.erb
From @selfthinker on PR #828:
> When the alt-text section was changed in #500 it was to align with the GOV.UK content guidance. But The GDS Way is not about GOV.UK content specifically (although it can, and maybe should, be included) but about products from all of GDS. And the guidance in here currently does not align with the Design System guidance.

I think this is a valid point and one we should address. My view is that it might make sense to make this section much smaller (probably keeping the bit about `figcaption`) and link to the [Design System guidance on alt text](https://design-system.service.gov.uk/styles/images/#alt-text). It's a good introduction to using alt text, it explains most of the same things we're saying here, and it's not overly prescriptive, so GOV.UK and other teams can have their own more specific guidance if they want to.

I wanted to keep the changes in #828 limited so didn't make this change in that PR - capturing in an issue and assigning to myself so it still gets done.

DavidBiddle:
Includes 2 fairly minor updates to the 'Building accessible services' page:

- updating references to WCAG 2.1 -> 2.2
- simplifying the alt text section and linking to the Design System alt text guidance (closes #834)

### DIFF
--- a/source/manuals/accessibility.html.md.erb
+++ b/source/manuals/accessibility.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Building accessible services
-last_reviewed_on: 2023-10-13
+last_reviewed_on: 2024-05-09
 review_in: 6 months
 owner_slack: "#accessibility-community"
 ---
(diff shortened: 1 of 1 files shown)


source/manuals/accessibility.html.md.erb
```diff
@@ -11,13 +11,13 @@ Government services must be accessible to everyone. This includes anyone with a
 
 Building a service with accessibility in mind not only allows those with access needs to use your service, it also improves the service for everyone else. An accessibility problem with a website can be something that affects everyone, not just people who can only access the web with a keyboard or screen reader.
 
-Government services are legally required to be accessible. This means services must comply with the international WCAG 2.1 AA accessibility standard. This requirement has applied to all new and existing public sector websites since 2020, and to native applications since June 2021.
+Government services are legally required to be accessible. This means services must comply with the international WCAG 2.2 AA accessibility standard. This requirement applies to all new and existing public sector websites and native applications.
```
andrewhick, line 14:
```suggestion
Government services are legally required to be accessible. This means services must comply with the international WCAG 2.2 AA accessibility standard. This requirement applies to all new and existing public sector websites and mobile applications.
```

andrewhick: One minor suggestion to reflect the wording in the [accessibility regulations](https://www.legislation.gov.uk/uksi/2018/952).

andrewhick (approved):
Thanks, all looks good to me 👍

alexbishop1 (approved):
This is all looks reasonable.

PR #899, opened 2024-05-09
Update frontend performance

DavidBiddle:
Makes some updates to the FE performance page, as suggested by the frontend community. Main changes:

- Some updates to style and wording on existing recommendations in the table
- Removal of a couple of suggestions we don't think are relevant any more
- Addition of new sections on HTTP2+ and code splitting


source/standards/optimise-frontend-perf.html.md.erb
```diff
@@ -11,27 +11,25 @@ You should focus on [frontend performance][] when developing your service’s we
...
+|          | Investigate [lazy loading][]                                                     | For pages with many images, only load images in the immediate browser viewport.                                                                                                                                                                                |
+|          | Investigate the impact of loading [@font-face][] assets on perceived performance | Use the CSS `font-display` property or other [font-loading strategies][] to manage common issues like [FOUT, FOIT and FOFT][].                                                                                                                                 |
+|          | Minimise HTTP requests                                                           | Minimise the number of CSS and JavaScript files to reduce size and loading time. See ‘Code splitting’ below.                                                                                                                                                   |
```
tombye, line 27: This could read as minimising the size of the requests but the action sounds more like you're talking about the number of them. The reason given also talks about their size. I wonder if we should instead talk about reducing the number of round-trips to the server needed? There has generally been consensus that modern networks (inc mobile) can shunt around enough data that the latency overhead of a HTTP request tends to have more of an impact on performance than download size + file parsing (though that's a massive generalisation).

tombye (approved):
I added a comment but it seems like it also could just be my opinion so approving as everything else makes total sense and looks good to go 👍🏻

Thanks for doing the hard work of meeting and coming up with changes!

source/standards/optimise-frontend-perf.html.md.erb
DavidBiddle: Good point, thanks Tom - would something like this make sense do you think?

```suggestion
|          | Minimise HTTP requests                                                           | Minimise the number of CSS and JavaScript files to reduce the number of round-trips to the server. See ‘Code splitting’ below.                                                                                                                                                   |
```

tombye: Perfect 👌

DavidBiddle:
Sorry @tombye, looks like your review got dismissed when I updated the bit about server round-trips - would you mind re-reviewing if you've got time?

tombye (approved):
Looking great, thanks again for doing the work 👍🏻

Issue #908, opened 2024-06-24
Re: 'Understand the risks to your service'


BrettEvansUKHSA:
Problem with 'Understand the risks to your service' (https://gds-way.digital.cabinet-office.gov.uk/standards/understanding-risks.html)

The link to https://sites.google.com/a/digital.cabinet-office.gov.uk/gds/directorates-and-groups/cto-and-ciso-office/information-security is now leading to a 404

galund:
Hi Brett, thanks for raising this, it sparked some useful discussion.

These links are to our internal wiki. This is OK, because the GDS Way is primarily a resource for people within GDS.

We're considering whether to make it more obvious when a link is only expected to be accessible internally.

cheers
George

BrettEvansUKHSA:
Cheers wasn't entirely obvious was internal, in the end I went around the houses and followed NCSC guidance on clasification of risks.

PR #909, opened 2024-06-24
Warn where links are only accessible internally

galund:
- use `::before` as `::after` does not work where links are inside a word/phrase such as in `/standards/understanding-risks.html#understand-the-risks-to-your-service`
- avoid text-decoration underline propagation per https://stackoverflow.com/questions/8820286/how-to-remove-only-underline-from-abefore
    
https://github.com/alphagov/gds-way/issues/908

### DIFF
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -1,2 +1,10 @@
 @import "govuk_tech_docs";
 @import "modules/page-banner";
+
+a[href^="https://sites.google.com/a/digital.cabinet-office.gov.uk/"]::before,
+a[href^="https://sites.google.com/cabinetoffice.gov.uk/"]::before,
+a[href^="https://gds.slack.com/"]::before
+{
+  content: "\1F512\00a0";  // padlock
+  display: inline-block;  // avoid text-decoration propagation
+}


source/stylesheets/screen.css.scss
```diff
@@ -1,2 +1,9 @@
 @import "govuk_tech_docs";
 @import "modules/page-banner";
+
+a[href^="https://sites.google.com/a/digital.cabinet-office.gov.uk/"]::after,
+a[href^="https://sites.google.com/cabinetoffice.gov.uk/"]::after,
+a[href^="https://gds.slack.com/"]::after
+{
+  content: " (GDS staff only)";
```
cjfryer, line 8: As discussed on Slack: this CSS appends a non-breaking space and a [Unicode padlock](https://unicodeplus.com/U+1F512) to the links in scope of the rule.

```suggestion
  content: " "\00a0\01F512";
```

Tested in this shonky bit of HTML:

```
<html>
<head>
<style type="text/css">
a::after {
    content: "\00a0\01F512";
}
</style>
</head>
<body>
<a href="">Link</a>
</body>
</html>
```

galund: 😅 ecb7c30e4f28ab1802c4475d0ef89e8ba04d5703

DavidBiddle (approved):
I think this is a neat solution 🎉 

Interesting note: you've timed this PR well. Until very recently using the `::before` and `::after` pseudo elements for meaningful content was considered a failure of WCAG 1.3.1 - [Technique F87: Failure of Success Criterion 1.3.1 due to inserting non-decorative content by using ::before and ::after pseudo-elements and the 'content' property in CSS](https://www.w3.org/WAI/WCAG22/Techniques/failures/F87.html). 

This was removed from the draft spec last month (https://github.com/w3c/wcag/pull/3818) and appears to have been removed from the published version of the [Understanding SC 1.3.1 page](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships) last week.

galund:
tbf the screen reader I tried does not read the padlock. However, having considered what my previous version of this sounded like when read out, actually I think it would be more of an accessibility fail to try and incorporate this. The availability or otherwise of the link is accessible to everyone by following the link, so this feels like a very good case for progressive enhancement anyway.

PR #920, opened 2024-07-02
GDS branding

galund:
Branding - add logo and change font

- to make more clear that this is by *and for* GDS

References:

GDS brand guidelines: https://drive.google.com/file/d/1UU0i_cudRpDdf_qRfgCIRB9YYrFkdGVI/view

### DIFF
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -9,3 +9,16 @@ a[href^="https://gds.slack.com/"]::before
   content: "\1F512\00a0";  // padlock
   display: inline-block;  // avoid text-decoration propagation
 }
+
+.govuk-header__logo::before {
+  content: "GDS";
+  font-family: "Helvetica neue", arial, sans-serif;
+  font-size: 1.2em;
+  background: #2F88C9;
+  color: white;
+  padding: 0.4em 0.7em;
+  margin-right: 0.8em;
+  font-weight: normal;
+  letter-spacing: -0.02em;
+  float: left;
+}


DaveLee-GDS:
Looks nice and clean to me.

galund:
@cjfryer the quote from the page I linked is:

> While you can use the patterns, your site or service must not:
...
> * use the GDS Transport typeface

:-)

erino (approved):

cjfryer:
@galund yes, and that section talks (vaguely) about services that aren't "considered part of GOV.UK."

An equally strict interpretation of the rules would prevent GOV.UK One Login from using the Design System, since it's on `account.gov.uk`

All academic, I suspect, since the PR is approved 😁

huwd:
@galund I like the GDS logo helps situate this well and perhaps distinguish from something like the service manual.

However I'm also not sure about that call on the font, I get that the quote from service manual seems clear... but in practice I'd understood it came down to site intent + if you were on a *.gov.uk domain... 🤔 might ask the senior GDS hive mind there.

I do find transport a bit clearer to read, so would like to keep it if we can.

timblair:
> An equally strict interpretation of the rules would prevent GOV.UK One Login from using the Design System, since it's on account.gov.uk

For use of the Design System, this is a constraint that we have placed on ourselves, that we are free to change. For New Transport has [a licence agreement that restricts its legal use to a limited number of domains](https://designnotes.blog.gov.uk/2015/03/11/can-i-use-the-gov-uk-fonts/): `[www.]gov.uk`, `*.service.gov.uk`, and `*.blog.gov.uk`.

timblair (changes requested):
I'm happy with the logo addition (I think it's a useful addition to the clarity of the scope of the GDS Way), but I think we should park the GDS Transport change.

timblair:
Actually, the `service_name` config entry is also used for (at least) the page `<title>`, so we shouldn't be adding arbitrary HTML to it. So 👍 for the idea, 👎 for the implementation.

erino (changes requested):
The title of the page Safari's history is `The GDS Way - <span class="gds-logo">GDS</span> The GDS Way`. I don't think a `span` can be in a `title` block

galund:
Re fonts I def feel like arguing about hypocrisy wrt the Service Manual, in the GDS Way of all places :-D ... but it's clear the implementation here is borked so will go away and re-work!

galund:
I found the guidance which I claim is fairly clearly points us to Helvetica for the content, but happy to leave for another day. https://docs.google.com/presentation/d/1qmhofgwZpGcebvl2-xsZjiEQNw5fAhEPAcI5qrnqZAs/edit#slide=id.g2c271dfb668_0_29

timblair:
At a screen width where the menu is compressed to a dropdown (even though there's only one item in the menu 🤔) but the side navigation is still shown, there's an issue with the spacing between the logo and the horizontal blue bar.

Screenshot 2024-11-18 at 13 14 42

Also, should the text of the logo and the title have the same baseline, or should the title next vertically aligned with the logo box? Either way, it's not consistent between different views: when the side navigation is shown and the title text is larger, the text baselines are the same. When the side navigation isn't shown (i.e. a mobile view), and the title text is smaller, the alignment changes (although it's not obvious to what the title text is now aligning).

Screenshot 2024-11-18 at 13 18 50

galund:
Needs doing again anyway as we have new branding

PR #935, opened 2024-08-09
Add TypeScript to the editorconfig file

gclssvglx:
As noted in the this guide - TypeScript is already in use at GDS. A recent addition is the GOV.UK App backend.

It would be good for the team developing this to make use of the GDS editorconfig file for consistency reasons.


louzoid-gds (approved):
Looks good to me. Thanks for the PR @gclssvglx

PR #939, opened 2024-08-28
Kick the Python style guide can down the road

robinjam:
We really don't need to review this every 6 months


louzoid-gds:
Yep! Might also be worth bumping the review date if we're happy with the page as it stands now but up to you folks

robinjam:
I haven't done a thorough review so felt it best to leave the review date as-is 😅

@jonodrew @louzoid-gds I don't have write access to this repo, so will need one of you to hit the merge button when ready. Thank you 🙏

PR #941, opened 2024-10-07
Expand languages

huwd:
One Login are using TypeScript / Node / JS a lot for backend so promoting it to a first class option.

We'll need to come back around for mobile languages - but will do that in a separate PR with an eye to style guides.

image

### DIFF
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -26,7 +26,6 @@ This should make it easier for developers to:
 The Service Manual has information on
 [using client-side JavaScript][manual_js]. For server-side JavaScript we use [Node.js][nodejs].
 
-
 [nodejs]: https://nodejs.org/
 [manual_js]: https://www.gov.uk/service-manual/technology/using-progressive-enhancement
 
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -44,11 +44,11 @@ You can use Node.js to render a web interface for your service. For example,
 GOV.UK Pay has created thin, client-facing applications that do not store
 data (although they may retrieve data from an API).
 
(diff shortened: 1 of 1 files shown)


source/standards/programming-languages.html.md.erb
```diff
@@ -59,18 +58,16 @@ Our core languages for backend development are:
 - [Java](/manuals/programming-languages/java.html)
 - [Python](/manuals/programming-languages/python/python.html)
 - [Ruby](/manuals/programming-languages/ruby.html)
+- [JavaScript / Node.js](/manuals/programming-languages/nodejs/index.html)
```
huwd, line 61: Open for challenge on this... I never know what to call the thing. Do we refer to Language here -> `JavaScript` or Superset `TypeScript` or runtime `Node.js`

I use all three relatively interchangably in speech, but should probably be more rigerous in docs... the slash feels like a cop-out.

Perhaps JavaScript is  most right?

two-first-names (approved):

galund (commented):
I think you asked a question that seemed to demand a pedantic answer, so sorry that's what you got!

Do you have a nice editor extension that's doing your curly quotes - I approve in principle if but we value consistency I wouldn't bother.

DaveLee-GDS (commented):
It reads fine to me and is a worth addition. It might be out of scope for this document, but is it worth stating something about recruitment possibly being easier by way of utilising a mainstream language that is more likely covered in the candidate pool?
Thats the only comment i could think to add.

source/standards/programming-languages.html.md.erb
willp-bl: TypeScript / Node.js is better, particularly as a backend language.  It's unlikely we'll want much regular JavaScript.

willp-bl (approved):

huwd:
I'm going to come back around on the TypeScript point, I think I agree but I'd rather get this up and tweak after.

PR #943, opened 2024-10-08
Remove Go content

huwd:
One for folks with more GDS history than me to 👀 but as I reviewed around these changes felt right to me.

### DIFF
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -79,9 +79,23 @@ production.
 
 ### Languages we do not use for new projects
 
+#### Scala
+
 We used Scala in the early days of GDS. GOV.UK Licensing is the only remaining
 application written in Scala but we've found it hard to support because of a lack
 of skills in GDS. Do not use Scala for new projects.
+
+#### Go
+
+Go is no longer a core backend development language in GDS.
+
+The only Go service currently in production operation is the [GOV.UK
+Foundry](https://www.cloudfoundry.org/), which was used by GOV.UK PaaS.
+
+PaaS has now been [decommissioned][pass-decom].
+
+[router]: https://github.com/alphagov/router
+[pass-decom]: https://gds.blog.gov.uk/2022/07/12/why-weve-decided-to-decommission-gov-uk-paas-platform-as-a-service/
 
 ## Using other languages
 
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -77,24 +77,6 @@ production.
 
 [PEP373]: https://www.python.org/dev/peps/pep-0373/
 
-### Go
-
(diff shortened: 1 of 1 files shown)


huwd:
Question... does removing it here mean we should remove the styleguides and links?

https://gds-way.digital.cabinet-office.gov.uk/manuals/programming-languages.html
&
https://gds-way.digital.cabinet-office.gov.uk/manuals/programming-languages/go.html

In my head:
Argument for - Clean up, why maintain redundant pages when we don't have to. These pages are unlikely to see much use, should they have a future need they can be resurrected from the git history.

Argument against - It's still used in GOV.UK's Router... so might be useful to see how to refactor / write fixes in good style. Plus there's a good banner at the top saying it's not core, so limited risk someone gets excited and starts using it, better chance they search find this and find out there's a head of tech to speak to?

source/standards/programming-languages.html.md.erb
```diff
@@ -77,30 +77,26 @@ production.
...
+#### Go
 
 Go is no longer a core backend development language in GDS.
 
 The only Go service currently in production operation is the [GOV.UK
-router][router], and it is the core language for [Cloud
-Foundry](https://www.cloudfoundry.org/), which GOV.UK PaaS uses, although
-GOV.UK PaaS is being [decommissioned][pass-decom]. As such, the knowledge and
-experience of building and running services in Go is small and decreasing.
+Foundry](https://www.cloudfoundry.org/), which was used by GOV.UK PaaS.
```
paroxp, line 93: You may want to keep the precious copy? Router is still used, right? And CloudFoundry is no-mo, along the paas-decom.

timblair: I'm not sure what's happened with this commit, but this paragraph makes no sense. I'm assuming this is supposed to say GOV.UK Router, and there's no need to mention PaaS:

> the GOV.UK Foundry, which was used by GOV.UK PaaS

source/standards/programming-languages.html.md.erb
```diff
@@ -79,10 +79,24 @@ production.
...
 of skills in GDS. Do not use Scala for new projects.
 
+#### Go
+
+Go is no longer a core backend development language in GDS.
+
+The only Go service currently in production operation is the [GOV.UK
+Foundry](https://www.cloudfoundry.org/), which was used by GOV.UK PaaS.
+
+PaaS has now been [decommissioned][pass-decom].
```
timblair, line 95: PaaS has not yet been completely decommissioned. While it's no longer available for general use, it's still being operated within GDS, and used by DBT.

louzoid-gds (commented):
I think I'm mostly minded to leave the guidance on Go as it presently stands as GOV.UK PaaS is still in use and supported. We are also already making it pretty clear that we shouldn't be using it for new things. 
Thoughts?

huwd:
cool cool closing down

PR #944, opened 2024-10-08
Update links to point to govuk-one-login GitHub

huwd:
All our repos should be over there now and pointed to:

https://github.com/search?l=TypeScript&q=user%3Agovuk-one-login+topic%3Adigital-identity&type=Repositories and
https://github.com/search?l=JavaScript&q=user%3Agovuk-one-login+topic%3Adigital-identity&type=Repositories


louzoid-gds (approved):
Good spot thanks Huw

PR #945, opened 2024-10-09
docs: adds mobile programming languages

mcgarj:
This expands on the programming languages used in GDS to include the mobile languages.

This is still in progress and will be updated as specific info is confirmed/discussed amongst the mobile teams

### DIFF
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -26,7 +26,6 @@ This should make it easier for developers to:
 The Service Manual has information on
 [using client-side JavaScript][manual_js]. For server-side JavaScript we use [Node.js][nodejs].
 
-
 [nodejs]: https://nodejs.org/
 [manual_js]: https://www.gov.uk/service-manual/technology/using-progressive-enhancement
 
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -98,6 +98,24 @@ We used Scala in the early days of GDS. GOV.UK Licensing is the only remaining
 application written in Scala but we’ve found it hard to support because of a lack
 of skills in GDS. Do not use Scala for new projects.
 
+## Mobile Development
+
+For developing mobile apps, we use Swift for iOS and Kotlin for Android.
+
+These languages have been chosen over using frameworks such as Flutter or .NET MAUI so that the apps
+feel like a native Android/iOS app without writing extra code to account for the specific OS in Flutter, for example. 
+
+### Swift
+
+The Swift version is generally tied to the version of Xcode you're using. As long as the Xcode version supports the minimum targeted iOS version 
+of your team's app, you can use the most recent version if desired. 
+
+// TODO: Do we need to mention minimum iOS targets for DI / GOV.UK?
+
+### Kotlin 
+ 
+// TODO: Fill in specific Kotlin/Android Studio versions/ OS targets?
+
 ## Using other languages
 
 There will be sensible reasons to not follow the above guidance on languages.
--- a/source/standards/programming-languages.html.md.erb
+++ b/source/standards/programming-languages.html.md.erb
@@ -44,11 +44,11 @@ You can use Node.js to render a web interface for your service. For example,
 GOV.UK Pay has created thin, client-facing applications that do not store
 data (although they may retrieve data from an API).
 
-You can use [TypeScript](https://www.typescriptlang.org/) when teams think it's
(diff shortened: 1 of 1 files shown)


source/standards/programming-languages.html.md.erb
```diff
@@ -45,11 +44,11 @@ You can use Node.js to render a web interface for your service. For example,
 GOV.UK Pay has created thin, client-facing applications that do not store
 data (although they may retrieve data from an API).
 
-You can use [TypeScript](https://www.typescriptlang.org/) when teams think it's
+You can use [TypeScript](https://www.typescriptlang.org/) when teams think it’s
 appropriate. For example, the GOV.UK PaaS team uses TypeScript
 because they are used to working with a statically typed, compiled language,
 and they think the compilation and static-analysis tooling is better for
-their workflow. There's more information about TypeScript on the
+their workflow. Ther’s more information about TypeScript on the
```
cjfryer, line 51:
```suggestion
their workflow. There’s more information about TypeScript on the
```

source/standards/programming-languages.html.md.erb
```diff
@@ -98,6 +98,24 @@ We used Scala in the early days of GDS. GOV.UK Licensing is the only remaining
 application written in Scala but we’ve found it hard to support because of a lack
 of skills in GDS. Do not use Scala for new projects.
 
+## Mobile Development
+
+For developing mobile apps, we use Swift for iOS and Kotlin for Android.
```
huwd, line 103: @mcgarj in line with other docs, consider annotating some of the brand names here with links back to the product pages of that stuff.

So Swift, Kotlin, hell even iOS and Android.

Imagine you're reading this as a dev who (ahem... perhaps like me a few months ago) has zero starting point for this. Let's link richly through to the right pages so someone can go find out what we mean by Swift, Android or even iOS.

source/standards/programming-languages.html.md.erb
```diff
@@ -98,6 +98,24 @@ We used Scala in the early days of GDS. GOV.UK Licensing is the only remaining
...
+
+These languages have been chosen over using frameworks such as Flutter or .NET MAUI so that the apps
+feel like a native Android/iOS app without writing extra code to account for the specific OS in Flutter, for example.
```
huwd, line 106: I think this is a great statement, and one I might like to see expanded.

I wonder if one of the mobile seniors or folks involved with these decisions wouldn't mind chipping in here?

I think there's a candidate here for a separate GDS Way page about "preferences for native first / web first or hybrid". 

It'd be good if we could write this paragraph in a way we could link off to that when we have more guidance later.

source/standards/programming-languages.html.md.erb
```diff
@@ -98,6 +98,24 @@ We used Scala in the early days of GDS. GOV.UK Licensing is the only remaining
...
+
+For developing mobile apps, we use Swift for iOS and Kotlin for Android.
+
+These languages have been chosen over using frameworks such as Flutter or .NET MAUI so that the apps
+feel like a native Android/iOS app without writing extra code to account for the specific OS in Flutter, for example. 
+
+### Swift
+
+The Swift version is generally tied to the version of Xcode you're using. As long as the Xcode version supports the minimum targeted iOS version 
+of your team's app, you can use the most recent version if desired.
```
huwd, line 111: huh the suggestion tool disappeared from github, I must be doing something odd...

Consider putting this on one line

huwd (commented):
Great start Joe thanks for this.
Keen to get some eyes from other senior colleagues and those language specific bits filled in

source/standards/programming-languages.html.md.erb
kevindew: I think we'd be overreaching here to specify in this that we have an organisational preference for single platform mobile apps over cross platform tooling - as I understand DI and GOV.UK both made the decision to go single platform based on individual project requirements.

I think it'd be better to just leave how someone would approach a cross platform app as undefined here as we've not yet established a pattern/preferred framework in the org - and maybe we never will? And then to amend the above sentence to be: "For developing native mobile apps, we use Swift for iOS and Kotlin for Android."

I imagine there could be a sentence or two to explain why Kotlin over an alternative like Java for native Android development and why Swift over Objective-C for native iOS - but it could be a little superfluous given how dominant both languages in the sphere. So as an alternative something that just says "These languages have been chosen as they are the common industry standard and vendor recommended languages to use for their respective platforms"

pcornish (approved):

mcgarj:
Closing this one, replacing it with https://github.com/alphagov/gds-way/pull/946 as that one has a clean start

PR #949, opened 2024-10-23
Update mention of Puppet

nicholsj:
GOV.UK no longer uses Puppet. It might be that the whole Puppet section can be removed, but I'll leave that to others to decide.

### DIFF
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -10,7 +10,7 @@ Use [configuration management][] to manage, automate and standardise your infras
 
 ## Puppet
 
-The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services. It's mainly still in use on [GOV.UK](https://github.com/alphagov/govuk-puppet) but this will decline as more services are moved over to AWS EKS.
+The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services.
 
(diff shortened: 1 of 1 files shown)


source/standards/configuration-management.html.md.erb
```diff
@@ -10,7 +10,7 @@ Use [configuration management][] to manage, automate and standardise your infras
 
 ## Puppet
 
-The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services. It's mainly still in use on [GOV.UK](https://github.com/alphagov/govuk-puppet) but this will decline as more services are moved over to AWS EKS.
+The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services.
 
 If your environment consists of a simple deployment artefact like an [Amazon Machine Image (AMI)][], Puppet may not be necessary, but the process for building that artefact must still be codified and version controlled.
```
willp-bl, line 15: we should keep this line or something like it

source/standards/configuration-management.html.md.erb
```diff
@@ -10,7 +10,7 @@ Use [configuration management][] to manage, automate and standardise your infras
 
 ## Puppet
 
-The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services. It's mainly still in use on [GOV.UK](https://github.com/alphagov/govuk-puppet) but this will decline as more services are moved over to AWS EKS.
+The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services.
```
willp-bl, line 13: As far as I can tell we now don't use it at all

louzoid-gds (approved):
Let's get this change merged

PR #950, opened 2024-10-23
Updated Dockerfile reference and updated Dockerfile to match guidance

andyloughran:
Updated the Dockerfile reference documentation as needed a refresh.

Referencing the OCI so that we can link back to the open standards that we're following, and the reason for following a particular implementation of the open standard.

Updated our ~/Dockerfile to match our guidance regarding sha pinning.


mrwilson (approved):

source/manuals/programming-languages/docker.html.md.erb
```diff
@@ -20,7 +27,8 @@ A tag is a short label you can use to reference an image.
 For example:
 
 ```
-FROM alpine:3.9
+# 3.9
+FROM alpine
```
andyloughran, line 31:
```suggestion
FROM alpine:3.9
```

source/manuals/programming-languages/docker.html.md.erb
```diff
@@ -20,6 +27,7 @@ A tag is a short label you can use to reference an image.
 For example:
 
 ```
+# 3.9
```
andyloughran, line 30:
```suggestion
```

source/manuals/programming-languages/docker.html.md.erb
```diff
@@ -31,11 +38,15 @@ where:
...
-expect.
+expect.  Add a comment above the FROM line with the intended tag and version.
+Though dependabot will not use this, please update it when the digest changes
+so that it is possible to understand the magnitude of the change by knowing
+the significance of the version change.
 
 For example:
 
 ```
+# alpine:3.9
```
andyloughran, line 49:
```suggestion

```

source/manuals/programming-languages/docker.html.md.erb
```diff
@@ -55,6 +65,16 @@ As [Dependabot](https://dependabot.com) has [support for updating `FROM` lines
 which use digests](https://github.com/dependabot/dependabot-core/pull/100),
 you can still use Dependabot to keep your images up-to-date.
 
+If you specify both the tag and the digest, then the digest takes precedence.
+The standard implies that you may only specify one or the other.  Do not specify
+both as the behaviour may vary across different platforms.
+
```
andyloughran, line 71:
```suggestion

Instead, the recommended way to keep track of the intended version is to add it in a line above FROM as a comment.

Please double-check that the digest and the commented version are consistent each time you upgrade, as dependabot does not have perfect capability at either identifying the magnitude of the upgrade, or necessarily updating the commented pin.
```

galund (commented):
I feel I don't know how I'd go about updating the Dockerfile, but that's a me problem not necessarily something for the GDS Way to fix. It might want to be in the readme for the repo, but I guess if it's the same in all cases then maybe not.

andyloughran:
> I feel I don't know how I'd go about updating the Dockerfile, but that's a me problem not necessarily something for the GDS Way to fix. It might want to be in the readme for the repo, but I guess if it's the same in all cases then maybe not.

Do you want me to document how to update the sha of a Dockerfile based on the comment above the FROM?  I'll happily do that and I think you're right - it's implicit but it should be more explicit.

willp-bl:
> > I feel I don't know how I'd go about updating the Dockerfile, but that's a me problem not necessarily something for the GDS Way to fix. It might want to be in the readme for the repo, but I guess if it's the same in all cases then maybe not.
> 
> Do you want me to document how to update the sha of a Dockerfile based on the comment above the FROM? I'll happily do that and I think you're right - it's implicit but it should be more explicit.

That will be a good addition.  Once that is done is this PR ready to be merged?

source/manuals/programming-languages/docker.html.md.erb
```diff
@@ -31,7 +38,10 @@ where:
 As you cannot rely on the tag pointing to the exact same image over time, you
 should instead use a digest, which identifies the image by a hash of its
 contents. This makes sure that you are always referencing the image that you
-expect.
+expect.  Add a comment above the FROM line with the intended tag and version.
+Though dependabot will not use this, please update it when the digest changes
+so that it is possible to understand the magnitude of the change by knowing
+the significance of the version change.
 
 For example:
```
galund, line 46: The example should include the comment, really

galund: tbh actually the new 41-45 feel superfluous given the new guidance below about including the tag and the comment.

Why not just remove all the examples except the example that shows how we _want_ this to be done, which is the new one at line 78?

galund: I just did it, hope that's ok

galund (approved):

huwd (approved):

PR #954, opened 2024-10-31
Update to the Secret Auditing page

minglis:
As per discussion in GDS Lead Tech meeting - action was to review the guidance on GitHub secret scanning with security colleagues. 

Proposal is to upgrade the use of GitHub secret scanning to a must from should. 

This change also recommends that teams must review alerts on a regular basis, timing of this suggested to be aligned to existing team processes.

Teams also must record outcomes, even for false positives.


willp-bl (approved):

huwd (approved):
I agree with this DI are making similar moves right now.

timblair (changes requested):
We discussed this in the GDS Way meeting, and are :+1: with it. Please can you update the review date before you merge this?

timblair (changes requested):
Just one additional bump to the review date, otherwise 👍

timblair (approved):

PR #962, opened 2024-11-13
New Diagrams as code page

james-sheppard-gov-uk:
The GDS Way doesn't yet contain guidance for creating diagrams. This is a new page to promote a diagrams as code approach.

### DIFF
--- a/source/standards/diagrams-as-code.html.md.erb
+++ b/source/standards/diagrams-as-code.html.md.erb
@@ -27,7 +27,7 @@ Diagrams are a form of documentation, and some teams already use a '[docs as cod
 
 Refer to [guidance on making images or charts accessible](https://www.gov.uk/guidance/publishing-accessible-documents).
 
-Since the source of the diagram provides a text based representation of the generated image, this probably makes it more accessible than a diagram drawn through a graphic editor. It may be preferably to signpost users to get to the source of the diagram rather than duplicating it alongside the diagram itself.
+Since the source of the diagram provides a text based representation of the generated image, this probably makes it more accessible than a diagram drawn through a graphic editor. It may be preferable to signpost users to get to the source of the diagram rather than duplicating it alongside the diagram itself.
 
 ## Diagrams as code tools
 
--- a/source/standards/diagrams-as-code.html.md.erb
+++ b/source/standards/diagrams-as-code.html.md.erb
@@ -13,7 +13,7 @@ Code can be used to define the content, structure and appearance of a diagram wh
 > A good diagram enables you to more easily convey your design to others, and it provides a focus for your creative, as well as your technical thinking.
 — [Drawing good architecture diagrams](https://www.ncsc.gov.uk/blog-post/drawing-good-architecture-diagrams) tips from NCSC
 
-There are many ways to create a diagram, from pen and paper or whiteboard through to specialist diagramming software. The right tool will depend on the circumstances, but it's always worth considering a diagram generated from code.
+There are many ways to create a diagram, from pen and paper or whiteboard through to specialist diagramming software. The right tool will depend on the circumstances, including what the diagram is for and who needs to be able to contribute to it, but it's always worth considering a diagram generated from code.
 
 Diagrams are a form of documentation, and some teams already use a '[docs as code](https://technology.blog.gov.uk/2017/08/25/why-we-use-a-docs-as-code-approach-for-technical-documentation/)' approach. Diagrams as code builds on the rationale and benefits of docs as code, including
 
--- a/source/standards/diagrams-as-code.html.md.erb
+++ b/source/standards/diagrams-as-code.html.md.erb
@@ -1,0 +1,41 @@
+---
+title: Diagrams as code
+last_reviewed_on: 2024-11-13
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+Code can be used to define the content, structure and appearance of a diagram which can then be rendered, instead of using a drawing tool directly. This approach is known as "diagrams as code".
+
+## What are diagrams for?
+
+> A good diagram enables you to more easily convey your design to others, and it provides a focus for your creative, as well as your technical thinking.
+
+https://www.ncsc.gov.uk/blog-post/drawing-good-architecture-diagrams
+
+There are many ways to create a diagram, from pen and paper or whiteboard through to specialist diagramming software. The right tool will depend on the circumstances, but it's always worth considering a diagram generated from code.
+
+Diagrams are a form of documentation, and some teams already use a '[docs as code](https://technology.blog.gov.uk/2017/08/25/why-we-use-a-docs-as-code-approach-for-technical-documentation/)' approach. Diagrams as code builds on the rationale and benefits of docs as code, including
+
+* keeping documentation close to source code
+* ease of change
+* a workflow to review changes
+* version history
+* consistency
+
+### Diagrams as code tools
+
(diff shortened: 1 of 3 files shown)


louzoid-gds (commented):
Thanks for getting this draft up James! I've added a couple of format comments.

In regards to the content, I wonder if it's missing a section on 'When not to use diagrams as code?' i.e. when you know non-technical folks need to edit them (barrier to entry concerns). I'm not wedded to that but keen on your thoughts

source/standards/diagrams-as-code.html.md.erb
```diff
@@ -0,0 +1,41 @@
...
+
+# <%= current_page.data.title %>
+
+Code can be used to define the content, structure and appearance of a diagram which can then be rendered, instead of using a drawing tool directly. This approach is known as "diagrams as code".
+
+## What are diagrams for?
+
+> A good diagram enables you to more easily convey your design to others, and it provides a focus for your creative, as well as your technical thinking.
+
+https://www.ncsc.gov.uk/blog-post/drawing-good-architecture-diagrams
```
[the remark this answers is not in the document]
james-sheppard-gov-uk: I improved the attribution following this https://stackoverflow.com/a/65598531

james-sheppard-gov-uk:
> Thanks for getting this draft up James! I've added a couple of format comments.
> 
> In regards to the content, I wonder if it's missing a section on 'When not to use diagrams as code?' i.e. when you know non-technical folks need to edit them (barrier to entry concerns). I'm not wedded to that but keen on your thoughts

Thanks @louzoid-gds ! I've added a note about "who needs to be able to contribute to it". I've not yet encountered a problem along these lines so interested if anyone else has?

galund:
It might be worth mentioning that there is a maintenance burden, depending on the support available in your publication pipeline. If you're publishing on GitHub, it's great, but for example:

* the Tech Docs Template doesn't support diagrams-as-code (AFAIK)
* (even more obscurely but relevant to DI) some [magic had to be worked](https://github.com/govuk-one-login/architecture/blob/main/scripts/publish-mermaid.sh) to render them for Confluence (and our results are still a bit suboptimal)

With enough will it's all do-able, you just have to put some effort into a quality publication process, and making capacity for that seems to be difficult.

source/standards/diagrams-as-code.html.md.erb
```diff
@@ -0,0 +1,40 @@
...
+
+Diagrams are a form of documentation, and some teams already use a '[docs as code](https://technology.blog.gov.uk/2017/08/25/why-we-use-a-docs-as-code-approach-for-technical-documentation/)' approach. Diagrams as code builds on the rationale and benefits of docs as code, including
+
+* keeping documentation close to source code
+* ease of change
+* a workflow to review changes
+* version history
+* consistency
+
+## Diagrams as code tools
```
minglis, line 26: Appreciate this is only an example, but perhaps worth calling out a couple of other tools? C4 perhaps? Others that are in use?

Ian-Nice-16: Notify have been using D2, and also mermaid in markdown depending on the diagram complexity and team member preference.  As mentioned on the last Technical Architects Fortnightly, it is good to suggest but not mandate the tool.

galund:
> Love using Mermaid for coding diagrams, so it's nice to see the approach promoted. 🥳 Thanks @james-sheppard-gov-uk.
> 
> One thought: should this be checked by an accessibility specialist, in case there are any impacts for folks using screen-readers and other kinds of assistive technologies to highlight (possible caveats to avoid, but also potential benefits, one hypothesis being that the source of the diagram provides a text based representation of the generated image, possibly making it more accessible than a diagram drawn through a graphic editor)?

Mostly agree, I think it's worth mentioning the guidance from https://www.gov.uk/guidance/publishing-accessible-documents

> If you’re using images or charts, think about how you’ll make the content accessible to people with a visual impairment. Two options are:
>
> * make the same point in the text of the document (so people with visual impairments get the information they need - the image or chart is there as an extra for people who are able to see it)
> * give the person converting or uploading the document for you alt text (‘alternative text’) for the image or chart

Frequently, the first of those options is better - I end up putting in the diagram, and describing it with bullet points underneath. This is good for clarification of what the diagram means and can provoke useful discussion. You then only need a very minimal alt text.

Ian-Nice-16:
Looks great @james-sheppard-gov-uk.  As this talks about source control and version history, I wonder if there is any value of recommending that diagrams not generated as code (perhaps legacy ones) are also stored in source form alongside diagrams as code.  I.e. storing a draw.io source file along side other diagram as code docs stored in source control. Then  we don't have to redraw diagrams from PDF exports where the original source was lost?  This may not belong in the diagrams as code page (it might belong in a more generic documenting architecture page)

source/standards/diagrams-as-code.html.md.erb
james-sheppard-gov-uk: This was discussed at the GDS Way Forum on 2024-11-27

I've added Mermaid sequence diagrams as that's the tool and diagram type I have the most experience using. The intention is that after this initial PR is merged and the page is created, others can add details about tools they have successfully used, ideally with links to examples. [C4](https://c4model.com/) is a tooling-independent model, and I'm very interested in exploring the use of [Structurizr](https://www.structurizr.com/) - I will of course add it here if successful...

source/standards/diagrams-as-code.html.md.erb
```diff
@@ -0,0 +1,40 @@
...
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+Code can be used to define the content, structure and appearance of a diagram which can then be rendered, instead of using a drawing tool directly. This approach is known as "diagrams as code".
+
+## What are diagrams for?
+
+> A good diagram enables you to more easily convey your design to others, and it provides a focus for your creative, as well as your technical thinking.
```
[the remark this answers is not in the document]
james-sheppard-gov-uk: The intention is that other guidance about documentation and diagrams should exist in The GDS Way but separate from the diagrams-as-code guidance.

james-sheppard-gov-uk:
> Love using Mermaid for coding diagrams, so it's nice to see the approach promoted. 🥳 Thanks @james-sheppard-gov-uk.
> 
> One thought: should this be checked by an accessibility specialist, in case there are any impacts for folks using screen-readers and other kinds of assistive technologies to highlight (possible caveats to avoid, but also potential benefits, one hypothesis being that the source of the diagram provides a text based representation of the generated image, possibly making it more accessible than a diagram drawn through a graphic editor)?

Great suggestion, thanks @romaricpascal ! I've had a chat with @selfthinker (thanks for your help!) on Slack which I'd summarise as:
* agreement that the source will probably be more accessible than the graph
* consider signposting a user to get to the source of the diagram, as duplicating the source code on the page itself would introduce a maintenance burden and a chance for things to get out of sync
* Mermaid can create SVGs with a [`<title>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) and/or [`<desc>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc) element that can be defined in the code, similar to the `alt` attribute of an `` element
* consider [contributing to Mermaid](https://mermaid.js.org/community/intro.html) to make it more accessible

I'll update the PR accordingly...

james-sheppard-gov-uk:
> Looks great @james-sheppard-gov-uk. As this talks about source control and version history, I wonder if there is any value of recommending that diagrams not generated as code (perhaps legacy ones) are also stored in source form alongside diagrams as code. I.e. storing a draw.io source file along side other diagram as code docs stored in source control. Then we don't have to redraw diagrams from PDF exports where the original source was lost? This may not belong in the diagrams as code page (it might belong in a more generic documenting architecture page)

Thanks @Ian-Nice-16 - let's consider this when the "more generic documenting architecture page" gets created...

PR #985, opened 2025-02-04
Deprecate Terraform, Add OpenTofu, Add CloudFormation, Add CDK, Add Helm

paroxp:
This Pull Request intends to spark some conversations among the tech community.

### TODO

* [ ] Add Helm paragraph

### DIFF
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -1,7 +1,7 @@
 ---
 title: Use configuration management
-last_reviewed_on: 2022-09-25
-review_in: 6 months
+last_reviewed_on: 2025-02-10
+review_in: 24 months
 ---
 
 # <%= current_page.data.title %>
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -68,6 +68,14 @@ benefits they could bring to your build pipelines.
 
  * [tofu test](https://opentofu.org/docs/cli/commands/test/) - "performs tests written for OpenTofu IaC"
 
+## CloudFormation and CDK
+
+Amazon Web Services provide two tools in the field of IaC. [CloudFormation] being more traditional YAML manifests or
+[CDK] being more of an SDK library for developers to program their way into configuration.
+
+Both are limited to only cover AWS infrastrcuture, and woould not be much use outside of that echosystem - for instance,
+configuring GitHub or Dynatrace would require you to invest in additional tooling such as [OpenTofu].
+
 ## Further reading
 
 Find out more about configuration management in the [Service Manual][].
@@ -83,3 +91,5 @@ Find out more about configuration management in the [Service Manual][].
 [govuk-aws]: https://github.com/alphagov/govuk-aws
 [large number of providers]: https://www.terraform.io/docs/language/providers/index.html
 [OpenTofu]: https://opentofu.org/
+[CloudFormation]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/Welcome.html
+[CDK]: https://docs.aws.amazon.com/cdk/v2/guide/home.html
--- a/source/standards/configuration-management.html.md.erb
+++ b/source/standards/configuration-management.html.md.erb
@@ -6,31 +6,56 @@ review_in: 6 months
 
 # <%= current_page.data.title %>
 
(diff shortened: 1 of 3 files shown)


source/standards/configuration-management.html.md.erb
```diff
@@ -1,36 +1,61 @@
...
 
-## Puppet
+## Avoided technologies
 
-The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services. It's mainly still in use on [GOV.UK](https://github.com/alphagov/govuk-puppet) but this will decline as more services are moved over to AWS EKS.
+### Puppet
 
-If your environment consists of a simple deployment artefact like an [Amazon Machine Image (AMI)][], Puppet may not be necessary, but the process for building that artefact must still be codified and version controlled.
+The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level
+services. It is mainly still in use on [GOV.UK](https://github.com/alphagov/govuk-puppet) but this will decline as more
```
sihugh, line 17: GOV.UK are now off puppet, so you may delete this bit too if you like. Or leave it and I'll do a separate update

nicholsj: https://github.com/alphagov/gds-way/pull/949

two-first-names:
Are there any teams in GDS who are currently using OpenTofu? I'm wary of swapping out Terraform to OpenTofu in our guidance if we don't have the evidence to back it up, you know?

roger-watkins:
My principal concerns around moving to OpenTofu would be:
- How well supported it is compared to Terraform (how likely is it to become abandonware)
- How difficult is it to rollback from recent versions of Terraform to OpenTofu

Concerns about remaining with Hashicorp Terraform:
- Will they do a docker and start requiring a license to work with their toolchain ?

galund:
The ADR in question is here https://github.com/govuk-one-login/architecture/blob/main/adr/0042-use-of-sam-and-cloudformation.md and would need revoking.

I think the ADR strikes an OK balance because in practice teams like FOG have found that for a large proportion of their work, it's just much more efficiently done in Terraform. And EE (and the rest of GDS) uses Terraform. The pull in favour of CloudFormation was (a) SAM and (b) native AWS-ness, but in practice some things are just better in Terraform.

Fine to be provocative if a discussion is needed, but it strikes me this needs consultation with Platform/SRE folks - and preferably alignment with the rest of GDS - before taking forwards. (George disappears again back into his wheelhouse, which is not to be commenting here!)

dazahern:
Terraform may no longer be the new kid on the block, but it is the industry standard. 

Concerns for me;
- the effort/people/months of work needed to unpick everything in favour of opentofu
- is it compatible (like for like replacement) with our various pieces of infra
- we are dropping TF in favour of further integration with the AWS suite (what if they deprecate one of these, raise prices, goes down)

source/standards/configuration-management.html.md.erb
```diff
@@ -39,7 +64,17 @@ benefits they could bring to your build pipelines.
...
+ * [tflint](https://github.com/terraform-linters/tflint) - "linter focused on possible errors, best practices and so on."
+
+ * [tofu test](https://opentofu.org/docs/cli/commands/test/) - "performs tests written for OpenTofu IaC"
+
+## CloudFormation and CDK
+
+Amazon Web Services provide two tools in the field of IaC. [CloudFormation] being more traditional YAML manifests or
+[CDK] being more of an SDK library for developers to program their way into configuration.
+
+Both are limited to only cover AWS infrastrcuture, and woould not be much use outside of that echosystem - for instance,
```
jamesmelville-gds, line 76:
```suggestion
Both are limited to only cover AWS infrastructure, and would not be much use outside of that ecosystem - for instance,
```

source/standards/configuration-management.html.md.erb
```diff
@@ -39,7 +64,17 @@ benefits they could bring to your build pipelines.
...
+configuring GitHub or Dynatrace would require you to invest in additional tooling such as [OpenTofu].
```
jamesmelville-gds, line 77: This could be investigated for dynatrace
https://github.com/aws-ia/cloudformation-dynatrace-resource-providers

paroxp:
Lack of appetite we have to pilot the changes, indicates the lack of readiness for these discussions. Instead of allowing this PR hanging indefinitely, we'll close it down, and encourage anyone to come back to it in the future if we get a little more excited again.

PR #986, opened 2025-02-06
Least privilege

jonodrew:
I've rephrased the top line and reduced the privileged access session time.

### DIFF
--- a/source/standards/principle-least-access.html.md.erb
+++ b/source/standards/principle-least-access.html.md.erb
@@ -52,5 +52,6 @@ If you’re using the gds-users account to log into your AWS accounts, you shoul
 - [NCSC - 10 steps to cyber security](https://www.ncsc.gov.uk/collection/10-steps)
 - [NCSC - privileged user management](https://www.ncsc.gov.uk/guidance/introduction-identity-and-access-management#section_6)
 - [NIST Special Publication 800-53 - AC-6 least privilege][polp]
+- [Government cyber policy - B2 identity and access control](https://www.security.gov.uk/policy-and-guidance/government-cyber-security-policy-handbook/principle-b2-identity-and-access-control/)
 
-[polp]: https://csrc.nist.gov/Projects/risk-management/sp800-53-controls/release-search
+[polp]: https://csf.tools/reference/nist-sp-800-53/r5/ac/ac-6/


source/standards/principle-least-access.html.md.erb
```diff
@@ -32,7 +32,7 @@ Your team should:
 - create the roles or credentials with the least possible privilege, with only necessary permissions required for normal users to perform their day-to-day jobs
 - use the role or credentials with the least possible privilege as the default option
 - use just-in-time (JIT) access provisioning to grant users an on-demand, time-limited privileged role or security token to access the privileged resources
-- make sure session time of the privileged access is set to no more than 12 hours, and/or terminates when the user logs out of their laptop
+- make sure session time of the privileged access is set to no more than 30 minutes, and/or terminates when the user logs out of their laptop
```
galund, line 35:
```suggestion
- make sure session time of the privileged access (in non-development environments) is set to no more than 30 minutes, and/or terminates when the user logs out of their laptop
```

galund:
Noticed there's also a '12 hours' at line 47 that might want to be consistent

source/standards/principle-least-access.html.md.erb
```diff
@@ -1,12 +1,12 @@
...
 title: Principle of Least Privilege
-last_reviewed_on: 2024-06-27
+last_reviewed_on: 2025-02-06
 review_in: 6 months
 ---
 
 # <%= current_page.data.title %>
 
-The [principle of least privilege][polp] involves setting up user accounts so they can only access and use the information they need for specific tasks. This can also apply to processes and individuals who might have to switch between normal access and the increased access of a superuser account as part of their work.
+The [principle of least privilege][polp] requires you to consider what permissions actors in your system need. For robotic or computer actors, like your deployment pipeline, this can be fairly easy to scope. Likewise, your end-users will have a clearly defined set of permissions they need to use your service.
```
lfdebrux, line 9: I feel like it would be good to keep the brief definition of the principle of least privilege.

galund: I don't know if it ever did, but the `polp` reference here doesn't actually seem to define the principle either... there's some [good discussion in NCSC guidance](https://www.ncsc.gov.uk/collection/zero-trust-architecture/know-your-user-service-device-identities#section_2) and it's worth noting that NCSC fairly consistently put the principle in quotes - my take is that's because this is almost a meta-principle: a guiding star to aim for rather than an inviolable policy with direct implications for designing a set of privileges.

jonodrew: I don't think I'd agree with that. It's a guiding star and therefore it has direct implications for how we design things. 

To @lfdebrux's point - I agree, I'll add it back in

galund (approved):
I've fixed the NIST link and added a UK policy one, and while I still feel like I might want to revisit this soonish in preparation for Entra, this is a definite improvement IMO.

PR #992, opened 2025-02-17
Update working-with-aws-accounts page to include new processes

p-berkeley:
- Updated outdated guidance on AWS organizations and account requests process
- Added links to GDS Engineering Hub for more sensitive information which can't be on the public internet
- Also replaced references from the defunct Reliability Engineering with the Engineering Enablement Team

### DIFF
--- a/source/manuals/working-with-aws-accounts.html.md.erb
+++ b/source/manuals/working-with-aws-accounts.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Working with AWS accounts
-last_reviewed_on: 2024-02-18
+last_reviewed_on: 2025-02-18
 review_in: 12 months
 ---
 
--- a/source/manuals/working-with-aws-accounts.html.md.erb
+++ b/source/manuals/working-with-aws-accounts.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Working with AWS accounts
-last_reviewed_on: 2024-2-18
+last_reviewed_on: 2024-02-18
 review_in: 12 months
 ---
 
--- a/source/manuals/working-with-aws-accounts.html.md.erb
+++ b/source/manuals/working-with-aws-accounts.html.md.erb
@@ -6,7 +6,7 @@ review_in: 12 months
 
 # <%= current_page.data.title %>
 
-Most teams in GDS use [Amazon Web Services (AWS)](https://aws.amazon.com/) as their infrastructure provider. GDS teams in GOV.UK and DSP manage their own AWS accounts, but users must first sign into a shared base AWS account called `gds-users`. They can then assume roles in their team's AWS account to perform administrative tasks using [AWS's cross-account access pattern](https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html).
+Teams in GDS should use [Amazon Web Services (AWS)](https://aws.amazon.com/) as their core infrastructure provider. GDS uses multiple AWS Organizations. You can find more details about how the core GDS AWS Platform is managed on the [GDS Hub](https://engineering-enablement.gds-reliability.engineering/interim-org-platform.html). GDS teams in GOV.UK and DSP manage their own AWS accounts, but users must first sign into a shared base AWS account called `gds-users`. They can then assume roles in their team's AWS account to perform administrative tasks using [AWS's cross-account access pattern](https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html).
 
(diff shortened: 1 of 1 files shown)


C-gyorfi (approved):

source/manuals/working-with-aws-accounts.html.md.erb
```diff
@@ -6,7 +6,7 @@ review_in: 12 months
...
+Teams in GDS should use [Amazon Web Services (AWS)](https://aws.amazon.com/) as their core infrastructure provider. GDS uses multiple AWS Organizations. You can find more details about how the core GDS AWS Platform is managed on the [GDS Hub](https://engineering-enablement.gds-reliability.engineering/interim-org-platform.html). GDS teams in GOV.UK and DSP manage their own AWS accounts, but users must first sign into a shared base AWS account called `gds-users`. They can then assume roles in their team's AWS account to perform administrative tasks using [AWS's cross-account access pattern](https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html).
 
 Note that GOV.UK One Login has its own AWS Organization separate from GDS / Cabinet Office accounts. That is managed by the programme, and separate guidance applies.
```
jonodrew, line 11: Change this to DSIT while we're here? Or just drop the parent department altogether?

tomvaughan77 (approved):

source/manuals/working-with-aws-accounts.html.md.erb
```diff
@@ -197,9 +197,9 @@ AWS Answers has more information about [AWS Multiple Account Security Strategy][
 
 ### Request an account
 
-To request an AWS account complete and submit the [request an AWS account form][].
+As of Febuary 2025, GDS users can request an AWS account by following instructions on the [GDS Engineering Hub](https://engineering-enablement.gds-reliability.engineering/interim-org-platform.html#account-vending-creating-a-new-account).
 
-Reliability Engineering will confirm if your account can be included in GDS AWS billing and perform the initial account setup.
+The GDS Engineering Enablement Team will confirm if your account can be included in GDS AWS billing and perform the initial account setup.
```
galund, line 202: I'd have thought the assumption was that we are looking after the billing for everything in the new org, so maybe this para should turn into something about scope (who in DSIT would we turn down an account request from?).

source/manuals/working-with-aws-accounts.html.md.erb
p-berkeley: Valid question, 
We still haven't confirmed every detail about which teams in which organisations are managing which functions once we've moved the AWS organizations to a DSIT contract (especially with invoicing and billing). 
So I think mentioning CO is fine for the time being, even if it's a bit clunky.

source/manuals/working-with-aws-accounts.html.md.erb
p-berkeley: I've deliberately not mentioned DSIT billing/invoicing for now, there's quite a few unknowns, 
Do you think that's a problem?

source/manuals/working-with-aws-accounts.html.md.erb
galund: Hear ya but it does seem a bit strange to mention CO but not DSIT. And also maybe we can just make this more useful (seeing as it's not even covered by the DI Team Manual, and you have to poke around in Confluence to find the answer).

```suggestion
Note that GOV.UK One Login has its own AWS organisation, managed by the Digital Identity directorate.
More information is available in the
[#di-aws-control-tower](https://gds.slack.com/archives/C04UJ0F91QA)
Slack channel, and the guidance here does not apply.
```

source/manuals/working-with-aws-accounts.html.md.erb
stephengrier: The process involves raising a PR against an internal alphagov repo so I wouldn't expect anyone from outside GDS to be requesting accounts via this process.

I think the previous wording was not great. Maybe just say something like "The GDS Engineering Enablement team will review your request and if acceptable will create the account."

source/manuals/working-with-aws-accounts.html.md.erb
```diff
@@ -1,14 +1,17 @@
 ---
 title: Working with AWS accounts
-last_reviewed_on: 2023-11-24
+last_reviewed_on: 2024-02-18
```
stephengrier, line 3: 2025

source/manuals/working-with-aws-accounts.html.md.erb
stephengrier: Given that we're only correcting the team name here and not the actual text I think this line is fine for now.

galund (approved):

PR #1006, opened 2025-02-25
DNS guidance update

galund:
Fairly radical rewrite including

- removing dual hosting [per GOV.UK thinking](https://github.com/alphagov/govuk-dns-tf/tree/main?tab=readme-ov-file#amazon-route53-vs-google-cloud)
- adding some notes about practice for non-prod domain names (someone needs to tell me this is actually correct - it's what DI do I believe!)
- created some instructions for how to get an internal domain as this didn't seem to be written down anywhere


source/standards/dns-hosting.html.md.erb
```diff
@@ -31,3 +31,18 @@ You can read more about [service domains and DNS][] in the Service Manual.
 [Amazon Route 53]: https://aws.amazon.com/route53/
 [Google Cloud DNS]: https://cloud.google.com/dns/
 [service domains and DNS]: https://www.gov.uk/service-manual/technology/get-a-domain-name
+
+## Production and non-production domain names
+
+Your non-production domains should be subdomains of your production domain.
+
+For example, if you service can be found at `myservice.service.gov.uk`, then you might also have Route 53 records for your other environments, such as:
```
jonodrew, line 28: This point's under discussion at the moment. I agree with it, but I don't think it's been properly agreed

source/standards/dns-hosting.html.md.erb
```diff
@@ -1,33 +1,58 @@
...
-When you implement your DNS strategy you should consider using:
+Your non-production domains should be subdomains of your production domain.
 
-* [Amazon Route 53][] as a cloud DNS web service
-* [Google Cloud DNS][] for high nameserver availability
+For example, if you service can be found at `myservice.service.gov.uk`, then you might also have Route 53 records for your other environments, such as:
 
-You can read more about [service domains and DNS][] in the Service Manual.
+* `staging.myservice.service.gov.uk`
+* `dev.myservice.service.gov.uk`
```
galund, line 31:
```suggestion
* `staging.myservice.service.gov.uk`
* `dev.myservice.service.gov.uk`

> [!TIP]
> In this example it's important to avoid setting cookies for the entire domain `myservice.service.gov.uk`.
> Always follow the
> [OWASP guidance on cookie domains](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#domain-and-path-attributes)
> such as to avoid using the `Domain` cookie attribute.
> You can also consider using a separate domain for your production service, such as `www.myservice.service.gov.uk`.```

File: source/standards/dns-hosting.html.md.erb (hunk 1 above)
Comment (galund) [reply source/standards/dns-hosting.html.md.erb:28]: Happy to leave this open assuming the discussion isn't gonna take forever, LMK :)

File: source/standards/dns-hosting.html.md.erb [hunk 3]
```diff
@@ -8,24 +8,13 @@ review_in: 12 months
 
 To ensure users can connect to your service, you'll need accurate Domain Name System (DNS) records and a DNS provider to supply your DNS nameservers.
 
-You can find information on choosing where to host your DNS and how to request nameserver delegation at GDS in the [Service Manual][].
+For services [hosted on AWS](./hosting.html) you should use [Amazon Route 53][] as a cloud DNS web service.
 
-As part of your DNS strategy and planning you need to ensure:
+> [!INFO]
+> In the past some services have dual-hosted using [Google Cloud DNS][] for high nameserver availability.
```
sihugh, line 14: To be fair GOV.UK do still dual host DNS, but we've opted to not do it on certain zones that we operate that are not part of our core service.
I think (particularly as GOV.UK went offline due to using [a single DNS provider](https://en.wikipedia.org/wiki/DDoS_attacks_on_Dyn)) that we'd need to do more work to assess the risk before we did go back to single hosting.
We haven't had outages due to dual-hosting.

sihugh: I accept the point that this it is not necessary for all services to follow this approach.

galund: Ah, perhaps the reference in the GOV.UK DNS repo is a bit misleading then?

> Historically, zones were dual-hosted for redundancy, but this adds complexity to the point where it arguably causes more risk of an outage itself than the risk it is trying to mitigate.
> 
> Ideally, we would make things consistent by removing all zones from Google Cloud, but this is more risk and effort than it's worth.

sihugh: Yep, I think that was more opinion than outright policy 👍

sihugh: The default in that repo is still to dual host - you can see [where that's been overridden](https://github.com/search?q=repo%3Aalphagov%2Fgovuk-dns-tf%20gcp_disabled&type=code)

galund: Cool, I've suggested a wording tweak below, but feel free to hack at it further.

source/standards/dns-hosting.html.md.erb
```diff
@@ -1,33 +1,58 @@
...
+[GDS IT Help Desk](https://gdshelpdesk.digital.cabinet-office.gov.uk/helpdesk)
+to request delegation of your subdomain - choose:
+
+* IT Requests
+* Changes - Request & Support
+* DNS
+
+You must follow the [GDS brand guidance][] for services that are accessible by the public but which don't use the GOV.UK brand.
+
+[GDS brand guidance]: https://docs.google.com/document/d/1MXFXP7oMxALiA4t6Gw8XFPksMI-aBjKqO5wCHvnccyo
```
galund, line 58: Turns out this is a link to _my own Google drive_ because we received this guidance by _email_. Gaaaaah.

source/standards/dns-hosting.html.md.erb
galund: Hopefully this is addressed by 242c4e1

source/standards/dns-hosting.html.md.erb
galund: No it wasn't, George confused is all

jonodrew (approved):
Looks good

source/standards/dns-hosting.html.md.erb
```diff
@@ -8,26 +8,56 @@ review_in: 12 months
...
+
+If your service has more complex needs you may alternatively
+[delegate](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/dns-routing-traffic-for-subdomains.html)
+the zone for a subdomain to an AWS account that relates specifically to that environment.
+
+## Internal GDS services
+
+Internal GDS services must be hosted on subdomains of the `digital.cabinet-office.gov.uk` domain.
+
+> This domain is likely to change as part of GDS moving to DSIT.
```
louzoid-gds, line 49: Can we swap DSIT for Department for Science, Innovation and Technology here

louzoid-gds (commented):
Looking good! Thanks for picking up this important change.
I do think this page probably now needs a rename seeing as it's no longer just about DNS records. How about naming it 'Domain names and DNS records for your service' or similar?
Note: if you do rename, you'll also need to change the page name in the main nav

louzoid-gds (approved):
As discussed in the GDS Way forum today, I think we're good to merge these changes now.

PR #1032, opened 2025-06-06
Update performance testing page

louzoid-gds:
It's long overdue an update but most of the content looks decent imo.
I've reworked the Plan test scenarios section to remove the old examples and moved 'Estimate costs' down a bit because you probably can't do that until you've planned your tests.


rpjgds (approved):
LGTM, the goal-based method to plan a tailored scenario rather than one specific example makes sense.

sihugh (approved):
Looks good. I did wonder whether there could be a "consider informing suppliers" bit to augment telling users of the system, but I think it's fine as is.

PR #1036, opened 2025-06-18
Turn on Dead Link checking as part of the test suite for this repo

huwd:
## What

- Turn back on Dead Link checking
- Fix all the broken links we've introduced since 2023
- Allowlist all the ones we can't fix or test appropriately

## Why

Docs contain links, good docs contain links that work. Dead link checking has been part of the test suite here before, looks like in 2023 we turned it off temporarily and never spotted that we needed to turn it back on.

## What does this fix

- Favicon
- External links to sites that block our scraper with a 403
- External links to GDS tools that require auth
- Genuinely broken links

## Continuing problems

Whilst I was doing this snyk.io's site had some sort of outage.
Annoyingly that genuinely meant they weren't handling requests and all our snyk links started failing.

That'll happen here and lead to annoyingly flappy CI sometimes.
Our deploys become as resilient as the internet we link to (and apparently that's not very).

I suggest we merge this and see how it goes.

## Plan B

If that is a problem we just need to add:
```
disable_external: true,
```

That way we get more than we started in that internal dead links can be checked, which is better than not checking at all. We lose the ability to prove any external links work (and this has many) and lose gradual link rot protection (apparently you can get a lot in only two years). But perhaps that's a tradeoff for being able to merge.


huwd:
So main issue right now is this:
image

Favicon not being there is kinda annoying will follow that up.
@romaricpascal / @36degrees - (just to pick on two frontend people I know at random),
Do you know where I can find an official pre-branding update favicon?

Is it your expectation these should be present within a repo like this? or are they all CDN'd?

huwd:
It does seem like a genuine error to me,
I can spot it in the <head> of all our HTML but a look at prod:
https://gds-way.digital.cabinet-office.gov.uk/images/favicon.ico

And it's missing, so might get one in an /images folder

huwd:
@heathd / @galund could I ask someone from EE to give this commit an 👀 
https://github.com/alphagov/gds-way/pull/1036/commits/45510aceb8a77872e02e6b539dee985317459fc1

huwd:
@36degrees might ask for your proof on this, had to swap a link out:
https://github.com/alphagov/gds-way/pull/1036/commits/70ac7960b10060b2422ace28cc900cbaf946d6dc

source/manuals/accessibility.html.md.erb
```diff
@@ -73,7 +73,7 @@ The guide outlines some approaches for testing websites and applications against
 
 You can use [Assistiv Labs](https://assistivlabs.com/) to test your service with JAWS and NVDA.
 
-Contact the [Engineering Enablement team](https://sites.google.com/a/digital.cabinet-office.gov.uk/gds/directorates-and-groups/cto-and-ciso-office/engineering-enablement#h.f8vfqmbkqtgf) to request an account.
+Contact the [Engineering Enablement team](https://sites.google.com/digital.cabinet-office.gov.uk/the-gds-wiki/people-and-teams/product-group/technology-and-security) to request an account.
```
galund:
```suggestion
[Contact the Engineering Enablement team](https://engineering-enablement.gds-reliability.engineering/support.html) to request an account.
```

source/manuals/aws-tagging.html.md.erb
```diff
@@ -57,6 +57,6 @@ In future, we may wish to consider mechanisms such as alerting on untagged resou
 This is based on:
 
  - [AWS documentation and best practices on tagging](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html)
- - [GDS's former tagging strategy](https://docs.google.com/presentation/d/1LHLKPclfrn5KVFrFd2WqyPOYpS6wXklE4Lexb2rJNW0/edit#slide=id.g10d43e3636_2_51)
+ - 🔐 [GDS's former tagging strategy](https://docs.google.com/presentation/d/1LHLKPclfrn5KVFrFd2WqyPOYpS6wXklE4Lexb2rJNW0/) (Google Access required)
```
galund, line 60: I think it might be better to add docs.google.com to https://github.com/alphagov/gds-way/blob/main/source/stylesheets/screen.css.scss#L4

source/manuals/browser-support.html.md.erb
```diff
@@ -9,6 +9,6 @@ owner_slack: '#frontend'
 
 Test your service in [the browsers listed in the Service Manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
 
-You can use [BrowserStack](https://www.browserstack.com/) to test browsers that run on Windows, mobile phones and tablets. Contact the [Engineering Enablement team](https://sites.google.com/a/digital.cabinet-office.gov.uk/gds/directorates-and-groups/cto-and-ciso-office/engineering-enablement#h.f8vfqmbkqtgf) to request an account. You should also test with physical devices whenever possible.
+You can use [BrowserStack](https://www.browserstack.com/) to test browsers that run on Windows, mobile phones and tablets. Contact the [Engineering Enablement team](https://sites.google.com/digital.cabinet-office.gov.uk/the-gds-wiki/people-and-teams/product-group/technology-and-security) to request an account. You should also test with physical devices whenever possible.
```
galund:
```suggestion
You can use [BrowserStack](https://www.browserstack.com/) to test browsers that run on Windows, mobile phones and tablets.
[Contact the Engineering Enablement team](https://engineering-enablement.gds-reliability.engineering/support.html)
to request an account.
You should also test with physical devices whenever possible.
```

source/manuals/working-with-aws-accounts.html.md.erb
```diff
@@ -327,7 +327,7 @@ and ideally conform to the following (except for extenuating circumstances):
 - Read only: Can read things in the account but not perform any side-effects;
             should NOT be able to do things like `ssm:DescribeParameters`
 
-[multiple account security strategy]: https://aws.amazon.com/answers/account-management/aws-multi-account-security-strategy/
+[multiple account security strategy]: https://d1.awsstatic.com/aws-answers/AWS_Multi_Account_Security_Strategy.pdf
```
galund: I think I'd prefer a non-PDF ref - does this one look ok?

```suggestion
[multiple account security strategy]: https://docs.aws.amazon.com/whitepapers/latest/organizing-your-aws-environment/organizing-your-aws-environment.html
```

source/standards/logging.html.md.erb
```diff
@@ -89,8 +89,8 @@ You should ensure that sensitive information, such as query parameters containin
 In order to allow for rich querying of log data you should ensure that your logs
 are in a structured format.
 
-[Splunk] can automatically parse many common types of structured data such as CSV, JSON, and XML. 
-A range of add-ons can be found on [Splunkbase](https://splunkbase.splunk.com/) to parse data from commonly used technologies such as AWS, Azure, Palo Alto Firewalls, and more. 
+🔐[Splunk] - (auth required) can automatically parse many common types of structured data such as CSV, JSON, and XML.
```
galund, line 92: Again I think we can handle this with the existing CSS thang. If you like. I can see the value in the unicode approach, this is just a consistency point that you can take or leave.

source/manuals/accessibility.html.md.erb
huwd: Thank you!

source/manuals/aws-tagging.html.md.erb
huwd: I covered that in this commit here:
https://github.com/alphagov/gds-way/pull/1036/commits/0818b17f4b8bbeea347b0814ac7d209db7f62d0d

Happy to chat more if you disagree, but that's a very wide net to cast at once.

huwd: oooh wait I see, you mean there's an existing CSS hint, huh TIL

galund: Yeah, I'm confident the vast majority of google docs links will be access controlled 🤷 :)

source/standards/logging.html.md.erb
huwd: Yeah I'm on the fence there... I think again either I'll massively expand the range of those sass rules by allow listing entire sites.

If you don't mind I might park this one and return in a later PR to look at how private links are best handled across the site. Can make an issue?

source/manuals/aws-tagging.html.md.erb
huwd: Made an issue will come back later
https://github.com/alphagov/gds-way/issues/1038

Hope that's OK?

source/standards/logging.html.md.erb
huwd: https://github.com/alphagov/gds-way/issues/1038

galund:
You might want to check if the default 'assume_extension' behaviour is problematic. On the Engineering Hub this was preventing us from noticing some incorrect links (because the assumption that the site will be able to route bare URLs to the nearest .html file does not hold for Middleman).

See what I did here (notwithstanding you're calling htmlproofer differently, so the fix would be different). https://github.com/alphagov/ee-engineering-hub/pull/96/files#diff-ee98e028c59b193d58fde56ab4daf54d43c486ae674e63d50ddf300b07943e0fR51

check_links.rb
```diff
@@ -9,23 +9,53 @@
     .filter { |path| path.start_with? "source/" }
     .map { |path| path.delete_prefix("source/").delete_suffix(".erb").delete_suffix(".md") }
 
+  individual_exceptions = [
```
galund, line 12: I don't understand this list, some of these are routable just fine, they aren't all behind a paywall/auth, so what's the decision process for deciding something needs to be in here?

heathd (approved):
brilliant improvement thanks Huw

check_links.rb
heathd: I'm assuming that Huw has added these because of test failures and that seems reasonable/pragmatic to me

galund: I tried following some of the links and they worked :)

huwd: Yeah they'll all work @galund, they're generally individual sites that 403 _even though we have put a user agent in_. So generally I think these are sites with scraper protections on.

The links will work, but the bot will fail

huwd:
Sorry @galund I rudely didn't spot your comment before I merged this.
Just seen it now as I was looking back over.

Interesting stuff, I've gone and checked out the baseurl [assume_extension docs](https://github.com/gjtorikian/html-proofer?tab=readme-ov-file#adjusting-for-a-baseurl).

And we're good, you're right you call this as a flag on the CLI call but we've got ours configured here as an option in the check_directory call. [See here](https://github.com/alphagov/gds-way/blob/main/check_links.rb#L37).

Thanks for flagging though, this whole thing is a bit messy, keen to learn from other static sites round the program

PR #1055, opened 2025-07-28
Podman

galund:
Enable use of [Podman](https://podman.io/)
- Docker alternative that's open source / not subject to Docker's licensing model
- small fix required to port mapping configuration
- separate startup script provided so folks with Docker can still use that


natf-gds:
Checked it out now and testing for you.

natf-gds:
This would be the bash-fu, less readable, but a single place for the launcher commands which avoids duplicating changes in two places.
```
if command -v podman &>/dev/null
then
    echo "Podman detected, launching in podman."
    LAUNCHER=podman
elif command -v docker &>/dev/null
then
    echo "Docker detected, launching in docker."\
    LAUNCHER=docker
else
    echo "Neither podman or docker detected, stopping."
    exit 1
fi
${LAUNCHER} build . --tag gds-way
${LAUNCHER} run --rm -p 4567:4567 -p 35729:35729 -v $(pwd):/usr/src/docs -it gds-way
```

natf-gds:
I wouldn't mind getting other people's opinions on whether the reduction in readability is worth it for the removal of 2 lines that will rarely be updated. 

It's one of those "you can do it, but should you do it" moments 👍 

The risk for podman and docker, is more likely whether they're running rather than whether the CLI is there, but then they 'should' get a different error (I can test tomorrow) which 'should' make it obvious (error codes are never generic and useless after all 😄 ) 

Let me know which way you want to go, we could flag to the EE slack group and ask them for their views? (unsure what process is the best at GDS 😄 )

galund:
If we can summon up support for running the Engineering Hub in Docker/Podman, then yes having EE folks' views on the best way to do that, and doing the same here, seems logical, yes please 👍

natf-gds:
I prompted this at our engineering enablement standup this morning and so far 3 people have voted for what I have called option 2, of the automatic detection between Podman and Docker, but not the bash-fu for the command that gets run, which I'm in agreement with.

Would you prefer I updated your branch, created a new branch, or would you rather do the work? :) 

I'm doing similar work on the engineering-hub right now.

natf-gds:
@davidr-gds 
It's a good shout to have another check, we could do it if you feel it's appropriate (though I'd do the same for podman and docker)

Alternatively the error thrown by startup.sh if podman isn't running is shown below, which probably gets us 'enough', what do you think?

image

galund:
My original readme update mentioned `podman machine start`, but tbh the error message is a pretty good explainer. I think this is ok - the less the podman and the docker versions diverge in the script, the better, IMO.

davidr-gds:
I'm thinking of a user that has both installed, but only docker running. But that's probably unusual - let's keep it as is

davidr-gds (approved):
LGTM

PR #1057, opened 2025-07-30
ENG-896: Fix live reload and enable podman support

levgorbunov1:
Fixes broken livereload in docker container, by injecting the dependency configuration into the tech docs gem, specifying the livereload browser IP.

References:
- https://github.com/middleman/middleman-livereload/issues/26

Also builds on this PR for podman support https://github.com/alphagov/gds-way/pull/1055 to allow live reload and middleman serve to work properly in podman, by monkey patching the listen gem to enable polling mode, as the podman filesystem is peculiar and doesn't emit the necessary signals for watchers like inotifywait to work. 

References: 
- https://github.com/containers/podman/discussions/19430
- https://github.com/guard/listen?tab=readme-ov-file#listen-adapters


levgorbunov1:
> Hey there,
> 
> I tried this in a docker container just now and it didn't seem to reload any changes, could you take a look?
> 
> I've attached the console output of running it using the ./startup.sh script image
> 
> The developer console shows this if that helps: image
> 
> Note, I don't run Docker on my laptop, I run Podman, the only change I've made is to invoke the `podman` binary rather than `docker` in startup.sh to test this.

@natf-gds could you try again by any chance? I have made a few changes to dynamically modify the live reload configuration.

galund (approved):

PR #1071, opened 2025-08-28
Update instructions on deleting AWS accounts

stephengrier:
The Engineering Enablement team has formalised its process for deleting AWS accounts in the GDS AWS organisation. Update the guidance here to make it clear to users how to request an account deletion and what to expect.


galund:
I think this might be better on the Engineering Hub, with a link to that - mirroring how the 'Request an account' section is written?

huwd:
@galund do you want to close this down / update or merge this and iterate later?

huwd:
I've rebased this, which should also mean we don't have the check_links issue.

@stephengrier feel free to merge if you are happy?

PR #1073, opened 2025-09-01
Deleted pages archive space

natf-gds:
**Commit messages:**

- Adding an initial GDS way page to archive deleted pages.
- This will be up for review by the team prior to merging in, and will likely be merged with some real data/page archives, currently it uses a fake VMWare page to show the concept behind the idea.

This pull request is a temporary pull request to enable conversations between those involved in GDS way, and provide a space to iterate on this prior to final merging.

Currently it has 'dummy data' to show how this would fit together and enable it to be searchable.

**Prior to merging:**

1. We should ensure people are content with the page structure (note it's in 'standards' currently which feels non-ideal
2. It needs VMWare removing, and a real set of pages adding.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -9,6 +9,7 @@ title: The GDS Way
         <ul>
             <li><a href='#about-the-gds-way'>About the GDS way</a></li>
             <li><a href='#how-to-add-new-guidance'>How to add new guidance</a></li>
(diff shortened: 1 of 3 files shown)


huwd:
@galund, @mrwilson and @timblair to have a chat with @natf-gds

natf-gds:
Closing after discussion with @galund @timblair @mrwilson - We're going to change tack and do it a different way see: https://github.com/alphagov/gds-way/pull/1093

PR #1089, opened 2025-10-15
Clarify relationship between the GDS Hub and the GDS Way

galund:
We're planning to move step-by-step instructional material onto the Hub - this PR is to get consensus/blessing for starting that journey.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -39,7 +39,7 @@ The GDS Way guides teams to build and operate brilliant, cost-effective digital 
 It documents policy, standards and guidance for the specific technology, tools and processes used by Product Group teams within the Government Digital Service (GDS).
 
 GDS staff can find more detailed instructions for the use of GDS tooling and infrastructure on the
-[GDS Hub](https://engineering-enablement.gds-reliability.engineering/),
+[GDS Engineering Hub](https://engineering-enablement.gds-reliability.engineering/),
 which complements this site.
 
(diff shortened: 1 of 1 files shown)


source/index.html.md.erb
```diff
@@ -36,7 +36,11 @@ title: The GDS Way
 
 The GDS Way guides teams to build and operate brilliant, cost-effective digital services.
 
-It documents the specific technology, tools and processes that the Product Group teams within the Government Digital Service (GDS) use.
+It documents policy, standards and guidance for the specific technology, tools and processes used by Product Group teams within the Government Digital Service (GDS).
+
+GDS staff can find more detailed instructions for the use of GDS tooling and infrastructure on the
+[GDS Hub](https://engineering-enablement.gds-reliability.engineering/),
```
lfdebrux: Might be worth calling it the GDS Engineering Hub to match the page title?

```suggestion
[GDS Engineering Hub](https://engineering-enablement.gds-reliability.engineering/),
```

galund: This is to save time later because it's going to get a rebrand, but shhh nobody knows 😊

galund:
@DaveLee-GDS you're right, the Hub is not available to the public. This means we should be reluctant to move stuff, but we think mostly the instructions we're talking about reflect our very specific responsibility splits and business processes. They don't seem particularly relevant to anyone else, and in the worst case might persuade someone that a practice GDS has is "good" when actually we might be trying to change it. AWS access is the case we're considering, and I think these arguments apply!

DaveLee-GDS:
Ahh i see, thanks George. I think this will be useful actually in helping us consider what we're writing, why and is the audience correct.

PR #1091, opened 2025-10-22
Add Guidance on Handling Instructions for private and internal repos

huwd:
Essentially a change that says we should better annotate the sensitivities of our non-public code, and help all staff understand why, and for how long, those sensitivities persist.

It should help support working in the open by putting the burden of proof on things that are kept out of the public domain. Reducing mistaken, overly cautious or unconsidered removal from public view.

Whilst also driving better and more careful handling of actually sensitive assets, a more informed conversation about security principles and precidents, and better care around how information may be hanlded by those who do have access.

---
Note: This has been loosely adapted from an ADR written by HuwD for the One Login programme.

The format currently doesn't fit "the way" very well and needs a fairly brutal edit for style.

However I wanted to start by including all the information and then crowdsourcing the right tone for the GDS way.

See source here:
https://github.com/govuk-one-login/architecture/blob/main/rfc/0088-handling-instructions.10/22/2025 and yes... i'm aware of the irony that that is private to One Login staff... We're working on it

### DIFF
--- a/source/standards/managing-sensitive-information.html.md.erb
+++ b/source/standards/managing-sensitive-information.html.md.erb
@@ -1,0 +1,164 @@
+---
+title: How to handle sensitive code
+last_reviewed_on: 2025-10-01
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+## Summary
+
+GitHub repositories that are not public should have a `HANDLING_INSTRUCTIONS.md` file at the root explaining why the code is not open and how it should be handled.
+
+This supports [Point 12 of the Service Standard][service-standard-open-code], and ensures decisions about visibility are documented, reviewable and time-limited.
+
+## What you need to do
+
+**Public repositories** do not need handling instructions. Public is the default for GDS code.
+
+**Internal or private repositories** must have a `HANDLING_INSTRUCTIONS.md` committed to the root of the repository, following the format below.
+
+Map your repository visibility to a handling group:
+
+| Visibility | Classification | Handling group |
+|---|---|---|
+| Internal | OFFICIAL-SENSITIVE | `HMG USE ONLY` |
+| Private | OFFICIAL-SENSITIVE | `[ORGANISATION] USE ONLY` or `RECIPIENTS ONLY` |
+
+New repositories should include this file from the start.
+Existing repositories should become compliant over time.
+Business units may choose to enforce this as policy for their area, referencing this page from their team documentation.
+
+### Contractors and outside collaborators
+
+Note that `internal` repositories are readable by all members of the 'Government Digital Service' GitHub enterprise and its various GitHub organizations, including any contractors who have been onboarded as enterprise members.
+
+Where GDS engages contractors (MSP or contingent labour), the programme's business operations team should establish how they are treated with respect to information handling. A common pattern is to treat staff under contract to government as within scope for `HMG USE ONLY`.
+
+Where sensitive matters require tighter boundaries (for example, contract renegotiation), use `RECIPIENTS ONLY` with an explicit description of access restrictions.
+
+### Write access
+
+This guidance covers **read access** (who can see the code).
+Write and admin access should be controlled separately using [CODEOWNERS][code-owners] and infrastructure as code for access management.
+
+## Formatting handling instructions
+
+Create a file called `HANDLING_INSTRUCTIONS.md` in the repository root with this structure:
+
+```md
+---
+classification: OFFICIAL-SENSITIVE
+handling_group: {handling_group}
+last_reviewed_on: yyyy-mm-dd
+review_in: {period}
+responsible_role: {role title}
+responsible_role_contact: {contact}
+---
+
+{Explanation body}
+```
+
+### Field reference
+
+| Field | Value |
+|---|---|
+| `classification` | Always `OFFICIAL-SENSITIVE`.<br> OFFICIAL repos should be public; SECRET/TOP SECRET must not be on GitHub. |
+| `handling_group` | One of: `RECIPIENTS ONLY`, `[ORG NAME] USE ONLY`, `HMG USE ONLY`, `EMBARGOED` (per [Government Security Classifications Policy][handling-instructions]). |
+| `last_reviewed_on` | Date of last review (YYYY-MM-DD). |
+| `review_in` | When to next review, e.g. "6 months" or "12 months". |
+| `responsible_role` | Role title responsible for reviews (not an individual's name). e.g. "Lead Developer, GOV.Uk". |
+| `responsible_role_contact` | A Slack channel or shared inbox. |
+
+### What to write in the body
+
+The body should allow any reader to quickly understand:
+
+- what aspects of this repository are sensitive
+- what concern triggered the deviation from public visibility
+- what might change to trigger re-evaluation
+- who to contact for access exceptions
+
+Write handling instructions on the basis that they will be made public when sensitivity lapses.
+Focus on describing the *causes* of classification, not the sensitive details themselves.
+
+## Making the file visible
+
+Add a banner to the repository's `README.md` so sensitivity is obvious without scrolling:
+
+```md
+> [!IMPORTANT]
+> This repository is OFFICIAL-SENSITIVE. Handle in line with [HANDLING_INSTRUCTIONS.md](/HANDLING_INSTRUCTIONS.md).
+```
+
+## Handling uncertainty
+
+It is fine to make a repository private first and add handling instructions shortly after.
+It is also fine to be explicit about uncertainty:
+
+> We were unsure if business rules in this component were sensitive.
+We made the repo private out of caution and set a 3-month review while we engage stakeholders.
+
+The important thing is to record whatever reasoning was known to the team at the time, clearly enough for a future reviewer to review and act on.
+
+## Reviews and returning to public
+
+### Review cadence
+
+Review handling instructions at least every 12 months.
+Each review should confirm the stated sensitivities still apply and note any new sensitivities introduced since the last review.
+
+Teams should set up reminders using whatever tooling suits them — calendar invites, CI checks, or similar automation.
+
+### Making a private repository public
+
+When sensitivity lapses:
+
+1. Confirm the handling instructions (and any sensitivities they describe) no longer apply.
+2. Audit commits since the repo was last public for unrecorded sensitivities, including in commit messages and metadata.
+3. Remove `HANDLING_INSTRUCTIONS.md` and the README banner.
+4. Set the repository to public.
+
+Where historic commits contain sensitive material, create a new public repository from the current main (without history) and archive the original privately.
+This removes the risk of disclosing historic sensitivities.
+
+## Example handling instruction
+
+```md
+---
+classification: OFFICIAL-SENSITIVE
+handling_group: HMG USE ONLY
+last_reviewed_on: 2024-02-01
+review_in: 12 months
+responsible_role: Lead Developer, Beverage Platform
+responsible_role_contact: #gds-beverages
+---
+
+This component monitors coffee availability in GDS offices.
+
+It displays the source of beans for today's coffee pot. This has been
+reviewed and deemed information that should not be disclosed publicly
+as it could enable a supply chain attack.
+
+This relates to the GDS's established low appetite for coffee-based threats.
+
+Sensitivity is likely to lapse when:
+- the source of beans is made public by another channel
+- the programme's risk appetite changes
+- supply chain attacks in this area are no longer credible
+
+We should review after a year.
+```
+
+## Further reading
+
+- [Government Security Classifications Policy][government-security-classifications-policy]
+- [When code should be open or closed][when-code-should-be-open-or-closed]
+- [Service Manual: making source code open and reusable][service-manual-open]
+
+[service-standard-open-code]: https://www.gov.uk/service-manual/service-standard/point-12-make-new-source-code-open
+[government-security-classifications-policy]: https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-html
+[handling-instructions]: https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-html#handling-instructions
+[when-code-should-be-open-or-closed]: https://www.gov.uk/government/publications/open-source-guidance/when-code-should-be-open-or-closed
+[service-manual-open]: https://www.gov.uk/service-manual/technology/making-source-code-open-and-reusable
+[code-owners]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners


source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+  - [Why we code in the open][2017-why-we-open]
+  - [Benefits of coding in the open][2017-benefits-of-open]
+  - [Don't be afraid to code in the open and how to do it securely][2017-dont-be-afraid]
+  - [Barriers to openness and how to overcome them][2018-barriers-to-open]
+- It's something our colleagues and peers also do as an industry norm across government
+  - [MOJ: How making our deployment code open improved our workflow][moj-open-blog]
+  - [NHS Service Standard:12. Make new source code open][nhs-service-standards]
+
+Potential frictions between working in the open and managing sensitive data are easily mediated with greater definition and description
+of in what way documents are sensitive, and why that decision was made.
```
galund: 'Easily' is maybe an unnecessary take, perhaps:

```suggestion
Any tradeoff between working in the open and managing sensitive data should be mediated through a clear description of how the information is sensitive, and why that decision was made.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+- **Private** repositories are only visible to individuals or teams who have been given explicit access, including through enterprise or organization base permissions.
+
+See [GitHub resource for more information](https://resources.github.com/learn/pathways/administration-governance/essentials/manage-your-repository-visibility-rules-and-settings/).
+
+## Scope
+
+Teams should make use of Handling Instructions to contextualise and aid decision making around who should have read access to code.
+
+Write and admin access to repositories should be more tightly controlled.
+This RFC does not change alter any previous decisions to manage GitHub access via the [configuration repository][config-repo].
```
galund: Remove 'RFC' which this is not :)

```suggestion
Good management of GitHub access controls, such as using infrastructure as code, is a separate concern.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+
+Where targeted controls are required to limit who can review and merge changes, [CODEOWNERS][code-owners] should be used as the mechanism.
+
+## GitHub repositories to include handling instructions
+
+New GitHub repositories in the GOV.UK any GDS organization should follow the requirements of this section.
```
galund:
```suggestion
New GitHub repositories should follow the requirements of this section.
```

This is impossible to reconcile with the thing at the front of the GDS Way about the scope being just Product Group, so we just have to gloss this kind of thing, I think.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+
+2. Use the Service Manual's [making source code open and reusable][service-manual-open] advice to decide to restrict visibility, where there are concerns that require deviation from the Service Standard.
+
+In the second case, they must record handling instructions in a file committed into version control of every non public repository, where:
+
+- Internal is mapped to `OFFICIAL-SENSITIVE` with the handling instruction `HMG USE ONLY`: there is an assumption of broad visibility within the GitHub enterprise, and ability to add outside collaborators across Government.
+- Private is mapped to `OFFICIAL-SENSITIVE` with the handling instruction `[INSERT ORGANISATION(S) NAME] USE ONLY` or `RECIPIENTS ONLY` as appropriate.
+
+Where GDS engages contractors, either as Managed Service Providers (MSP) or Contingent Labour (CL), the Programme Office should establish at the point of contract how they are to be treated with respect to information handling.
+A common pattern may be to assure that a staff member under contract by the government should be treated as within scope for `HMG USE ONLY`.
```
galund:
```suggestion
A common pattern may be to determine that a staff member under contract by the government should be treated as within scope for `HMG USE ONLY`.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+  - `[INSERT ORGANISATION(S) NAME] USE ONLY` - Named organisations with whom the data can be shared
+  - `HMG USE ONLY` - Information should only be shared with other HMG departments; not with external partners.
+  - `EMBARGOED` - Information that is only sensitive for a specific period of time and whose sensitivity will be reduced at the end of that period.
+- `last_reviewed_on` - should be a YYYY-MM-DD formatted date for when a last review was completed and the handling instruction affirmed.
+- `review_in` - a string describing an integer number of months until the next review, for example "6 months" or "120 months"
+- `responsible_role` - the role responsible for reviews, and to whom requests can be made for sharing outside of these terms. Please use role titles and contact points do not name individuals (for example: Lead Developer)
+- `responsible_role_contact` - either a generic shared inbox, or a slack channel where a request for more information can be made
+
+Followed by a body that MUST include:
```
galund:
```suggestion
Followed by a body that must include:
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+The kinds of services GDS runs has changed over time. Broadly we are increasingly engaged in more sentive areas,
+and we run many components that are designed to take burdens off other services. We run sensitive things so others
+do not have to.
+
+In additton, that working in the open has strong tangible benefits to delivery, whilst supporting a key philosophical approach to digital public service delivery.
```
galund:
```suggestion
Furthermore, working in the open has strong tangible benefits to delivery and at the same time supports a key philosophical approach to digital public service delivery.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+This manfests in a range of guidance, cultural norms, government policy and established custom and practice, that encourage a measured approach to openness.
+Examples of thsi can be found in:
```
galund:
```suggestion
Examples of this can be found in:
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+
+> We are publishing code within the period of Pre-election sensitivity. Whilst there is nothing actively concerning in the source, we were uncertain if making these elements public during the run up to the election was allowable under pre-election-period rules or not.
+So we have made it private, with the presumption we'll take the repo back into the public domain after the election
+
+The important aspect is to provide whatever reasoning was generally known to the team, as clearly as is possible to summarise.
+This would allow for any future resource owner, or stakeholders such as security, to quickly review in future.
+
+## Costs, benefits and remediation
+
+If we agree to this approach, there'll be marginally more administration to the act of taking a GitHub repository out of the public domain.
```
galund: This isn't an RFC so the 'if' formulation doesn't really work

```suggestion
This policy has a cost: there's marginally more administration required in order to take a GitHub repository out of the public domain.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+
+Their responsibilities also include:
+
+- auditing and amending handling instructions in repositories within their pod or area of responsibility
+- coming to a collective position on un-owned repositories
+- monitoring on an ongoing basis to ensure handling instructions are added to any repository newly created or taken out of the public domain.
+
+Any final calls will be made by the Head of Engineering.
+
+This should be completed within six months following the date of merging this RFC.
```
galund: This section no longer really works - we need a slightly different plan and an understanding of what might be in scope, and what across the broad reach of alphagov is actually not that realistic.

Perhaps something much more vague like

```suggestion
Heads of engineering and G6 lead technologists may wish to remediate existing repositories in accordance with this policy, for example by:

- auditing repositories that are not currently in the public domain
- adding a `HANDLING_INSTRUCTIONS.md`
- capturing any uncertainty in those the handling instructions, and reducing the review period accordingly

Lead technologists also have ongoing responsibilities to:

- audit and amend handling instructions in repositories for which they are responsible 
- participate in cross-GDS efforts to find and remediate un-owned repositories
- monitor compliance with this policy with respect to repositories that are newly created or are taken out of the public domain
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,404 @@
...
+### Government context
+
+As Civil Servants, or contractors to the Civil Service, we are bound to follow the [Government Security Classifications Policy][government-security-classifications-policy].
+
+Whilst the total landscape of Government Information Handling is complex, at GDS our subset is simpler.
+
+Neither our staff, nor our systems are generally rated to handle [SECRET][definition-of-secret] or [TOP SECRET][definition-of-top-secret] information.
+Where this does occur it will be by exception, and generally mediated through staff with Developed Vetting around specific tasks.
+Examples might be information gathering in relation to specific NCSC concerns.
+These areas are treated as being outside the scope of this RFC.
```
galund:
```suggestion
These areas are not in scope for this policy.
```

galund (commented):
starter for 10

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+[github-repos]: https://github.com/orgs/govuk-one-login/repositories
+[github-repos-private]: https://github.com/orgs/govuk-one-login/repositories?q=visibility%3Aprivate
+[github-repos-internal]: https://github.com/orgs/govuk-one-login/repositories?q=visibility%3Ainternal
+[github-repos-public]: https://github.com/orgs/govuk-one-login/repositories?q=visibility%3Apublic
+[github-visibility]: https://docs.github.com/en/enterprise-cloud@latest/repositories/creating-and-managing-repositories/about-repositories#about-repository-visibility
+[github-visibility-resource]: https://resources.github.com/learn/pathways/administration-governance/essentials/manage-your-repository-visibility-rules-and-settings/
+[when-code-should-be-open-or-closed]: https://www.gov.uk/government/publications/open-source-guidance/when-code-should-be-open-or-closed
+[service-manual-open]: https://www.gov.uk/service-manual/technology/making-source-code-open-and-reusable
+[code-owners]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+[config-repo]: https://github.com/govuk-one-login/configuration
```
huwd:
```suggestion
[config-repo]: https://github.com/govuk-one-login/configuration

```

New lines are good, for reasons I can never remember but have come to dogmatically believe

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+  - [Coding in the open][2012-coding-in-open]
+  - [Why we code in the open][2017-why-we-open]
+  - [Benefits of coding in the open][2017-benefits-of-open]
+  - [Don't be afraid to code in the open and how to do it securely][2017-dont-be-afraid]
+  - [Barriers to openness and how to overcome them][2018-barriers-to-open]
+- It's something our colleagues and peers also do as an industry norm across government
+  - [MOJ: How making our deployment code open improved our workflow][moj-open-blog]
+  - [NHS Service Standard:12. Make new source code open][nhs-service-standards]
+
+Any tradeoff between working in the open and managing sensitive data should be mediated through a clear description of how the information is sensitive, and why that decision was made.
```
natf-gds: Is it worth mentioning something around the fact we don't want people to default to 'private' as it's perceived to be an easier conversation than selecting 'public' and having this `handling-instructions.md` helps us ensure we have the same effort involved in determining open or closed?

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+title: How to handle sensitive code
+last_reviewed_on: 2025-10-01
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+The kinds of services GDS runs has changed over time. Broadly we are increasingly engaged in more sentive areas,
+and we run many components that are designed to take burdens off other services. We run sensitive things so others
+do not have to.
```
natf-gds:
```suggestion
The kinds of services GDS runs has changed over time. Broadly we are increasingly engaged in more sensitive areas,
and we run many components that are designed to take burden off other services.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+  - [NHS Service Standard:12. Make new source code open][nhs-service-standards]
+
+Any tradeoff between working in the open and managing sensitive data should be mediated through a clear description of how the information is sensitive, and why that decision was made.
+
+The theory and practice is very similar to GDS's standard approach to [writing good commit messages](/standards/source-code/working-with-git.html#commit-messages).
+
+In particular better metadata applied to sensitive information should help with the tasks of establishing:
+- the reasons or history of choices that led to something being made public or private
+- if those reasons remain valid or have changed with risk appetite or context
+- how any decisions may or may not serve as useful precedent to help expedite future decision making on similar topics.
```
natf-gds:
```suggestion
- The reasons and history of choices, that led to something being made public or private.
- If those reasons remain valid or have changed with risk appetite or context.
- How any decisions may or may not serve as useful precedent to help expedite future decision making on similar topics.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+
+Proper documentation and information availability around sensitivity and its causes when applied to our documentation and code helps staff
+more quickly navigate handling decisions. It ensures that we keep close to the service standard and work in the open, and understand where there are legitimate excpetions
```
natf-gds:
```suggestion
Proper documentation and information around sensitivity of code and documentation helps staff
quickly navigate handling decisions. It also ensures that we keep close to the service standard of working in the open, whilst understanding where there are legitimate exceptions
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+### Government context
+
+As Civil Servants, or contractors to the Civil Service, we are bound to follow the [Government Security Classifications Policy][government-security-classifications-policy].
+
+Whilst the total landscape of Government Information Handling is complex, at GDS our subset is simpler.
+
+Neither our staff, nor our systems are generally rated to handle [SECRET][definition-of-secret] or [TOP SECRET][definition-of-top-secret] information.
+Where this does occur it will be by exception, and generally mediated through staff with Developed Vetting around specific tasks.
+Examples might be information gathering in relation to specific NCSC concerns.
+These areas are not in scope for this policy.
```
natf-gds: I wonder if we can simplify to something like

```suggestion
Whilst the total landscape of Government Information Handling is complex, at GDS our subset is simpler as we primarily operate at OFFICIAL and OFFICIAL-SENSITIVE, higher classifications are not in scope for this policy
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+
+All our information will be treated as [OFFICIAL][definition-of-OFFICIAL] as it is generated as part of Civil Service duties.
+In most cases there are limited to no negative consequences if OFFICIAL information is compromised or disclosed.
```
natf-gds: Do we need this line?

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+---
+
+# <%= current_page.data.title %>
+
+The kinds of services GDS runs has changed over time. Broadly we are increasingly engaged in more sentive areas,
+and we run many components that are designed to take burdens off other services. We run sensitive things so others
+do not have to.
+
+Furthermore, working in the open has strong tangible benefits to delivery and at the same time supports a key philosophical approach to digital public service delivery.
+This manfests in a range of guidance, cultural norms, government policy and established custom and practice, that encourage a measured approach to openness.
```
natf-gds:
```suggestion
This manifests in a range of guidance, cultural norms, government policy and established custom and practice, that encourage a measured approach to openness.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+This should be evaluated against both the handling instructions that have lapsed and any other areas of program sensitivity.
+Any review should also take into account the HANDLING_INSTRUCTIONS.md itself.
+See [The Sensitivity of HANDLING_INSTRUCTIONS.md](#the-sensitivity-of-handling_instructions.md).
+
+The cost of doing such a review could easily become disproportionate, this should be mitigated in the following ways:
+
+#### How do I know if sensitivity has lapsed?
+
+HANDLING_INSTRUCTIONS.md establish a responsible role.
+Most often this will be the pereson directly involved in the decision to take the repository private or internal.
```
natf-gds:
```suggestion
Most often this will be the person directly involved in the decision to take the repository private or internal.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+[visibility][github-visibility]
+settings:
+
+- **Public** repositories are visible to the world and are primarily used for open source content.
+- **Internal** repositories are read-accessible to all full members of your enterprise account (GDS, CDDO, CO:Di.AI and other parts of Cabinet Office but not to outside collaborators).
+- **Private** repositories are only visible to individuals or teams who have been given explicit access, including through enterprise or organization base permissions.
+
+See [GitHub resource for more information](https://resources.github.com/learn/pathways/administration-governance/essentials/manage-your-repository-visibility-rules-and-settings/).
+
+## Scope
```
natf-gds: Should we put the classification scope in here from the context section above? 

Is it worth moving the background and context to lower on the page so that the guidance is at the top, the reasoning for the guidance can be lower in the page e.g.
TLDR - Do this
Long read - Here's the reasons why

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+
+Teams should make use of Handling Instructions to contextualise and aid decision making around who should have read access to code.
+
+Write and admin access to repositories should be more tightly controlled.
+Good management of GitHub access controls, such as using infrastructure as code, is a separate concern.
```
natf-gds: Open question - I don't mind the choice/decision.

I'm wondering if this page needs to split? e.g. "How to handle sensitive code" with sub pages
* Good practices to secure the repos
* Good practices to mark the sensitivities (handling-instructions.md)
* A.N.Other aspect. 

At the moment it feels like we're starting to blur multiple solutions to improve a topic into one page.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+
+Where targeted controls are required to limit who can review and merge changes, [CODEOWNERS][code-owners] should be used as the mechanism.
+
+## GitHub repositories to include handling instructions
+
+New GitHub repositories should follow the requirements of this section.
+Existing repositories will become compliant over time (see [remediation](#costs-benefits-and-remediation)).
```
natf-gds: I know the aim of this, but I do worry that it will be a reason not to do this.

Could we instead put something like new repos do it from the start, existing repos if under active development implement this in 12 months?

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+
+```md
+---
+classification: OFFICIAL-SENSITIVE
+handling_group: {handling_group}
+last_reviewed_on: yyyy-mm-dd
+review_in: {period for review}
+responsible_role: {role}
+responsible_role_contact: {contact_details}
+---
```
natf-gds: Could we enforce at least some reasoning to be part of the minimum and in this template.

I know you put below "Followed by a body" but I wonder if the template could have EXPLANATION_BODY as one of the 'fields'

```suggestion
responsible_role_contact: {contact_details}
---
EXPLANATION_BODY
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+- A contact point for that role (slack channel or email inbox)
+
+Depending on the context the following should also be included:
+
+- a link to the [central guidance][when-code-should-be-open-or-closed] if (and only if) the decision is in complete conformance with that policy
+- an outline of, or link to centrally held, threat assessments that help individuals understand the nature of the threat
+- more detail on key assumptions which were understood to be true at the time, but may change triggering re-evaluation later
+- where there are multiple individuals in a role (for example "Lead Developer" or "Security Architect"), clarify with an area of responsibility. For example "Lead Developer for Identity product teams"
+
+#### The Sensitivity of HANDLING_INSTRUCTIONS.md
```
natf-gds: Could we mandate that the file needs to be OFFICIAL? 

Then you could scrape all across GitLab and crunch data/reasonings etc without having to worry about the sensitivity of the data within this.

Maybe that's an implementation detail for us though.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,397 @@
...
+If a review acts as a rolling audit it can ensure HANDLING_INSTRUCTIONS.md remains accurate, but also provide confidence checkpoints to reduce the cost of an audit when a repository is being considered for release into the public domain.
+
+#### Split the repository along sensitive lines
+
+Where sensitivity has lapsed and a repository can now be made public, but uncertainty remains about the version history and accuracy of the HANDLING_INSTRUCTIONS.md.
+A decision can be made to split a suitable repository. This could be in the format of:
+
+- Functional changes, abstracting a component off into a new public repository and distributing it from there.
+- Partial transfer, for instance with documentation repositories, transferring files where this is confidence they are not OFFICIAL-SENSITIVE to a new repo and leaving ones where uncertainty remains.
+- Full transfer without version history, full transfer of the state of `main` to a new repository, whilst preserving the old archive as private for `git blame` purposes.
```
natf-gds: Is it worth mentioning the reason you may do this is to remove any sensitivities from the history?

natf-gds (commented):
I'm not precious about any of these, happy for them to be accepted or rejected.

source/standards/managing-sensitive-information.html.md.erb
huwd: Yeah interesting, I guess in the GDS way I'm happy to be more explicit about the ploy I'm pulling here.
I want to put the burden of proof on those _going_ private. I'd been a bit more cagey about framing it that way in other forums. But I think you are right in being more explicit here.

it shouldn't be easier to go `private` and we're making it not so, as we're saying the private stuff comes with an admin burden of describing why, how long for and over what matters, plus periodic reviews.

source/standards/managing-sensitive-information.html.md.erb
huwd: :D I'm happy to put a 12 month deadline on this. But feel like that's a question I might need some backing from the likes of @heathd or @davidr-gds on.

There might also be a @galund style comment here that we might want to be careful about having the GDS way slip into policy away from best practice.

Perhaps another version is to drop all this and have a non normative:
> This should apply to all GDS repos

And then handle the transition elsewhere?

source/standards/managing-sensitive-information.html.md.erb
huwd: 🤔 So I think, given you _can_ have these in private repos there may be some of these we don't release by default without another step.

That said, I'd be open to aggregating these and ensuring the contents stays OFFICIAL?

it might come down to some nuance of how worried are we about detail, or the statement that detail exists.
I think handling instructions should be able to describe the sensitivities enclosed "Fraud rules are sensitive because...", "these things have a security impact" generally

We might run into issue with embargoed stuff... like if we're saying:
> this is private because the minister hasn't announced {x}

it'd be a bit awkward to try and find a way to write these that didn't mention what that thing was that's unannounced?

Perhaps we should say:
> HANDLING_INSTRUCTIONS.md should focus on describing the causes for classification rather than the outline the actual sensitivity in detail. They should be more generally disclose-able and presumed to be OFFICIAL where possible. Where there is cause for the instructions to be more sensitive (admitting existence of a piece of work for instance) that should be stated in the handling instructions

source/standards/managing-sensitive-information.html.md.erb
huwd: ah I'd hoped that was implicit?

it's a lot of faff to transfer, so hoped it'd just be about removing those sensitivities.
Do you think i need to make that clearer for some audiences?

source/standards/managing-sensitive-information.html.md.erb
huwd: I'll prep a change

source/standards/managing-sensitive-information.html.md.erb
huwd: no probably not
```suggestion
```

source/standards/managing-sensitive-information.html.md.erb
huwd: Yeah I think @davidr-gds was making a similar point, perhaps about separating out some of the "whats" and "whys" of this too.

I'm open to it, just might need a bit of consensus on how best to chunk it up 🤔

source/standards/managing-sensitive-information.html.md.erb
galund: FWIW I'm cool with the idea of documenting policy in the GDS Way, but there has to be some agreement that policy might be enforced, and it has to be really clear where we're going there.

Having deadlines in the GDS Way definitely doesn't seem right, the place for deadlines is in some kind of working document between heads of engineering or something. But I doubt it's feasible for GDS right now to make any claims that this "will" happen.

```suggestion
Teams should aim for existing repositories to become compliant over time (see [remediation](#costs-benefits-and-remediation)).
Lead technologists may enforce this as policy for their area of responsibility, in which case they must reference this page from their team documentation.
```

galund: (I should have written 'I agree with Huw' and left it at that, probably.)

source/standards/managing-sensitive-information.html.md.erb
natf-gds: We could always get this in and then chunk/split it out #agile

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
+---
+title: How to handle sensitive code
+last_reviewed_on: 2025-10-01
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+The kinds of services GDS runs has changed over time. Broadly we are increasingly engaged in more sensitive areas,
```
claireashworth:
```suggestion
The services GDS runs has changed over time. We are increasingly engaged in more sensitive areas,
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
...
+and we run many components that are designed to take burden off other services.
```
claireashworth:
```suggestion
and run many components designed to take some burden off other services.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
...
+
+Furthermore, working in the open has strong tangible benefits to delivery and at the same time supports a key philosophical approach to digital public service delivery.
```
claireashworth:
```suggestion
Also, working in the open benefits digital public service delivery as well as supporting an important philosophical approach to delivery.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
...
+This manifests in a range of guidance, cultural norms, government policy and established custom and practice, that encourage a measured approach to openness.
```
claireashworth:
```suggestion
This results in a range of guidance, cultural norms, government policy and established practices, that encourage a measured approach to openness.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
...
+Examples of this can be found in:
+
+- Point 12 of the [Service Standard][service-standard] ([detail here][service-standard-open-code])
```
claireashworth: I recommend having this bullet say "Point 12 of the [service standard – make new source code open](and add link only to point 12)" as it makes the link surface area and context meet accessibility criteria

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
...
+---
+
+# <%= current_page.data.title %>
+
+The kinds of services GDS runs has changed over time. Broadly we are increasingly engaged in more sensitive areas,
+and we run many components that are designed to take burden off other services.
+
+Furthermore, working in the open has strong tangible benefits to delivery and at the same time supports a key philosophical approach to digital public service delivery.
+This manifests in a range of guidance, cultural norms, government policy and established custom and practice, that encourage a measured approach to openness.
+Examples of this can be found in:
```
claireashworth:
```suggestion
Established examples of this are in:
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,391 @@
...
+
+- Point 12 of the [Service Standard][service-standard] ([detail here][service-standard-open-code])
+- Well established in the [GDS Way][gds-way-open-standards] and operational practice around GDS (Including on Critical National Infrastructure components)
```
claireashworth:
```suggestion
- the [GDS Way][gds-way-open-standards] and in operational practice around GDS (Including on Critical National Infrastructure components)
```
I suggested a change to the lead in line for these bullets which included the word 'established', so if you accept the new lead in line then some of the bullets need a little amending as well.

source/standards/managing-sensitive-information.html.md.erb
natf-gds: The embargo aspect is important, I didn't spot that (OFFICIAL but not for public yet) happy with your suggestion :)

source/standards/managing-sensitive-information.html.md.erb
natf-gds: I think I would make it clear just in case, the impact if someone doesn't think about the implicit aspect could be poor.

source/standards/managing-sensitive-information.html.md.erb
natf-gds: I'm fine with not having a deadline if that's not the done thing, I think we should however assume that a lot of repos may not adopt this without a push (maybe leads can push separately) 

I like your update though @galund

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+The services GDS runs has changed over time. We are increasingly engaged in more sensitive areas,
+and run many components designed to take some burden off other services.
+
+Furthermore, working in the open has strong tangible benefits to delivery and at the same time supports a key philosophical approach to digital public service delivery.
+This manifests in a range of guidance, cultural norms, government policy and established custom and practice, that encourage a measured approach to openness.
+Established examples of this are in:
+
+- Point 12 of the [Service Standard][service-standard] ([detail here][service-standard-open-code])
+- Well established in the [GDS Way][gds-way-open-standards] and operational practice around GDS (Including on Critical National Infrastructure components)
+- A part of the [DI Source Control Management Policy][source-control] (see overview)
```
claireashworth:
```suggestion
- part of the [DI Source Control Management Policy][source-control] (see overview)
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+- Something we have talked about [publicly][2012-coding-in-open] and frequently as a benefit to GDS:
```
claireashworth:
```suggestion

We also have frequently published content discussing[coding in the open][2012-coding-in-open] and the benefits to GDS:
```
Suggest changing to a lead in line so the bullets can come up a level and avoid having nested bullet points.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+  - [Coding in the open][2012-coding-in-open]
```
claireashworth: Instead of commenting on each one, I'm suggesting bringing the bullet points in this list up a level

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+  - [Why we code in the open][2017-why-we-open]
+  - [Benefits of coding in the open][2017-benefits-of-open]
+  - [Don't be afraid to code in the open and how to do it securely][2017-dont-be-afraid]
+  - [Barriers to openness and how to overcome them][2018-barriers-to-open]
+- It's something our colleagues and peers also do as an industry norm across government
```
claireashworth:
```suggestion

In other departments, our colleagues and peers also code in the open as an industry norm across government:
```
Also trialing this as a lead in line for the next set of bullet points.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+  - [MOJ: How making our deployment code open improved our workflow][moj-open-blog]
```
claireashworth: So if you accept the new lead in lines, these 2 bullet points would go up a level as well.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+  - [NHS Service Standard:12. Make new source code open][nhs-service-standards]
+
+Any tradeoff between working in the open and managing sensitive data should be mediated through a clear description of how the information is sensitive, and why that decision was made.
```
claireashworth:
```suggestion
You should record your decision to code in the open using a clear description to explain which code is open and how you manage sensitive data.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,392 @@
...
+# <%= current_page.data.title %>
+
+The services GDS runs has changed over time. We are increasingly engaged in more sensitive areas,
+and run many components designed to take some burden off other services.
+
+Also, working in the open benefits digital public service delivery as well as supporting an important philosophical approach to delivery.
+This results in a range of guidance, cultural norms, government policy and established practices, that encourage a measured approach to openness.
+Established examples of this are in:
+
+- Point 12 of the [Service Standard][service-standard] ([detail here][service-standard-open-code])
```
huwd:
```suggestion
- Point 12 of the [service standard – make new source code open][service-standard-open-code]
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,394 @@
...
+If a review acts as a rolling audit it can ensure HANDLING_INSTRUCTIONS.md remains accurate, but also provide confidence checkpoints to reduce the cost of an audit when a repository is being considered for release into the public domain.
+
+#### Split the repository along sensitive lines
+
+Where sensitivity has lapsed and a repository can now be made public, but uncertainty remains about the version history and accuracy of the HANDLING_INSTRUCTIONS.md.
+A decision can be made to split a suitable repository. This could be in the format of:
+
+- Functional changes, abstracting a component off into a new public repository and distributing it from there.
+- Partial transfer, for instance with documentation repositories, transferring files where this is confidence they are not OFFICIAL-SENSITIVE to a new repo and leaving ones where uncertainty remains.
+- Full transfer without version history, full transfer of the state of `main` to a new repository, whilst preserving the old archive as private for `git blame` purposes.
```
huwd:
```suggestion
- Full transfer without version history, full transfer of the state of `main` to a new repository, whilst preserving the old archive as private for `git blame` purposes.

Choice between these options will depend on the nature of the sensitivity we initially guarded against.
It will also depend on if there are historic sensitivites. What is important is to ensure no historic sensitivites can be found in either the content of the commit history, or metadata like commit messages.
```

source/standards/managing-sensitive-information.html.md.erb
huwd: thank you, made a change!

source/standards/managing-sensitive-information.html.md.erb
huwd: amended these here: https://github.com/alphagov/gds-way/pull/1091/commits/57a1b2b1c764191c2b81f07fc56ca8bb76d11562

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,399 @@
...
+
+- **Public** repositories are visible to the world and are primarily used for open source content.
+- **Internal** repositories are read-accessible to all full members of your enterprise account (GDS, CDDO, CO:Di.AI and other parts of Cabinet Office but not to outside collaborators).
+- **Private** repositories are only visible to individuals or teams who have been given explicit access, including through enterprise or organization base permissions.
+
+See [GitHub resource for more information](https://resources.github.com/learn/pathways/administration-governance/essentials/manage-your-repository-visibility-rules-and-settings/).
+
+## Scope
+
+Teams should make use of Handling Instructions to contextualise and aid decision making around who should have read access to code.
```
claireashworth:
```suggestion
Teams should use handling instructions to contextualise and aid decision making about who should have read access to code.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,399 @@
...
+
+## GitHub repositories to include handling instructions
+
+New GitHub repositories should follow the requirements of this section.
+Teams should aim for existing repositories to become compliant over time (see [remediation](#costs-benefits-and-remediation)).
+Lead technologists may enforce this as policy for their area of responsibility, in which case they must reference this page from their team documentation.
+
+GitHub repository visibility must be:
+
+- aligned with the Service Standard, that is, public by default
```
claireashworth:
```suggestion
- aligned with the Service Standard, which requires code be public by default with caveats for things like secrets
```

claireashworth: I don't know if this needs to be in this bullet point, but felt it might be an appropriate inclusion.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,399 @@
...
+1. Presume a default classification of `OFFICIAL - FOR PUBLIC RELEASE` - public visibility - for all source code generated by the programme, with no handling instructions recorded in this case.
+
+2. Use the Service Manual’s [making source code open and reusable][service-manual-open] advice to decide to restrict visibility, where there are concerns that require deviation from the Service Standard.
+
+In the second case, they must record handling instructions in a file committed into version control of every non public repository, where:
+
+- Internal is mapped to `OFFICIAL-SENSITIVE` with the handling instruction `HMG USE ONLY`: there is an assumption of broad visibility within the GitHub enterprise, and ability to add outside collaborators across Government.
+- Private is mapped to `OFFICIAL-SENSITIVE` with the handling instruction `[INSERT ORGANISATION(S) NAME] USE ONLY` or `RECIPIENTS ONLY` as appropriate.
+
+Where GDS engages contractors, either as Managed Service Providers (MSP) or Contingent Labour (CL), the Programme Office should establish at the point of contract how they are to be treated with respect to information handling.
```
claireashworth: is programme office still a commonly used term in GDS? or is this a specific team?

galund:
```suggestion
Where GDS engages contractors, either as Managed Service Providers (MSP) or Contingent Labour (CL), the relevant programme's business operations team should establish how they are to be treated with respect to information handling.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,399 @@
...
+- [Benefits of coding in the open][2017-benefits-of-open]
+- [Don’t be afraid to code in the open and how to do it securely][2017-dont-be-afraid]
+- [Barriers to openness and how to overcome them][2018-barriers-to-open]
+
+In other departments, our colleagues and peers also code in the open as an industry norm across government:
+
+- [MOJ: How making our deployment code open improved our workflow][moj-open-blog]
+- [NHS Service Standard:12. Make new source code open][nhs-service-standards]
+
+You should record your decision to code in the open using a clear description to explain which code is open and how you manage sensitive data.
```
timblair: This is backwards. If the default position is for code to be open (which it is, based on the Service Standard etc), then you should be recording any decisions for code that is not open, not the other way around. Forcing justification for open code just makes it more likely that code will be left closed.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,399 @@
...
+
+2. Use the Service Manual’s [making source code open and reusable][service-manual-open] advice to decide to restrict visibility, where there are concerns that require deviation from the Service Standard.
+
+In the second case, they must record handling instructions in a file committed into version control of every non public repository, where:
+
+- Internal is mapped to `OFFICIAL-SENSITIVE` with the handling instruction `HMG USE ONLY`: there is an assumption of broad visibility within the GitHub enterprise, and ability to add outside collaborators across Government.
+- Private is mapped to `OFFICIAL-SENSITIVE` with the handling instruction `[INSERT ORGANISATION(S) NAME] USE ONLY` or `RECIPIENTS ONLY` as appropriate.
+
+Where GDS engages contractors, either as Managed Service Providers (MSP) or Contingent Labour (CL), the Programme Office should establish at the point of contract how they are to be treated with respect to information handling.
+A common pattern may be to determine that a staff member under contract by the government should be treated as within scope for `HMG USE ONLY`.
```
timblair: This is an important a subtle point that is easily overlooked. E.g. a programme bringing in an MSP/contractor is responsible not only for onboarding to relevant local services, but vetting that individual for access to content generated by _other parts_ of the organisation. While this is obvious and relatively easily controlled for `private` repos (they need access explicitly granted), it's _not_ obvious for `internal` repos, due to the enterprise-level visibility. How do we ensure that this is understood and taken into consideration?

galund: I think that this is okay as long as we make it clear that such contractors will have access to OFFICIAL-SENSITIVE HMG USE ONLY data. I think GDS needs (and tbh probably has) fairly standard looking contracts that bake this in already – we aren't asking directorates engaging contractors to review everyone else's `internal` stuff, we're just reminding them that GDS and DSIT data (and other data they may come across from other departments) must be handled with care, and they need to be aware of / trained with regards our (and Government) policy in this space.

source/standards/managing-sensitive-information.html.md.erb
galund: This is not the original wording we had, it used to read

> Any tradeoff between working in the open and managing sensitive data should be mediated through a clear description of how the information is sensitive, and why that decision was made.

I think @claireashworth was concerned about readability, but I agree the new sentence is introducing a strong requirement that was never intended. This section of the document isn't really supposed to be introducing prescriptive clauses at all, it's supposed to be giving context and maybe describing goals. So perhaps instead:

```suggestion
There are tradeoffs between working in the open and managing sensitive data: our goal is to make these explicit so that we can improve the quality of our decisions.
```

source/standards/managing-sensitive-information.html.md.erb
galund: The conversation on this point has moved [here](https://github.com/alphagov/gds-way/pull/1091#discussion_r2546897667) so I'm gonna resolve this one.

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,399 @@
+---
+title: How to handle sensitive code
+last_reviewed_on: 2025-10-01
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
```
davidr-gds, line 8: *I* appreciate the 4000 words of context-setting, reasoning and process. But thinking of the majority of users who just want to write a HANDLING_INSTRUCTIONS.md file, how about we add a quick summary like this, and link straight through to the details of the file format?

```suggestion

## Summary

GitHub repos that are not 'Public' should have a `HANDLING_INSTRUCTIONS.md` file at the root.

Key info in that file:
* Classification: 'OFFICIAL-SENSITIVE' (if it was only OFFICIAL, then the repo shoud probably be Public)
* Handling instructions:
    * Internal repos should be `HMG USE ONLY`
    * Private repos should be `[INSERT ORGANISATION(S) NAME] USE ONLY` or `RECIPIENTS ONLY` as appropriate.
* Explanation: what's the sensitivity of this repo, that explains the handling instructions

Format details: [Standard format for handling instructions](#standard-format-for-handling-instructions)

## Introduction

```

galund: Yes I agree. Actually how about we move all/most of the context-setting stuff into an ADR?

huwd: very keen, where to ADRs live @galund?

galund: Here I think for this: https://github.com/alphagov/github-enterprise-management/tree/main/decisions

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,414 @@
...
+The only sensitive materials committed to private repos should be those described in the HANDLING_INSTRUCTIONS.md.
+Any new sources of sensitivity should be added to the HANDLING_INSTRUCTIONS.md along with the commit.
+
+For any materials not detailed in HANDLING_INSTRUCTIONS.md expectations of what we commit should remain aligned between public and private repos.
+For instance, we would not expect secrets or access tokens to be committed to a private repo, without good reasoning described in HANDLING_INSTRUCTIONS.
+
+This will ensure the HANDLING_INSTRUCTIONS, and their version history, will remain a guide to the repository’s current and historic sensitivities.
+Allowing for a speedier audit upon a decision to return to the public domain.
+
+#### Regular reviews
```
jonnyandrew: Is there any advice on available tooling to track and automate review reminders?

I had this question come up on a handling instructions PR I raised: https://github.com/govuk-one-login/mobile-id-check-android-sdk/pull/392#pullrequestreview-3686422198

huwd:
Huw will re-draft this in the next two weeks...

huwd:
re-drafted, ready for it to be reviewed and go out

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,164 @@
...
+title: How to handle sensitive code
+last_reviewed_on: 2025-10-01
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+## Summary
+
+GitHub repositories that are not public must have a `HANDLING_INSTRUCTIONS.md` file at the root explaining why the code is not open and how it should be handled.
```
galund: This 'must' is inconsistent or at least confusing with respect to the 'Business units may choose to enforce this as policy for their area' thing. Maybe we should just make it a 'should' for now.

huwd: Thats a fair cop. How do you feel about a slightly more judgy SHOULD? All the flexibility of MAY but with an edge of moral impropriety if you don't 😂

huwd:
```suggestion
GitHub repositories that are not public should have a `HANDLING_INSTRUCTIONS.md` file at the root explaining why the code is not open and how it should be handled.
```

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,164 @@
...
+| Internal | OFFICIAL-SENSITIVE | `HMG USE ONLY` |
+| Private | OFFICIAL-SENSITIVE | `[ORGANISATION] USE ONLY` or `RECIPIENTS ONLY` |
+
+New repositories should include this file from the start.
+Existing repositories should become compliant over time.
+Business units may choose to enforce this as policy for their area, referencing this page from their team documentation.
+
+### Contractors and outside collaborators
+
+Note that `internal` repositories are readable by all members of the GitHub enterprise, including any contractors who have been onboarded as enterprise members.
```
davidr-gds:
```suggestion
Note that `internal` repositories are readable by all members of the 'Government Digital Service' GitHub enterprise and its various GitHub organizations, including any contractors who have been onboarded as enterprise members.
```

source/standards/managing-sensitive-information.html.md.erb
davidr-gds: Agreed, "should" is better here, and L19, to align with the wording of the [Government Security Classifications Policy](https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-html) when it says: "The -SENSITIVE marking should be applied to OFFICIAL  information that is not intended for public release".

And each business area can choose how strong to follow it

huwd: I've made that change!

source/standards/managing-sensitive-information.html.md.erb
```diff
@@ -0,0 +1,164 @@
...
+
+GitHub repositories that are not public should have a `HANDLING_INSTRUCTIONS.md` file at the root explaining why the code is not open and how it should be handled.
+
+This supports [Point 12 of the Service Standard][service-standard-open-code], and ensures decisions about visibility are documented, reviewable and time-limited.
+
+## What you need to do
+
+**Public repositories** do not need handling instructions. Public is the default for GDS code.
+
+**Internal or private repositories** must have a `HANDLING_INSTRUCTIONS.md` committed to the root of the repository, following the format below.
```
davidr-gds, line 19: "should" instead of "must"?

heathd: I vote for "must" -- rationale: 

1.  if we use "should" an internal repo with no handling instructions is still compliant with this policy
2.  if we use "must" an internal repo with no handling instructions is not compliant with this policy

I think we want (2) because we want all internal repos to have this statement, I don't think it's an onerous requirement. Of course there will be a period when repos are out-of-compliance and it will take some time to update, but that's also fine. But I think it's good to be clear on the intent which is that this should be mandatory and expected and I don't think it's an unreasonable burden

heathd (approved):
Reads really well and really clear. 

The only other follow up I would suggest is that we aim to have some of the metadata in custom properties eventually. But I think that could be done as a follow up.

source/standards/managing-sensitive-information.html.md.erb
davidr-gds: We can debate "should" vs "must" - maybe in slack?

L11 and L19 need at least be consistent

huwd:
@heathd I'll take you up on that but perhaps at the next forum... for now i'll ship with SHOULD, and then if the steer from folks at gds-way or particularly EE want to set a standard on top of this that hardens to MUST... i'll be the first one to 👏🏻 :)

So @heathd going to slightly rudely merge this against your recommendation, but very open for a fast follow here.

PR #1140, opened 2026-01-08
Add aws-cdk as a possible iac tool

mikep2468:
# changes:
- rename the configuration management page to use-infrastructure-as-code
- include aws-cdk as an iac tool with guidance on when to use

# Rationale:
This is being done in reflection of that fact that cdk is already in use within GDS and also because AWS latest guidance recommends using a general-purpose programming languages to generate Iac:
https://docs.aws.amazon.com/wellarchitected/latest/devops-guidance/dl.eac.6-use-general-purpose-programming-languages-to-generate-infrastructure-as-code.html

# Assessment of equivalent options:
https://docs.google.com/document/d/1QR2OedX3rT1JZSX4-G6kbafUAM75eku3qPI13CPq7-c/edit?tab=t.0#heading=h.ofrrxzhpzk7w


DaveLee-GDS:
Totally sound argument. Configuration Management has a totally different meaning to my mind from what this page is actually about anyway. But maybe i'm just old school.

source/partials/_nav-hosting-infrastructure.html.erb
```diff
@@ -1,7 +1,7 @@
 <ul>
   <li><a href="/standards/hosting.html">How to host a service</a></li>
   <li><a href="/standards/dns-hosting.html">Domain names and DNS records for your service</a></li>
-  <li><a href="/standards/configuration-management.html">Use configuration management</a></li>
+  <li><a href="/standards/infrastructure-as-code.html">Use infrastructure as code</a></li>
```
jfharden, line 4: You should add a redirect for this in https://github.com/alphagov/gds-way/blob/main/config/tech-docs.yml#L39 so the old page name redirects to the new one

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+ * [checkov](https://github.com/bridgecrewio/checkov) - "detects security and compliance misconfigurations"
+
+ * [tfsec](https://github.com/aquasecurity/tfsec) - "spots potential security issues"
+
+ * [tflint](https://github.com/terraform-linters/tflint) "linter focused on possible errors, best practices and so on."
+
+## AWS-CDK
+
+Use [AWS-CDK][] to configure Amazon Web Services (AWS) infrastructure.
```
davidr-gds: "Teams may use AWS-CDK" is probably better than someone reading this as telling all teams on AWS to use it

source/partials/_nav-hosting-infrastructure.html.erb
mikep2468: @jfharden - thanks, done

source/standards/infrastructure-as-code.html.md.erb
mikep2468: updated

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+# <%= current_page.data.title %>
+
+GDS uses a small selection of infrastructure as code tools because different tools are better suited to different tasks. 
+Use [infrastructure as code][] to define, deploy and manage your infrastructure. Store your infrastructure code in a version control system such as 
+Git to provide robust change tracking, change approval processes and auditability.
+
+## Puppet
+
+The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers and higher level services.
```
AP-Hunt: I thought Puppet was entirely gone? My understanding is that GOV.UK was the last bastion of its use, and it no longer uses it.

AP-Hunt:
Terraform has been the default, preferred, and recommended IaC tool at GDS for as long as I've been here (2019), and to my knowledge it was true before then too. 

I'm aware that there are teams within GDS (such as the Domains team) who make use of the AWS CDK, but I don't, personally, think that one or two teams choosing to use AWS CDK against the recommendations of the GDS Way should cause us to amend its guidance.

For my money, Terraform (or OpenTofu) is still the correct choice of IaC tool at GDS, because of its big library of providers, major support from every cloud provider, ubiquity within the industry, and vendor agnosticism[^1]

I'm not against teams using the AWS CDK or similar in principal. It cab be done for the right reasons, but I think it should happen by exception, and not be an avenue that's advertised or encouraged in our org-wide guidance.

[^1]: I mention the vendor agnosticism not to say "we might switch cloud providers", but to say that it isn't tied to one vendor, which means we can do things like configuring Fastly AND AWS in the same code

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+If your environment consists of a simple deployment artefact like an [Amazon Machine Image (AMI)][], Puppet may not be necessary, but the process for building that artefact must still be codified and version controlled.
+
+## Terraform
+
+Teams may use [Terraform][] to configure third party cloud infrastructure like Amazon Web Services (AWS) or [Fastly][].
```
galund:
```suggestion
Teams should use [Terraform][] to configure third party cloud infrastructure like Amazon Web Services (AWS) or [Fastly][].
```

(We should only change our policy once the experiments are successful and written up.)

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+ * [checkov](https://github.com/bridgecrewio/checkov) - "detects security and compliance misconfigurations"
+
+ * [tfsec](https://github.com/aquasecurity/tfsec) - "spots potential security issues"
+
+ * [tflint](https://github.com/terraform-linters/tflint) "linter focused on possible errors, best practices and so on."
+
+## AWS-CDK
+
+Teams may use [AWS-CDK][] to configure Amazon Web Services (AWS) infrastructure.
```
galund:
```suggestion
Some teams are trying out [AWS-CDK][] to configure Amazon Web Services (AWS) infrastructure.
```

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+CDK allows you to define your cloud infrastructure using a familiar programming language. This can be useful when you need to use programmatic logic to define your infrastructure, or when you want to take advantage of type safety and IDE support.
+
+### When to use CDK over Terraform
+Consider using CDK when:
```
galund, line 78: This is where we could explain that there are cases where Terraform simply doesn't work or is very unwieldy. And that CDK is often preferable to trying to wrangle CloudFormation directly. At this stage, we _do_ need to explain completely the motivations for moving away from existing guidance, for the benefit of teams that want understand the evolving context but are still Terraform shops.

huwd:
@mikep2468 could we address @galund's outstanding comments?
And then I know there's a wider debate here, but i think we either need to agree to close this or merge it.

I know there is outstanding concern from our SRE colleagues about the role of Terraform here. Could folks return to voice those concerns here, otherwise once those comments are resolved I think we should look to merge this in the next two weeks and iterate from there?

source/standards/infrastructure-as-code.html.md.erb
CoralSmith: Considering terraform remains the baseline recommendation for any complicated infrastructure, and most parts of GDS still manage multiple third party providers (AWS and Github, AWS and GCP) I've softened the stance somewhat on CDK so it's more directly _for teams in a 100% AWS context_.

Hopefully these reasons seem legitimate and thorough for folks on both sides of the argument to feel like they've got good reasons for the decisions they've made.

huwd (dismissed):

huwd (commented):
Sounds like there are still a few concerns, so removing approval

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,126 @@
...
+The use of [Puppet][] at GDS is diminishing as we move more of our infrastructure to containers 
+and higher level services.
+
+If your environment consists of a simple deployment artefact like an [Amazon Machine Image (AMI)][], 
+Puppet may not be necessary, but the process for building that artefact must still be 
+codified and version controlled.
+
+## Terraform
+
+Teams should use [Terraform][] to configure third party cloud infrastructure like Amazon Web
```
gavinmontgomery-gds, line 27: I'd recommend also being explicit about the use of the open source version or a paid enterprise service.
Secondly, guidance around the use of open Tofu may be suitable?

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,126 @@
...
+Terraform supports a [large number of providers][], and you can configure it to support multiple 
+environments with different parameters. See the [govuk-aws][] repository as an example.
+
+### When to use Terraform
+
+Consider Terraform as the default choice for managing infrastructure as code. 
+
+Terraform should be used when: 
+
+* you need to manage resources for one or more third-party providers in a single codebase
```
kevcjones-gds, line 39: Could broaden to something 'when you want to be provider agnostic' which includes multi-cloud and a lot of those resources for tools we we use

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,126 @@
...
+* you want to leverage existing Terraform modules and community resources
+* you want to benefit from an extensive library of existing terraform code within GDS
+* you do not need the complex language features of a high-level programming language
```
kevcjones-gds, line 42: Would consider adding "When reviewability is preferred. HCL is explicit and readable, which makes PRs easier to review,  what's written is close to what's deployed."

source/standards/infrastructure-as-code.html.md.erb
```diff
@@ -0,0 +1,126 @@
...
+cloud infrastructure using a familiar programming language. This can be useful when you need 
+to use programmatic logic to define your infrastructure, or when you want to take advantage 
+of type safety and IDE support.
+
+### When to use CDK
+
+Consider using CDK when:
+
+* your infrastructure is designed to be 100% AWS Serverless resources 
+* you build experimental infrastructure or prototypes that are not intended to be long lived
```
kevcjones-gds, line 81: Don't feel this one is a good thing to promote CDK for - it positions it as a 'lesser' choice

CoralSmith:
The above statements swing us back away from the last amendments. 
I'll have one last revision of this to re-emphasise the fact CDK is in trial still - with terraform still being the recommended. If that's still controversial I'm going to draft/close this until the tech radar work the @louzoid-gds is working on.

source/standards/infrastructure-as-code.html.md.erb
CoralSmith: The intention was to promote it for situations needing low complexity - so you can pick a language to write your infrastructure and your business logic. Can you think of a less controversial phrasing?

source/standards/infrastructure-as-code.html.md.erb
CoralSmith: I think that falls into familiarity/preference territory. Personally I find reviewing high-level languages easier than HCL because I spend more time in that context... would this work better? 
```suggestion
* you do not need or want the complex language features of a high-level programming language
```

source/standards/infrastructure-as-code.html.md.erb
CoralSmith: ^ agreed we'll look at this another time

source/standards/infrastructure-as-code.html.md.erb
CoralSmith: Given terraform modules are not exactly drop in for any generic unit of compute I've never quite bought this. If I move cloud providers I'm rewriting by HCL. I understand the argument though; it closely resembles the second point on this list so would you be happy considering point 2 to cover this? Or would you like me to make more call out

source/standards/infrastructure-as-code.html.md.erb
kevcjones-gds: If we reworded the 'Consider using CDK when:' to 
`Consider using CDK when one or more of the following is true:`

Pedantic maybe, but i initially read this as a conditions/criteria list but i could see someone hacking something in CDK just because its there, and easy.. it's a fair assertion but there will also be teams entirely using it because of points 1 and 4 (personal experience)

PR #1150, opened 2026-01-30
Adds Swift style guide page

joe-mcgarry:
# Changes 
- Creates a new page for the Swift Style Guide
- Gives some very barebones information about how Swift is used within GDS as a starting point

# Related PRs
- #946

### DIFF
--- a/source/manuals/programming-languages/swift.html.md.erb
+++ b/source/manuals/programming-languages/swift.html.md.erb
@@ -1,0 +1,21 @@
+---
+title: Swift style guide
+last_reviewed_on: 2026-01-30
+review_in: 6 months
+owner_slack: '#mobile-community'
+---
+
+# <%= current_page.data.title %>
+
+We use Swift for iOS development at GDS because it's the recommended language for building iOS apps. The preference for predominantly using SwiftUI or UIKit should be based on the needs of the project - mixing both SDKs in one project to utilise the strengths of both is completely fine. 
+
+Swift is not currently used outside of iOS development - even though it can be used for [backend development](https://www.swift.org/documentation/server/), we prefer using one of the following:
+
+- [Python](python/python.html)
+- [Ruby](ruby.html)
+- [Java](java.html)
+- [JavaScript/Node.js](nodejs/index.html)  
+
+Swift can now also be used for [Android development](https://www.swift.org/blog/nightly-swift-sdk-for-android/), however we prefer to use [Kotlin](kotlin.html) still.
+
+As a result, this guide may lean towards iOS-specific tools and conventions.
--- a/source/manuals/programming-languages/swift.html.md.erb
+++ b/source/manuals/programming-languages/swift.html.md.erb
@@ -20,8 +20,7 @@ As a result, this guide may lean towards iOS-specific tools and conventions.
 For naming conventions, the Swift API Design Guidelines lays out best practices for naming conventions in Swift.
 
 ### SwiftLint
-[SwiftLint] is used for linting Swift code. The default rules are based on the styling accepted by the Swift community, a good write up of Swift conventions can be found on [Kodeco]'s site.
+[SwiftLint] is used for linting Swift code. The default rules are based on the styling accepted by the Swift community.
 
 [SwiftLint]: https://github.com/realm/swiftlint
-[Kodeco]: https://github.com/kodecocodes/swift-style-guide
-[API Design Guidelines]: https://www.swift.org/documentation/api-design-guidelines/ 
+[API Design Guidelines]: https://www.swift.org/documentation/api-design-guidelines/
--- a/source/manuals/programming-languages/swift.html.md.erb
+++ b/source/manuals/programming-languages/swift.html.md.erb
@@ -9,13 +9,19 @@ owner_slack: '#mobile-community'
 
 We use Swift for iOS development at GDS because it's the recommended language for building iOS apps. The preference for predominantly using SwiftUI or UIKit should be based on the needs of the project - mixing both SDKs in one project to utilise the strengths of both is completely fine. 
 
-Swift is not currently used outside of iOS development - even though it can be used for [backend development](https://www.swift.org/documentation/server/), we prefer using one of the following:
+Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backened development. 
 
-- [Python](python/python.html)
-- [Ruby](ruby.html)
-- [Java](java.html)
-- [JavaScript/Node.js](nodejs/index.html)  
-
-Swift can now also be used for [Android development](https://www.swift.org/blog/nightly-swift-sdk-for-android/), however we prefer to use [Kotlin](kotlin.html) still.
+Swift can now also be used for [Android development](https://www.swift.org/blog/nightly-swift-sdk-for-android/), however [Kotlin](kotlin.html) is the preferred choice.
 
 As a result, this guide may lean towards iOS-specific tools and conventions.
+
+## Code style and formatting
+
+For naming conventions, the Swift API Design Guidelines lays out best practices for naming conventions in Swift.
+
+### SwiftLint
+[SwiftLint] is used for linting Swift code. The default rules are based on the styling accepted by the Swift community, a good write up of Swift conventions can be found on [Kodeco]'s site.
+
+[SwiftLint]: https://github.com/realm/swiftlint
+[Kodeco]: https://github.com/kodecocodes/swift-style-guide
+[API Design Guidelines]: https://www.swift.org/documentation/api-design-guidelines/ 
--- a/source/manuals/programming-languages/swift.html.md.erb
+++ b/source/manuals/programming-languages/swift.html.md.erb
@@ -9,7 +9,7 @@ owner_slack: '#mobile-community'
 
 We use Swift for iOS development at GDS because it's the recommended language for building iOS apps. The preference for predominantly using SwiftUI or UIKit should be based on the needs of the project - mixing both SDKs in one project to utilise the strengths of both is completely fine. 
 
-Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backened development. 
+Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backend development. 
 
 Swift can now also be used for [Android development](https://www.swift.org/blog/nightly-swift-sdk-for-android/), however [Kotlin](kotlin.html) is the preferred choice.
 
--- a/source/manuals/programming-languages.html.md.erb
+++ b/source/manuals/programming-languages.html.md.erb
@@ -19,6 +19,7 @@ review_in: 12 months
             <li><a href="/manuals/programming-languages/nodejs/index.html">Node.js</a></li>
             <li><a href="/manuals/programming-languages/python/python.html">Python</a></li>
             <li><a href="/manuals/programming-languages/ruby.html">Ruby</a></li>
+            <li><a href="/manuals/programming-languages/swift.html">Swift</a></li>
         </ul>
     </li>
 </ul>
@@ -69,6 +70,7 @@ We've got a consistent style for:
 - [Node.js](programming-languages/nodejs/)
 - [Python](programming-languages/python/python.html)
 - [Ruby](programming-languages/ruby.html) (tested with [RSpec](../standards/testing-with-rspec.html))
+- [Swift](programming-languages/swift.html)
 
 Some of the guidelines in the style guides are codified in a
 [.editorconfig](programming-languages/editorconfig) file. Place a copy of this


source/manuals/programming-languages/swift.html.md.erb
```diff
@@ -0,0 +1,21 @@
+---
+title: Swift style guide
+last_reviewed_on: 2026-01-30
+review_in: 6 months
+owner_slack: '#mobile-community'
+---
+
+# <%= current_page.data.title %>
+
+We use Swift for iOS development at GDS because it's the recommended language for building iOS apps. The preference for predominantly using SwiftUI or UIKit should be based on the needs of the project - mixing both SDKs in one project to utilise the strengths of both is completely fine.
```
jonnyandrew, line 10: About SwiftUI vs UIKit - do you think our advice (to prefer neither) might quickly fall out of date?

As I understand it, SwiftUI is the newer, more modern UI framework for iOS, based on the general trend towards declarative UI.

Although Apple might not be deprecating UIKit or recommending that we prefer SwiftUI, the docs imply that you'd migrate from UIKit to SwiftUI and not the other way around:

> Adopt SwiftUI at your own pace
> SwiftUI is designed to work alongside UIKit and AppKit, so you can adopt it incrementally in your existing apps

– https://developer.apple.com/swiftui/

I'm wondering if we'd be better off waiting until we do have a preference for one or the other framework, and add the guidance when that time comes?

source/manuals/programming-languages/swift.html.md.erb
```diff
@@ -0,0 +1,21 @@
...
+last_reviewed_on: 2026-01-30
+review_in: 6 months
+owner_slack: '#mobile-community'
+---
+
+# <%= current_page.data.title %>
+
+We use Swift for iOS development at GDS because it's the recommended language for building iOS apps. The preference for predominantly using SwiftUI or UIKit should be based on the needs of the project - mixing both SDKs in one project to utilise the strengths of both is completely fine. 
+
+Swift is not currently used outside of iOS development - even though it can be used for [backend development](https://www.swift.org/documentation/server/), we prefer using one of the following:
```
jonnyandrew: Reading this section, I wasn't sure it was 100% clear where we're talking about 'how Swift can be used in general' versus 'how we use it at GDS'.

Perhaps we can simplify some of the detail and link out to the existing programming languages page? E.g.

> Although Swift is a general purpose programming language, we only use it for iOS development at GDS. See [_programming languages_](standards/programming-languages.html) for languages that we recommend for other use cases.

jonnyandrew (commented):
Thanks @joe-mcgarry !

zzcgumn:
Here are a couple of examples:
- Google's [swift style](https://google.github.io/swift/) guide which is based on the Swift standard library's style.
- Ray Wenderlich official [style guide](https://github.com/RedMadRobot/RMR-swift-style-guide), it is focused on readability in blog posts and printed materials.
- API [design guidelines](https://www.swift.org/documentation/api-design-guidelines/) from the Swift project.

source/manuals/programming-languages/swift.html.md.erb
zzcgumn: We are currently leaning towards `UIKit`, I have been told that the reason for this is that some accessibility features were missing in `SwiftUI`. This might be something that I have misunderstood and I have not seen the evidence.

`SwiftUI` should be preferred once it offers the same accessibility support as `UIKit`.

joe-mcgarry: It's a very good question, and one I'd probably want more experienced iOS devs to weigh in on. My take is that in DI, it's predominantly UIKit - mostly due to when the ID Check app was being made SwiftUI wasn't quite fleshed out enough, especially for accessibility - maybe this is something that should be checked with how important accessibility is. I'm unsure on the GOV.UK side of things, but I think from a very quick peek into their repo, it's a mix of both UIKit and SwiftUI

SwiftUI is absolutely the direction Apple is wanting people to go in, but what you can do in Swift (and maybe this isn't clear) is have your project mostly in UIKit, but you could make your views in the slightly easier/less verbose SwiftUI. At one point, all of the BRP journeys in the ID Check app were using SwiftUI while the others were all using UIKit (which is really all I was trying to say, as some tasks are easier in one than the other). 

And as it's a very common question for beginners to Swift, I figured it was worth saying something, even if what I've currently put isn't quite right. But if the majority think it isn't worth mentioning, I'm more than happy to simply remove it!

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: I get what you mean, but I would personally be hesitant to say it's general purpose (and I fully admit this is me being very nit-picky, and I could be missing some developments here and there). As far as I know, in the web world, it really is only the back end which Swift handles (I don't think the common packages have something like Ruby's ERB files to do front end pages), and equally for non-MacOS applications, I don't think it can be used. 

Take that all with a pinch of salt, but that was always my understanding. But I would love for others who know more, to chime in - because I do agree what I've written at the moment is a bit silly. Maybe just delete that section altogether, and when the page is up for review get people to chime in and discuss if something needs adding?

jonnyandrew: Ah - I think my suggestion changes too many things here at once, and introduces different issues. Definitely ignore 'general purpose' if Swift is not that :)

My points were aimed to address the following:

**1. Reducing ambiguity about the context we're referring to**

- Before: `Swift is not currently used outside of iOS development - even though it can be used for backend development ...`
- After: `Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry ...`

**2. Reducing detail already covered by other pages that might fall out of date**

- Before: `we prefer using one of the following: Python, Ruby, Java, Javascript`
- After: `we prefer using other [programming languages](/programming-languages)`

source/manuals/programming-languages/swift.html.md.erb
jonnyandrew: > some accessibility features were missing in `SwiftUI`

> SwiftUI wasn't quite fleshed out enough, especially for accessibility

I think this is the context I was missing. Knowing this, I think it's definitely an important point, worth including in this style guide.

Perhaps it deserves its own small subsection later on in the doc (e.g. User interfaces), so that we can include that bit of context?

Maybe we can include an example of one of the accessibility limitations we're talking about and (if available) a link to official sources e.g. public SwiftUI issue trackers or SwiftUI accessibility roadmap.

batesyuk: We lean towards SwiftUI more now on GOV.UK, mostly because that is the direction the industry seems to be heading, but also because that is what developers seem to be wanting to use, so can help with recruitment.
We still have to make the decision on a case by case basis (often design-dependent as much as accessibility)
I can't see UIKit going anywhere soon, so we can just update down the line when it does.

huwd:
@batesyuk sorry GDS Way forum came around again, anyone on your side able to look at this?

source/manuals/programming-languages/swift.html.md.erb
jonnyandrew: Thanks for the updates!

joe-mcgarry:
@huwd @jonnyandrew @batesyuk @zzcgumn Sorry for the wait in updating this, but I think the Swfit page is now in-line with what Kotlin looks like, so at the very least there's some level of parity for the 2 languages. Would be good to figure out how to get a more fleshed style guide out in a way that's easier for iOS folk to contribute to

huwd:
thanks @joe-mcgarry i've pinnged @batesyuk on slack to see if we can get some GOV.UK App reviews. broadly if you can get a green tick off a One Login app person (go fish around in our slack) and a GOV.UK App person, let's merge it!

Great work and thanks for picking this up.

If you could respond and resolve that remaining comment too that'd be ace!

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: Thanks for your input @jonnyandrew! Are you happy with the changes made as a first iteration?

source/manuals/programming-languages/swift.html.md.erb
```diff
@@ -0,0 +1,27 @@
...
+Swift can now also be used for [Android development](https://www.swift.org/blog/nightly-swift-sdk-for-android/), however [Kotlin](kotlin.html) is the preferred choice.
+
+As a result, this guide may lean towards iOS-specific tools and conventions.
+
+## Code style and formatting
+
+For naming conventions, the Swift API Design Guidelines lays out best practices for naming conventions in Swift.
+
+### SwiftLint
+[SwiftLint] is used for linting Swift code. The default rules are based on the styling accepted by the Swift community, a good write up of Swift conventions can be found on [Kodeco]'s site.
```
batesyuk: This write-up includes things like using 2 spaces. While I would prefer to use 2 spaces, I don't think any of our apps currently do.
Should we link to a document that could contradict the practices we actually use? If so how do we know what should and shouldn't be followed for the style?

Side note: I would much rather use something like this than make and maintain a bespoke one

source/manuals/programming-languages/swift.html.md.erb
```diff
@@ -0,0 +1,27 @@
...
+
+[SwiftLint]: https://github.com/realm/swiftlint
+[Kodeco]: https://github.com/kodecocodes/swift-style-guide
+[API Design Guidelines]: https://www.swift.org/documentation/api-design-guidelines/
```
danielmason-gds: Probably good to stick an End Of Line at the End Of File 😉

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: Thanks for the input everyone. I tried capturing that and making it clearer to basically use the most appropriate Swift SDK for the job (which I will imagine will be SwiftUI for any new apps). 

Is there anything we'd like changed or added?

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: That is a good catch! I was mostly going by the SwiftLint [github repo](https://github.com/realm/swiftlint) which points to that style guide. As far as I know in OneLogin SwiftLint wasn't changed, so I've removed the reference to Kodeco and left it as `...The default rules are based on the styling accepted by the Swift community.` Which, I think for now at least, takes away any slightly-more-urgent need for a cross-team discussions for a common linting strategy

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: I can't tell if this is still an issue or not, but I swear locally I do have a newline at the end of file, but to me (at least) it doesn't look like it does in the file changes 🙃

source/manuals/programming-languages/swift.html.md.erb
jonnyandrew: Yes, your latest changes resolve this comment :)

source/manuals/programming-languages/swift.html.md.erb
```diff
@@ -0,0 +1,26 @@
...
+last_reviewed_on: 2026-01-30
+review_in: 6 months
+owner_slack: '#mobile-community'
+---
+
+# <%= current_page.data.title %>
+
+We use Swift for iOS development at GDS because it's the recommended language for building iOS apps. The preference for predominantly using SwiftUI or UIKit should be based on the needs of the project - mixing both SDKs in one project to utilise the strengths of both is completely fine. 
+
+Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backened development.
```
jonnyandrew: I just spotted a small typo:
```suggestion
Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backend development. 
```

zzcgumn:
```suggestion
Swift is not currently used outside of iOS development at GDS - even though it is used for other members of the Apple echo-system and backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backend development. 
```

source/manuals/programming-languages/swift.html.md.erb
```diff
@@ -0,0 +1,26 @@
...
+
+Swift is not currently used outside of iOS development at GDS - even though it is used for backend development in the wider industry, we prefer using other [programming languages](/programming-languages) for backened development. 
+
+Swift can now also be used for [Android development](https://www.swift.org/blog/nightly-swift-sdk-for-android/), however [Kotlin](kotlin.html) is the preferred choice.
+
+As a result, this guide may lean towards iOS-specific tools and conventions.
+
+## Code style and formatting
+
+For naming conventions, the Swift API Design Guidelines lays out best practices for naming conventions in Swift.
```
zzcgumn, line 20:
```suggestion
For naming conventions, the Swift API Design Guidelines lays out best practices for naming conventions in Swift.

[Swift API Design Guidelines]: https://www.swift.org/documentation/api-design-guidelines/
```

zzcgumn: I see that the link is in fact mentioned below but am no allowed to resolve the comment myself.

source/manuals/programming-languages/swift.html.md.erb
jonnyandrew: I'm happy to resolve this comment 👍

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: Good spot! I've fixed that, thanks!

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: Ah yes! This was something I noticed from @jonnyandrew's Kotlin PRs! I find it a much neater way of handling with links in markdown, but if there is a preference to have it within the actual text so it's easier to see the link has actually been added, I'm more than happy to do that!

batesyuk (approved):

shukriahmed-dd (approved):

source/manuals/programming-languages/swift.html.md.erb
CoralSmith: I can resolve threads, if you're happy for me to do that for you @zzcgumn?

huwd:
Amazing @joe-mcgarry / @CoralSmith looks like we've got the ticks!

Can we resolve any threads (or can someone put a block on if they feel super strongly?)
It'd be great to merge this up tomorrow.

👏🏻 to everyone for driving this iOS now has a place in the GDS way!

jonnyandrew (approved):

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: If it's fine with you, I'm going to resolve this now - but happy to follow up on anything if this needs any further discussion!

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: @zzcgumn If it's alright with you, I'm going to resolve this now, but more than happy to discuss this (and your other comments too) if you feel strongly about the wording and changes being needed

source/manuals/programming-languages/swift.html.md.erb
joe-mcgarry: Adding a link to this comment to explain why I've resolved the comment: https://github.com/alphagov/gds-way/pull/1150#discussion_r3394246438

PR #1163, opened 2026-03-02
Add a section on possible tech radar language to the home page

louzoid-gds:
Proposal: A new section on the homepage explaining some new labels that we might like to use across the site. These labels use tech radar style language to explain when a tool/practice should be adopted, trialed, assessed or held.

We've talked about adding this in the past and I think might be very handy to help us articulate what we'd like to do in PR #1140 and no doubt some other pages on the site.

Feedback please! Is this a good idea? Are the words against each label ok? All thoughts please :)

Screenshot 2026-03-02 at 09 36 21

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -8,6 +8,7 @@ title: The GDS Way
     <li><a href='#the-gds-way'>The GDS Way</a>
         <ul>
             <li><a href='#about-the-gds-way'>About the GDS way</a></li>
+            <li><a href='#how-we-categorise-tools-and-technologies'>How we categorise tools and technologies</a></li>
             <li><a href='#how-to-add-new-guidance'>How to add new guidance</a></li>
             <li><a href='#the-gds-way-forum'>The GDS Way Forum</a></li>
         </ul>
@@ -67,6 +68,19 @@ Products at GDS in discovery or alpha development phases must follow [agile deli
 
 Products in beta and live phases must follow both the instructions set out in the Service Manual and the standards in this repository.
 They must be [secure by design](https://www.security.gov.uk/guidance/secure-by-design/).
+
+## How we categorise tools and technologies
+
+We are experimenting with using [Tech Radar](https://www.thoughtworks.com/radar) language to label some tools and technologies across the GDS Way. This is not yet applied consistently and only some pages use these labels at the moment.
+
+Where labels do appear, they mean:
+
+- <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Product Group unless you have a specific reason not to
+- <strong class="govuk-tag govuk-tag--blue">Trial</strong> — being evaluated; it may be possible to use this but check with the tech leadership in your area first
+- <strong class="govuk-tag govuk-tag--yellow">Assess</strong> — possibly worth exploring, but not yet ready for production use
+- <strong class="govuk-tag govuk-tag--red">Hold</strong> — avoid for new work; migrate away from this if you can
+
+Where a tool or technology has no label, follow the guidance on that page.
 
 ## How to add new guidance
 
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -75,8 +75,8 @@ We are experimenting with using [Tech Radar](https://www.thoughtworks.com/radar)
 
 Where labels do appear, they mean:
 
-- <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Product Group unless you have a specific reason not to
-- <strong class="govuk-tag govuk-tag--blue">Trial</strong> — being evaluated; it may be possible to use this but check with the tech leadership in your area first
+- <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Digital Products unless you have a specific reason not to
+- <strong class="govuk-tag govuk-tag--blue">Trial</strong> — being evaluated; it may be possible to use this but check with the lead technologists in your area first
 - <strong class="govuk-tag govuk-tag--yellow">Assess</strong> — possibly worth exploring, but not yet ready for production use
 - <strong class="govuk-tag govuk-tag--red">Hold</strong> — avoid for new work; migrate away from this if you can
 
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -78,7 +78,7 @@ Where labels do appear, they mean:
 - <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Digital Products unless you have a specific reason not to
 - <strong class="govuk-tag govuk-tag--blue">Trial</strong> — being evaluated; it may be possible to use this but check with the lead technologists in your area first
 - <strong class="govuk-tag govuk-tag--yellow">Assess</strong> — possibly worth exploring, but not yet ready for production use
-- <strong class="govuk-tag govuk-tag--red">Hold</strong> — avoid for new work; migrate away from this if you can
+- <strong class="govuk-tag govuk-tag--red">Caution</strong> — avoid for new work; migrate away from this if you can
 
 Where a tool or technology has no label, follow the guidance on that page.
 


danielmason-gds (dismissed):
I like this! Hopefully it'll encourage people to innovate while adding a light layer of structure to the process.

mrwilson:
Happy to support this initial incarnation, pending further discussion (as on Slack) about how we organise and expose these labels; individual page labels and a single "one stop shop" page for every tech radar category.

CoralSmith (dismissed):
I like this initial incarnation - the thoughtworks categories are probably familiar to most people and it's a good way to set context for the reader and track innovation in our standards.

source/index.html.md.erb
```diff
@@ -68,6 +69,19 @@ Products at GDS in discovery or alpha development phases must follow [agile deli
 Products in beta and live phases must follow both the instructions set out in the Service Manual and the standards in this repository.
 They must be [secure by design](https://www.security.gov.uk/guidance/secure-by-design/).
 
+## How we categorise tools and technologies
+
+We are experimenting with using [Tech Radar](https://www.thoughtworks.com/radar) language to label some tools and technologies across the GDS Way. This is not yet applied consistently and only some pages use these labels at the moment.
+
+Where labels do appear, they mean:
+
+- <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Product Group unless you have a specific reason not to
```
galund:
```suggestion
- <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within the GDS Digital Products group unless you have a specific reason not to
```

galund (commented):
"Hold" sounds more like "deprecated" to me but yeah I like this idea.

source/index.html.md.erb
huwd: These changes make sense to me

huwd:
@louzoid-gds what do you think? is this ready for review?

louzoid-gds:
@huwd I think I need to work some of the feedback in (from both here and in Slack) and then will mark as ready for review 👍 
We'll also need to publish alongside a page where we're actually using it so will need a bit of wrangling.

heathd:
I'd suggest being more direct. Whilst Thoughtworks deserve a credit I think we can be more explicit about _our_ intent for using the labels. 

> We are introducing [Tech Radar](https://www.thoughtworks.com/radar)-inspired labels to express our level of confidence in each piece of guidance. This is a work in progress, so if a page isn't labelled follow the guidance as described on the page.
> 
> Where labels do appear, they mean:
> 
> - <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Product Group unless you have a specific reason not to
> - <strong class="govuk-tag govuk-tag--blue">Trial</strong> — being evaluated; it may be possible to use this but check with the tech leadership in your area first
> - <strong class="govuk-tag govuk-tag--yellow">Assess</strong> — possibly worth exploring, but not yet ready for production use
> - <strong class="govuk-tag govuk-tag--red">Hold</strong> — avoid for new work; migrate away from this if you can



I agree with other commenters that the "Hold" label serves multiple purposes, and is sometimes used with a slightly tounge-in-cheek rhetorical voice (which I personally enjoy but may be a bit confusing). It's used to say:

1. Too immature, wait and see
2. We tried it and recommend against it
3. Overused or applied in wrong contexts

In our context I think we might want an explicit _we're deprecating this thing_ label. I think (2) will be rare in our world or overlap with the "we're deprecating" steer.

However, in the interest of forward progress I'd vote for just keeping the thoughtworks labels including "Hold" and maybe adding a "sunset" one later if we think we need it.

Jacqui-R-GDS:
A note to say that ThoughtWorks have started using _Caution_ rather than _Hold_. Something to consider.
(@galund I saw you commented on Hold)

louzoid-gds:
Thanks for the feedback all. I've tweaked some of the wording but only ever so slightly.
I don't think I'll move this out of draft until we've had a go at implementing on an actual page and can see it in action.

kevcjones-gds:
Late to this one (think I wandered in from a different PR) but +1 on the idea - having one place to check "have we looked at this before?"

Bit of a similar to @heathd shaped thought (maybe). The Thoughtworks radar is advisory by design, "things we think you should consider," your call in the end, and they deliberately won't even have an "avoid" ring. Ours reads more like policy, which makes sense given it's the GDS *Way*. The label wording in this PR goes the same way: "use unless you have a specific reason not to", "check with the lead technologists first", "migrate away". And "when a tool should be adopted or held" is more policy/rule, than a nudge.

None of that's a problem if setting direction is the point. I might consider avoiding using their labels at all possibly to avoid confusing the two things and making it your own entirely.

The question is i suppose - are we sharing what we'd recommend to GDS engineers, or saying what's allowed? If its the latter will it need an owner and some way to argue a label?

Jacqui-R-GDS:
Naming things is definitely one of the hardest things to do in software... I'd agree it is important to think about, and also agree that having a clear definition is very important (more important than the name, which will usually be misunderstood by somebody...).

### Thoughts on nudge vs. policy
I've always leaned towards the _use this unless you have a reason not to, then talk to us about what you need to use instead_ approach. This gives top-down direction and guidance for efficiency, compatibility, stability, not recreating the wheel, saving time on research, etc.

It also gives wiggle room for different needs, innovation, and for taking change into account, but again with that top-down guidance. When a decision is superseded or an exception approved then that change is communicated via the radar for future projects to consider.

#### Top-Down Guidance
**Talk to us if it doesn't work for you:** This allows for flexibility while keeping the top-down guidance and oversight. It helps avoid shadow development, where projects use the technology that meets their needs without others knowing. The exception to this is where technologies have been flagged as a danger or similar, although communication channels should remain open for queries (and the radar updated with missing information when these queries are answered).

**Use what works:** Stability, security, and compatibility are usually important. Using older technologies that are tried and tested, with many bugs fixed, is often a better decision than using the latest and greatest tech.

#### Bottom-Up Flexibility

**Differing needs:** Software projects are varied in requirements and context. If a technology doesn't meet the requirements and context it can't be forced on the project if the project is to be successful.

**Innovation:** The software domain is inherently flexible, but also complex, competitive, and fast moving. Project must be able to innovate in order to manage complexity, meet requirements, remain competitive (less important in the public sector), and adapt to new threats and opportunities.

**Change:** Things change incredibly quickly in software. The reasoning for a previous decision to recommend a technology may now not be relevant. New options may now be available. The technology may be moving towards being sunset or now be unmaintained. A new fork may have fixed problems or security holes. Security problems may have surfaced. There may now be a directive to not allow data to be stored in a particular region. So. Much. Change.

#### Trust and Less Bureaucracy
This approach is a step towards less bureaucracy. The combination of guidance and open communication fosters trust. Those giving the guidance do not need to enforce gates and checks, because they know that projects are following guidance or talking to them. Working in the open and transparency (open meetings, documentation, etc) can also contribute to higher trust.

### Current labels
- _Adopt:_ essentially this is _use this unless you have a reason not to, then talk to us about what you need to use instead_.
- _Trial:_ this helps avoid duplicating research and invites communication. **We should encourage a 2-way conversation by asking for input into the trial.**
- _Assess:_ **This needs more guidance in the definition. If a team wants to use this technology who is responsible for assessing it? What is the process for trialing it? We want 2-way communication, rather than teams going off on their own to assess technologies. Especially to avoid duplication of effort.**

It may be a good idea to have 2 different labels here. One for technologies to not use for new projects and preferably migrate away from, and one to migrate from ASAP. Unless the urgent label is handled in another way.
- _Hold:_ avoid for new work; migrate away from this if you can
- _Migrate:_ **migrate to the recommended technology urgently (this definitely needs linking to guidance on why and how)**

**Linking to the reasoning for the label, e.g. an ADR, is important so that teams understand why a label applies. They then know whether they have a good reason for using another technology.**

This radar may contribute to projects having an SBOM (software bill of materials). I don't know if this already exists or is in progress.

Note: sorry, this seems to have turned into a blog post...

source/index.html.md.erb
```diff
@@ -68,6 +69,19 @@ Products at GDS in discovery or alpha development phases must follow [agile deli
...
+## How we categorise tools and technologies
+
+We are experimenting with using [Tech Radar](https://www.thoughtworks.com/radar) language to label some tools and technologies across the GDS Way. This is not yet applied consistently and only some pages use these labels at the moment.
+
+Where labels do appear, they mean:
+
+- <strong class="govuk-tag govuk-tag--green">Adopt</strong> — the recommended default; you should use this within GDS Digital Products unless you have a specific reason not to
+- <strong class="govuk-tag govuk-tag--blue">Trial</strong> — being evaluated; it may be possible to use this but check with the lead technologists in your area first
+- <strong class="govuk-tag govuk-tag--yellow">Assess</strong> — possibly worth exploring, but not yet ready for production use
+- <strong class="govuk-tag govuk-tag--red">Hold</strong> — avoid for new work; migrate away from this if you can
```
danielmason-gds:
```suggestion
- <strong class="govuk-tag govuk-tag--red">Caution</strong> — avoid for new work; migrate away from this if you can
```

I believe this change provides a middle ground to address concerns raised by @Jacqui-R-GDS @kevcjones-gds @heathd and @galund. The word "Caution" is clearer while being in the spirit of "look if it really works for you, go ahead, just be careful".

Acknowledging that this isn't a 1:1 match of how Thought Work's use this tool, and that's ok, this is for us 🙂 

Regardless of this change, I think we should trial this system and see what does and doesn't work for us. We can always make changes later, and it'll be easier to get feedback from people when they can see it in action rather than as a hypothetical in a PR 😉

PR #1176, opened 2026-03-25
programming-languages: deprecate go

paroxp:
Go has been used in parts of GDS for a long while. The GOV.UK Router, to GOV.UK PaaS, various Kubernetes projects and CLI tools - it shall remain in our hearts and undoubtedly will return in some shape or form.

However, there is not a ~~single team~~ lot of teams in GDS actively working with Go. Hasn't been for a while now... Which makes it really difficult to keep on top of the best practices and keeping this GDS Way section up to date with best standards in mind, turning this effectively into an annoying annual ritual of bumping the date...

I look forward to the day, Go returns as a first class language of choice. Farewell you little gopher.


AP-Hunt:
> However, there is not a single team in GDS actively working with Go.

This isn't true. GOV.UK Platform Engineering has committed to writing in Go as one of our primary languages for new development, such as CLI tools, Kubernetes operators, APIs, and [Crossplane functions](https://github.com/crossplane/crossplane/blob/main/contributing/guide-provider-development.md)

We also still maintain and improve the following in Go:

* Router
* Mirror
* GOV.UK exporter.

We have no plans or desire to stop writing in Go.

image

paroxp:
Sorry Andy, my wording could have been better.

I do realise there are still teams and projects around that are based on Go. And I also do realise that any future development around K8s may still be mainly go. I suspect what I mean, is unlike Java, TypeScript or Ruby, Go is rather self-contained within a specialised team and not a GDS wide supported language. It's unlikely at the moment that DI or DSP for instance will pick up Go and benefit from this page existing.

But I am not against keeping it - if GOV.UK were to be keen to take care, modernise and keep it up to date for the rest of GDS?

Go is changing, as anything else. And I am worried that this guidance could actually cause some friction for people if someone were to follow it?

dj-maisy:
Many of the toolsets we rely on across GDS such as Kubernetes (plus the vast majority of the CNCF framework) and Terraform are built on top of Go. It might have a bit of a niche across GDS teams outside of GOV.UK, but it definitely has its place, especially when there are situations where Java and Ruby aren't quick enough, Go comes out as something that is flexible, fast and has a lot of community support where other libraries and frameworks sometimes don't.

I'm keen to not go on too much of a tangent here as I feel @AP-Hunt has already repped GOV.UK's stake here.

dj-maisy:
We're going to close this as [Platform Engineering is now assuming ownership of the Go pages](https://github.com/alphagov/gds-way/pull/1210).

PR #1201, opened 2026-04-28
Python dependencies rewrite

davidr-gds:
Removed a lot of prose in the Dependencies section that is not really necessary any more, not unique to GDS, and buries the important GDS-specific guidance. @galund who wrote it a long time ago is supportive of this.

**uv** recommendation now comes with explanation is now explained.

`requirements.in` is not really acceptable security now, with the likes of litellm malware being published. Replaced with advice to use a **lockfile** with its "hash checking" feature.

Moved the "library" section to the bottom, because that seems clearer to treat it as an exception, than to make most readers consider if they are writing an "application or library" equally.

Added section "Updating [dependencies]" which defers to [How to manage third party software dependencies
](https://gds-way.digital.cabinet-office.gov.uk/standards/tracking-dependencies.html) for high level advice on updating dependencies frequently etc (and hopefully soon will recommend cooldowns), and focuses on Python-specific concerns of Dependabot config for uv.

### DIFF
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -122,16 +122,26 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
 
 ## Dependencies
 
-**Dependency manager** - you should use [uv][uv-readme]. This is recommended because it is fast, modern, aligned with standards like pyproject.toml and is well-maintained. Prefer it to pip, poetry and pipenv.
+### Dependency manager
 
-**Specifying dependencies** - you should use pyproject.toml to specify your direct dependencies. (This is because it is the modern recommended standard.)
+You should use [uv][uv-readme]. This is recommended because it is fast, modern, aligned with standards like `pyproject.toml` and is well-maintained. Prefer it to pip, poetry and pipenv.
 
-**Pinning and lockfiles** - you **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
+### Specifying dependencies
 
-**Updating** - you **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
-* You should configure a **cool-off period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
-* You should have a process to **update quickly** to new versions, after the cool-off period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
-* Document your update process in your README.
+You should use `pyproject.toml` to specify your direct dependencies because it is the modern, recommended standard.
+
+### Pinning and lockfiles
+
+You **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
+
+### Updating
+
+You **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
+* You should configure a **cooldown period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
+
+You should have a process to **update quickly** to new versions, after the cooldown period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
+
+Document your update process in your README.
 
 ### Libraries
 
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -1,6 +1,6 @@
 ---
-title: Python style guide
-last_reviewed_on: 2025-02-14
+title: Python style and standards
+last_reviewed_on: 2026-04-28
 review_in: 12 months
 owner_slack: '#python'
 ---
@@ -122,93 +122,24 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
 
 ## Dependencies
 
-A Python application project typically brings together Python packages from PyPI, with
-others written in-house (or otherwise not distributed through PyPI).
+**Dependency manager** - you should use [uv][uv-readme]. This is recommended because it is fast, modern, aligned with standards like pyproject.toml and is well-maintained. Prefer it to pip, poetry and pipenv.
 
-These packages are the applications immediate dependencies. Additionally, any
-package can have its own immediate dependencies, where they draw on other
-packages. From an application's point of view, the dependencies of the packages
-it requires are its sub-dependencies.
+**Specifying dependencies** - you should use pyproject.toml to specify your direct dependencies. (This is because it is the modern recommended standard.)
 
-The below diagram shows a simplified view of the resulting pyramid of dependencies;
-however it is important to note that the hierarchy can repeat itself infinitely.
+**Pinning and lockfiles** - you **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
 
-```
-                               Application
-
-                                    |
-                              +-----+-----+
-                              |     |     |
-
-                          Immediate dependencies
-
-                              |     |     |
-                        +--+--+--+--+--+--+--+--+
-                        |  |  |  |  |  |  |  |  |
-
-                         ... Sub-dependencies ...
-```
-
-There are two ways of specifying dependencies in Python world: as a specific version or
-as an allowable range.
-
-Different considerations apply to dependency management depending whether you are packaging a library, or creating
-an end-product such as an application or a bundle of scripts. In general, you should only specify specific versions
-when creating a Python system that sits at the top of the dependency pyramid; otherwise there is a danger of creating
-version conflicts.
-
-### Applications
-
-These recommendations apply wherever you need a reproducible set of dependencies, such
-as a complete web application, perhaps with many dependencies and sub-dependencies. It also
-applies to a collection of scripts that are deployed into the cloud and run automatically
-(for example, batch jobs).
-
-A good strategy for specifying your application's Python dependencies has two desirable
-characteristics - they should be:
-
-1. Reproducible (predictable)
-
-    Pin your application's full dependencies – specific versions, rather than ranges – or you'll
-    get unpredictability between your dev environment and other environments. You want a
-    new starter to avoid small hard-to-spot problems. And you want parity between what you
-    test locally, what is tested by CI, and what you deploy, or you risk new issues appearing on
-    a live server. Additionally these things can be hard to diagnose.
-
-2. Kept up-to-date
-
-    Security issues are found in libraries, so it is important to choose libraries that are
-    maintained and to ensure your team has a strategy to ensure security updates are installed
-    without significant delay. The [how to manage third party software dependencies][gds-way-deps] section
-    gives further context and discusses tools that can help, such as [snyk.io][].
-
-Your README should document an easy-to-follow process by which all your Python
-dependencies can be upgraded to get bug fixes and security fixes, without introducing
-breaking changes to your build.
-
-Your pinned dependencies should be fully specified in a file called `requirements.txt`, and checked
-into your version control system (VCS). For projects with only a small number of dependencies, maintaining
-this manually (for example, installing with `uv pip install`, then using `pip freeze`) may be adequate.
-
-GDS recommends using [uv][uv-readme] for managing dependencies.
-
-Put your top-level requirements in a `requirements.in` file, and then use
-`uv pip compile requirements_for_test.in -o requirements_for_test.txt` to generate a
-`requirements.txt` file. Both the .in and .txt files should be commited to your repository.
-
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+**Updating** - you **must** keep dependencies up to date with security patches, and the latest major & minor versions (or older versions which are expected to receive security patches). You should **automate the updating**, for example with Dependabot and making sure the PRs are merged, lockfiles updated and deployed in a timely fashion. This is to ensure that you quickly move to the secure versions of packages. Document this update process in your README. See also: [how to manage third party software dependencies][gds-way-deps].
 
 ### Libraries
 
-This recommendation applies to any Python repository that intended to be installable (into
+Libraries are the exception to the rule about exact dependencies being specified and pinned.
+This applies to any Python repository that intended to be installable (into
 a virtual environment, a container, or onto bare metal) as a dependency of some larger system
 or application. It may be applicable to repositories that provide scripts to be run by
 developers or other end-users, but is not recommended for code that's intended to be deployed
 on its own into the cloud.
 
-Use a [`pyproject.toml`][pyproject-toml] to specify your library's configuration.
-When specifying the dependencies of your library and the version ranges with which it
-can be reasonably expected to work:
+Put **version ranges** in your [`pyproject.toml`][pyproject-toml] to specifying the dependencies of your library. These are the rangeswith which it can be reasonably expected to work:
 
 - The range you choose will depend on the guarantees each dependency makes about
   backward-compatibility. For example, if you're currently using version 1.3.1 of
@@ -240,10 +171,8 @@ file otherwise.
 [slack-python]: https://gds.slack.com/messages/python
 [GPSG]: https://google.github.io/styleguide/pyguide.html
 
-[snyk.io]: https://snyk.io/
 [pyproject-toml]: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/
 
-[Black]: https://black.readthedocs.io/en/stable/
 [PEP 8]: https://www.python.org/dev/peps/pep-0008/
 [PEP 440]: https://www.python.org/dev/peps/pep-0440/#direct-references
 
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -129,7 +129,7 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
 **Pinning and lockfiles** - you **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
 
 **Updating** - you **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
-* Configure a **cool-off period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
+* You should configure a **cool-off period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
 * You should have a process to **update quickly** to new versions, after the cool-off period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
 * Document your update process in your README.
 
--- a/source/manuals/programming-languages/python/python.html.md.erb
+++ b/source/manuals/programming-languages/python/python.html.md.erb
@@ -128,7 +128,10 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
 
 **Pinning and lockfiles** - you **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
 
-**Updating** - you **must** keep dependencies up to date with security patches, and the latest major & minor versions (or older versions which are expected to receive security patches). You should **automate the updating**, for example with Dependabot and making sure the PRs are merged, lockfiles updated and deployed in a timely fashion. This is to ensure that you quickly move to the secure versions of packages. Document this update process in your README. See also: [how to manage third party software dependencies][gds-way-deps].
+**Updating** - you **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
+* Configure a **cool-off period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
+* You should have a process to **update quickly** to new versions, after the cool-off period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
+* Document your update process in your README.
 
 ### Libraries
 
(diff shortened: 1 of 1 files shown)


source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,24 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-this manually (for example, installing with `uv pip install`, then using `pip freeze`) may be adequate.
-
-GDS recommends using [uv][uv-readme] for managing dependencies.
-
-Put your top-level requirements in a `requirements.in` file, and then use
-`uv pip compile requirements_for_test.in -o requirements_for_test.txt` to generate a
-`requirements.txt` file. Both the .in and .txt files should be commited to your repository.
-
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+**Updating** - you **must** keep dependencies up to date with security patches, and the latest major & minor versions (or older versions which are expected to receive security patches). You should **automate the updating**, for example with Dependabot and making sure the PRs are merged, lockfiles updated and deployed in a timely fashion. This is to ensure that you quickly move to the secure versions of packages. Document this update process in your README. See also: [how to manage third party software dependencies][gds-way-deps].
```
ipdigital24: in notify, we have made a conscious choice to disable dependabot creating any pull requests to keep away from any supply chain related attacks; instead we have opted a manual process of freezing requirements with 7 days cooling period for all dependencies which we are running on weekly basis. we do have dependabot alerts enabled tho to get notified of any vulnerability on existing dependencies

davidr-gds: Thank @ipdigital24. In 6ea5bcb I've changed the advice on updating, from emphasising **automation** to emphasising **updating quickly**. And recommended a cool-off period.

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
+This applies to any Python repository that intended to be installable (into
 a virtual environment, a container, or onto bare metal) as a dependency of some larger system
 or application. It may be applicable to repositories that provide scripts to be run by
 developers or other end-users, but is not recommended for code that's intended to be deployed
 on its own into the cloud.
 
-Use a [`pyproject.toml`][pyproject-toml] to specify your library's configuration.
-When specifying the dependencies of your library and the version ranges with which it
-can be reasonably expected to work:
+Put **version ranges** in your [`pyproject.toml`][pyproject-toml] to specifying the dependencies of your library. These are the rangeswith which it can be reasonably expected to work:
```
quis:
```suggestion
Put **version ranges** in your [`pyproject.toml`][pyproject-toml] to specifying the dependencies of your library. These are the ranges with which it can be reasonably expected to work:
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-
-GDS recommends using [uv][uv-readme] for managing dependencies.
-
-Put your top-level requirements in a `requirements.in` file, and then use
-`uv pip compile requirements_for_test.in -o requirements_for_test.txt` to generate a
-`requirements.txt` file. Both the .in and .txt files should be commited to your repository.
-
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+**Updating** - you **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
+* You should configure a **cool-off period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
```
quis:
```suggestion
* You should configure a **cooldown period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
```

Dependabot and UV both call it a cooldown period, not a cool-off

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
+* You should have a process to **update quickly** to new versions, after the cool-off period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
```
quis:
```suggestion

You should have a process to **update quickly** to new versions, after the cooldown period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
 
-A Python application project typically brings together Python packages from PyPI, with
-others written in-house (or otherwise not distributed through PyPI).
+**Dependency manager** - you should use [uv][uv-readme]. This is recommended because it is fast, modern, aligned with standards like pyproject.toml and is well-maintained. Prefer it to pip, poetry and pipenv.
 
-These packages are the applications immediate dependencies. Additionally, any
-package can have its own immediate dependencies, where they draw on other
-packages. From an application's point of view, the dependencies of the packages
-it requires are its sub-dependencies.
+**Specifying dependencies** - you should use pyproject.toml to specify your direct dependencies. (This is because it is the modern recommended standard.)
```
quis:
```suggestion
### Specifying dependencies

You should use `pyproject.toml` to specify your direct dependencies because it is the modern, recommended standard.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
 
 ## Dependencies
 
-A Python application project typically brings together Python packages from PyPI, with
-others written in-house (or otherwise not distributed through PyPI).
+**Dependency manager** - you should use [uv][uv-readme]. This is recommended because it is fast, modern, aligned with standards like pyproject.toml and is well-maintained. Prefer it to pip, poetry and pipenv.
```
quis:
```suggestion
### Dependency manager

You should use [uv][uv-readme]. This is recommended because it is fast, modern, aligned with standards like `pyproject.toml` and is well-maintained. Prefer it to pip, poetry and pipenv.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
 
-These packages are the applications immediate dependencies. Additionally, any
-package can have its own immediate dependencies, where they draw on other
-packages. From an application's point of view, the dependencies of the packages
-it requires are its sub-dependencies.
+**Specifying dependencies** - you should use pyproject.toml to specify your direct dependencies. (This is because it is the modern recommended standard.)
 
-The below diagram shows a simplified view of the resulting pyramid of dependencies;
-however it is important to note that the hierarchy can repeat itself infinitely.
+**Pinning and lockfiles** - you **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
```
quis:
```suggestion
### Pinning and lockfiles

You **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-this manually (for example, installing with `uv pip install`, then using `pip freeze`) may be adequate.
-
-GDS recommends using [uv][uv-readme] for managing dependencies.
-
-Put your top-level requirements in a `requirements.in` file, and then use
-`uv pip compile requirements_for_test.in -o requirements_for_test.txt` to generate a
-`requirements.txt` file. Both the .in and .txt files should be commited to your repository.
-
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+**Updating** - you **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
```
quis:
```suggestion
### Updating

You **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,27 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
+* You should configure a **cool-off period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
+* You should have a process to **update quickly** to new versions, after the cool-off period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
+* Document your update process in your README.
```
quis:
```suggestion

Document your update process in your README.
```

quis (commented):
Few small suggestions for typos/style/readability

source/manuals/programming-languages/python/python.html.md.erb
CrystalPea:
```suggestion
### Dependency manager

Use [uv][uv-readme] for dependency management. It is fast, modern, aligned with standards like pyproject.toml, and well-maintained. Prefer it to other dependency managers (for example `pip`, `poetry`, or `pipenv`).
```

CrystalPea: Nice work on the new guidelines!

I just polished the sentence a bit - happy to help with the rest if desired 🙌🏼 .

davidr-gds: I appreciate the effort to make it more readable :) Just some quibbles with these particular sugestions:

I'm trying to make this guidance explicitly "should" or "must", to separate the recommendations from the quasi-mandatory things. "Use" doesn't seem as clear.

And I'm saying its preferred to these 3 particular dependency managers, because likely uv will be superseded in a couple of years, so I don't want this advice to suggest at that point that uv is preferred to all dependency managers.

source/manuals/programming-languages/python/python.html.md.erb
davidr-gds: good catch, thanks

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,37 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-    Security issues are found in libraries, so it is important to choose libraries that are
-    maintained and to ensure your team has a strategy to ensure security updates are installed
-    without significant delay. The [how to manage third party software dependencies][gds-way-deps] section
-    gives further context and discusses tools that can help, such as [snyk.io][].
+### Pinning and lockfiles
 
-Your README should document an easy-to-follow process by which all your Python
-dependencies can be upgraded to get bug fixes and security fixes, without introducing
-breaking changes to your build.
+You **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
```
quis: I think it’s more coherent to refer to the file, not the command here:
```suggestion
You **must** pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv.lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,35 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
+You **must** keep dependencies up to date with security patches - see also: [how to manage third party software dependencies][gds-way-deps]
+* You should configure a **cooldown period** (for example 7 days) meaning that new releases are not adopted immediately, to reduce the risk of using compromised new releases, which are usually spotted by the community and security companies and are pulled within hours. [uv and Dependabot have cool-off settings](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
 
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+You should have a process to **update quickly** to new versions, after the cooldown period. Ensure the lockfile is also updated. For example, use Dependabot to automate creating PRs, and a daily team process to manually review and merge, triggering automated deployment.
 
 ### Libraries
 
-This recommendation applies to any Python repository that intended to be installable (into
+Libraries are the exception to the rule about exact dependencies being specified and pinned.
```
currycoder, line 153: "exact dependencies being specified and pinned" - am I correct that the intention here is e.g. `flask==2.0.1`?

If so, would also suggest using add-bounds; https://docs.astral.sh/uv/reference/settings/#add-bounds
In datagovuk we have that set as follows; https://github.com/alphagov/datagovuk/blob/main/pyproject.toml#L55-L56

This means that `uv add flask` automatically adds the dependency and pins at the latest exact version.

source/manuals/programming-languages/python/python.html.md.erb
currycoder: BTW @ipdigital24 we've found the dependabot `cooldown` option to be pretty useful in datagovuk; https://github.com/alphagov/datagovuk/commit/bc88a8b4b091bab60a33cb1390988655dd03777f
It allows us to keep dependabot creating PRs but it will only do so for releases older than two weeks.

source/manuals/programming-languages/python/python.html.md.erb
davidr-gds: Nice! Changed in ea7460e

source/manuals/programming-languages/python/python.html.md.erb
davidr-gds: @currycoder As discussed privately, this is about non-libraries, and maybe the lockfile alone does the pinning we want. But feel free to follow-up if you've identied an advantage to having deps in pyproject.toml specifed with `==` as well 👍

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,36 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-GDS recommends using [uv][uv-readme] for managing dependencies.
+See the overall advice in: [How to manage third party software dependencies][gds-way-deps]. The remainder of this section covers Python-specific practicalities.
 
-Put your top-level requirements in a `requirements.in` file, and then use
-`uv pip compile requirements_for_test.in -o requirements_for_test.txt` to generate a
-`requirements.txt` file. Both the .in and .txt files should be commited to your repository.
+**Dependabot** supports uv - see [Using uv with Dependabot](https://docs.astral.sh/uv/guides/integration/dependabot/). It will bump dependencies in `pyproject.toml` and regenerate `uv.lock`.
 
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+A **cooldown period** should be configured in both [Dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-) and [uv](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
```
galund: This didn't quite make sense to me as written

```suggestion
A **cooldown period** should be configured in both
[Dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-)
and using the UV
[exclude-newer](https://docs.astral.sh/uv/reference/settings/#exclude-newer)
setting
([more information](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).
```

However I'd also like us to be able to reference something a bit more concrete about what cooldown period is appropriate.

We could _absolutely_ be missing out on patching zero days if we just go 'cooldown of a week seems fine'. Without policy here, we're going to get some serious divergence of practice and a potential incident.

[GOV.UK apparently mandates 3 days](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-178-require-dependabot-cooldown-for-external-auto-merging.md). That feels too long, to me. Their RFC points out that bad/pwned deps tend to get pulled within hours.

quis: > The `cooldown` option is only available for _version_ updates, not _security_ updates.

– https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

Dependabot should still be raising PRs for zero days straight away, even if you configure a cooldown.

galund: This is becoming _more_ clear to me - thanks, so: how does Dependabot decide whether an update is a security update? Assuming there's a GitHub human in the loop, that's fine. But then, presumably Dependabot cannot actually create a working PR, because per UV docs

> [you end] up with pull requests where uv would not be able to lock the dependencies

It seems fine to have to manually intervene for these. But it would be good to know that some team has this configuration actually set up (for Python) and is finding it workable before this PR lands :)

davidr-gds: Agreed that it seems a bit contradictory. To avoid saying anything doubtful, I'll remove the uv cooldown. Someone can add it in when folks have experience with it to answer this question.

The Dependabot cooldown though seems a widely held approach that has stand up to scrutiny.

davidr-gds: @galund are you happy to resolve this with just dependabot cooldowns or should we remove mention of cooldown for now?

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,36 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-GDS recommends using [uv][uv-readme] for managing dependencies.
+See the overall advice in: [How to manage third party software dependencies][gds-way-deps]. The remainder of this section covers Python-specific practicalities.
 
-Put your top-level requirements in a `requirements.in` file, and then use
-`uv pip compile requirements_for_test.in -o requirements_for_test.txt` to generate a
-`requirements.txt` file. Both the .in and .txt files should be commited to your repository.
+**Dependabot** supports uv - see [Using uv with Dependabot](https://docs.astral.sh/uv/guides/integration/dependabot/). It will bump dependencies in `pyproject.toml` and regenerate `uv.lock`.
 
-List dependencies only needed for development or testing into a separate `requirements-dev.txt` file.
+A **cooldown period** should be configured in [Dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-).
```
galund:
```suggestion
A **cooldown period** should be configured in [Dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-).

You **should not** specify a cooldown using the `uv`
[exclude-newer](https://docs.astral.sh/uv/reference/settings/#exclude-newer)
feature.
Otherwise, you may miss important security updates (for example fixing zero-day vulnerabilities) which Dependabot will suggest regardless of the cooldown period specified for non-security version updates.

> [!NOTE]
> This is an intentional divergence from the
> [advice in the `uv` manual](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown)
> which suggests keeping the `uv` configuration aligned with the Dependabot configuration.
```

source/manuals/programming-languages/python/python.html.md.erb
galund: Yes, tho I've made a suggestion to explain why we think using exclude-newer may not be a good idea. Will resolve this thread.

quis: We are hitting this conflict at the moment. Will feed back once we’ve resolved it.

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,36 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
+This applies to any Python repository that intended to be installable (into
 a virtual environment, a container, or onto bare metal) as a dependency of some larger system
 or application. It may be applicable to repositories that provide scripts to be run by
 developers or other end-users, but is not recommended for code that's intended to be deployed
 on its own into the cloud.
 
-Use a [`pyproject.toml`][pyproject-toml] to specify your library's configuration.
-When specifying the dependencies of your library and the version ranges with which it
-can be reasonably expected to work:
+Put **version ranges** in your [`pyproject.toml`][pyproject-toml] to specifying the dependencies of your library. These are the ranges with which it can be reasonably expected to work:
```
galund: Grammar trips

```suggestion
Put **version ranges** in your library's [`pyproject.toml`][pyproject-toml].
For each of your library's dependencies you'll need to identify the range of versions likely to be compatible:
```

source/manuals/programming-languages/python/python.html.md.erb
galund: Not sure this suggestion is at all valid as probably Dependabot doesn't care about the uv configuration. Maybe we can nuke this after all.

quis: Trouble is Dependabot will create a manifest which uv will refuse to install, if a cooldown is configured in uv.

Cooldowns in uv (via `exclude-newer`) are a good idea because they prevent developers from installing newly-compromised packages locally. 

This is where we’ve got to on Notify: https://github.com/alphagov/notifications-manuals/wiki/Dependencies#if-you-need-a-very-fresh-version-of-a-dependency-released-in-the-last-7-days

…with the caveats that:
1. We’re not using Dependabot to raise PRs 
2. It requires manual intervention

currycoder: This makes sense @quis - I think there might be a few different ways to cut the same thing.

Just to offer an alternative view of what we are currently doing on https://github.com/alphagov/datagovuk;
- We set `add-bounds: exact` for uv; https://github.com/alphagov/datagovuk/blob/d04c8ac09d166d045fe875ee2b826039bbf953d9/pyproject.toml#L57  This means that a new `uv add flask` will automatically add and pin flask at the latest version e.g. `flask==3.1.3`
- We set Dependabot cooldown to two weeks; https://github.com/alphagov/datagovuk/blob/d04c8ac09d166d045fe875ee2b826039bbf953d9/.github/dependabot.yml#L51-L52
- We let Dependabot raise PRs for version bumps - these respect the two week cooldown.

The combination of `add-bounds: exact` and letting dependabot do cooled-off version updates is working well so far.

Of course, there's a hole here where a dev may `uv add flask` when it is newly compromised but this is much less likely to happen in practice.

davidr-gds: I've now removed mention of cooldown from this PR, to hopefully yield a quick merge. It's because this cooldown has become a rabbit hole that I think deserves a separate discussion & PR. I invite @currycoder / @quis to create a follow-on PR about cooldown in Python 🙏

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,35 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
 
-Your README should document an easy-to-follow process by which all your Python
-dependencies can be upgraded to get bug fixes and security fixes, without introducing
-breaking changes to your build.
+### Pinning and lockfiles
 
-Your pinned dependencies should be fully specified in a file called `requirements.txt`, and checked
-into your version control system (VCS). For projects with only a small number of dependencies, maintaining
-this manually (for example, installing with `uv pip install`, then using `pip freeze`) may be adequate.
+You must pin your dependencies with a lock-file (unless it's a library - see Libraries) e.g. `uv.lock`. The lockfile must be used whenever installing/deploying packages, to main consistency between environments. The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
```
galund: So I think we need to have some specific guidance about which commands to use in your CI/CD pipelines. I like uv but the basic workflows don't seem that well documented (to me coming to this completely fresh trying to understand it without really having used it, which may be lots of people finding this guidance).

From what I've read and @AnnaDodson has investigated, we could perhaps write:


```suggestion
You must pin your dependencies with a lock-file (unless it's a library - see Libraries) for example `uv.lock`.
The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.

The lockfile must be used whenever installing/deploying packages (such as in CI/CD or some other non-development environment), to maintain consistency between environments. If you're using uv this probably means using `uv run --frozen` or `uv sync --frozen`.

> [!IMPORTANT]
> The above uv commands don't work the same as `uv pip sync` where the default behaviour is to use the lockfile without re-locking dependencies.
> The uv developers have
> [explained their approach balancing convenience and safe defaults](https://github.com/astral-sh/uv/issues/10793#issuecomment-2743956736).
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -122,93 +122,43 @@ We recommend you put `source .venv/bin/activate` in your `.envrc` file to ensure
...
-characteristics - they should be:
-
-1. Reproducible (predictable)
-
-    Pin your application's full dependencies – specific versions, rather than ranges – or you'll
-    get unpredictability between your dev environment and other environments. You want a
-    new starter to avoid small hard-to-spot problems. And you want parity between what you
-    test locally, what is tested by CI, and what you deploy, or you risk new issues appearing on
-    a live server. Additionally these things can be hard to diagnose.
+You should use `pyproject.toml` to specify your direct dependencies because it is the modern, recommended standard.
```
AnnaDodson, line 131: Is there a link to back up this statement?

AnnaDodson (approved):
Small comment but it looks good to me.

PR #1214, opened 2026-05-20
document health events dashboard for ops mon

myjaffoo:


huwd (dismissed):
content looks good, seems like it could use a member of EE to ✅ ?

Can you build out a bit more of a PR description too - would be good to know why we're making these changes now and what you're setting out to achive in future PRs?

galund:
Most of the content original in this PR is now on the Hub. So I'm refactoring this to link over there.

However what I've got still isn't right, because we really want to link to some documentation about the AWS Health Events dashboard and some related things... currently the docs in the Hub rather muddle this up with FinOps, so I've raised this with the EE Cloud Platform team.

PR #1234, opened 2026-06-26
Domain guidance fixes re CO:D GDS helpdesk and use cases

galund:
- fix link to CO:D service desk
- also add link to ICS helpdesk for use cases that may not be in-scope for the options suggested


louzoid-gds (approved):
LGTM

PR #1235, opened 2026-06-26
Python: uv --frozen slightly reworded. Formatting

davidr-gds:


AnnaDodson (dismissed):

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -133,14 +133,16 @@ You should use `pyproject.toml` to specify your direct dependencies because it i
 ### Pinning and lockfiles
 
 You must pin your dependencies with a lock-file (unless it's a library - see Libraries) for example `uv.lock`.
+
 The lockfile must store a **cryptographic hash of the package**, and check the hash on install, to guard against installing tampered packages.
 
-The lockfile must be used whenever installing/deploying packages (such as in CI/CD or some other non-development environment), to maintain consistency between environments. If you're using uv this probably means using `uv run --frozen` or `uv sync --frozen`.
+The lockfile must be used whenever installing/deploying packages during build or deployment (or some other non-development environment), to maintain consistency between environments. If you're using uv this probably means using `uv run --frozen` or `uv sync --frozen`.
```
galund:
```suggestion
The lockfile must be used whenever installing/deploying packages during build or deployment (or some other non-development environment), to maintain consistency between environments.
If you're using uv this probably means using `uv run --frozen` or `uv sync --frozen`.
```

source/manuals/programming-languages/python/python.html.md.erb
```diff
@@ -133,14 +133,16 @@ You should use `pyproject.toml` to specify your direct dependencies because it i
...
 
-> [!IMPORTANT]
-> The above uv commands don't work the same as `uv pip sync` where the default behaviour is to use the lockfile without re-locking dependencies.
-> The uv developers have
-> [explained their approach balancing convenience and safe defaults](https://github.com/astral-sh/uv/issues/10793#issuecomment-2743956736).
+> **_Note:_**
+> The above uv commands work differently to pip (e.g. `uv pip sync`):
+>
+> `uv run` and `uv sync` have situations where they *update* the lockfile, re-resolving the dependencies (unless you use `--frozen`). The uv developers have
```
galund:
```suggestion
> `uv run` and `uv sync` have situations where they *update* the lockfile, re-resolving the dependencies (unless you use `--frozen`).
> The uv developers have
```

galund: tbh I don't really like 'have situations'. They behave in a predictable way, and they always (AIUI) update the lock file, unless you choose the right options. They're optimised for development, at the expense of needing to be careful when setting up CI/CD. I'm all for improving my wording, tho :)

davidr-gds: My reading of https://docs.astral.sh/uv/concepts/projects/sync/#checking-the-lockfile is that it only updates the lock file if pyproject.toml has been changed in either of two ways: addition of a new dependency OR a constraint in pyproject.toml is no longer satisfied by uv.lock. So it triggered in our case because of the addition  to pyproject.toml of the new cooldown constraint.

AnnaDodson (approved):

PR #1241, opened 2026-07-13
Update supporting different browsers

DavidBiddle:
A fairly significant rewrite of the 'Supporting different browsers' page, based on a community review earlier this year.

The whole page has been expanded and reorganised, but the most significant change is the addition of a new 'Progressive enhancement' section that lists techniques to actually practically support different browsers.


JamesCGDS (approved):
Looks good to me 👍

